PR #3, opened 2015-10-15
Make npm install build the es6 project, and include the build/node/ne…

jakewins:
…o4j.js as the main entry point


jakewins:
See eg. http://mammal.io/articles/using-es6-today/

PR #18, opened 2015-11-25
Add version to API

pontusmelke:
Moved `neo4j` to `neo4j.v1` in order to facilitate API migrations in
future versions.


jakewins:
IMO, we should keep the `neo4j` root namespace, and introduce a `v1` namespace inside it instead:

    require('neo4j.v1');
    import 'neo4j.v1';

jakewins:
This should be possible by adding a `v1` folder inside the src dir, and exposing that `v1` folder though the `neo4j.js` file exports.

pontusmelke:
@jakewins any better, I think it is `require('neo4j').v1;` but otherwise I think it is what you meant?

jakewins:
+1

PR #22, opened 2015-12-17
Move summary to callback result/onCompleted, add examples

jakewins:
- # summarize() lived on the ResultStream from run - which was unhelpful,
  
  since the summary is not available until the result is consumed. This
  instead makes a summary available as an argument to `onCompleted` and
  as a field on the result object we get when we ask for a full result
- Add examples file for the manual.

### DIFF
--- a/src/v1/result.js
+++ b/src/v1/result.js
@@ -38,7 +38,6 @@ class Result {
     this._p = null;
     this._statement = statement;
     this._parameters = parameters;
-    this.summary = {};
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/v1/result.js
```diff
@@ -97,20 +99,12 @@ class Result {
   subscribe(observer) {
     let onCompletedOriginal = observer.onCompleted;
     let onCompletedWrapper = (metadata) => {
-      this.summary = new ResultSummary(this._statement, this._parameters, metadata);
-      onCompletedOriginal.call(observer);
+      var sum = new ResultSummary(this._statement, this._parameters, metadata);
```
oskarhane: `var` => `let`

jakewins:
@oskarhane you +1 on this otherwise? Fixed `var`->`let`

PR #23, opened 2016-01-08
1.0 tck

RagnarW:
TCK for javascript

https://trello.com/c/KLvMNddz/639-js-driver-results-need-to-have-eg-length-that-is-correct
shuld be fixed to resolve result handling.


oskarhane:
lgtm now, merging.

PR #28, opened 2016-02-15
export graph-types under 'types' key

canibanoglu:
Well, here it goes. The types are currently under a `types` key because I felt exporting them all under root would clutter the API. I'm more than happy to change things according to your preferences of course :)


oskarhane:
I like this change, very much needed. Could you resolve the conflicts and update the PR and I'll go ahead and merge it @canibanoglu

canibanoglu:
@oskarhane Alright, this no longer has any conflicts :) Let me know if I can do something else :)

oskarhane:
Awesome, thanks

PR #35, opened 2016-03-04
Distribute via bower

oskarhane:
- cleanup distributed packages to just include what consumers need
  - for npm
  - for bower


oskarhane:
Before merge, make sure build pipeline build and push before release tagging.

PR #36, opened 2016-03-04
1.0 tck - ResultSummary API and Equality

RagnarW:
The follow needs to be fixed before merge:

Equality:
- Decide what equality means. The feature file is designed for using IDs
- Should Path direction matter? The feature file is designed to test direction
- Should JS graph-objects have equals. There is no equality in JS should this be implemented?

ResultApi:
- ProfiledPlan: should the returned rows/records return "records" or "rows"
- Statistics: should statics be specified and should update statistics be part of it?


pontusmelke:
@RagnarW needs rebase

PR #41, opened 2016-03-17
Support for encrypted connections

jakewins:
See first commit message, and docs in second commit message, for details.


src/v1/internal/ch-node.js
```diff
@@ -18,16 +18,152 @@
...
+    return socket;
+  },
+  TRUST_ON_FIRST_USE : function( opts, onSuccess, onFailure ) {
+    let tlsOpts = {
+      // Because we manually verify the certificate against known_hosts
+      rejectUnauthorized: false
+    };
+
+    let socket = tls.connect(opts.port, opts.host, tlsOpts, function () {
+      var serverCert = socket.getPeerCertificate(true);
```
jakewins, line 103: Add raw=true to highlight why we pass this in here

PR #47, opened 2016-04-06
Unified errors and made Neo4jError part of public API

pontusmelke:


oskarhane:
This is very welcome, thanks!  
I think you forgot to import `newError` to `integer.js`.

pontusmelke:
@oskarhane fixed, thanks!

PR #49, opened 2016-04-06
Fix the powershell script for starting server for tests on windows

zhenlineo:


pontusmelke:
Looks good, could you update this method as well https://github.com/neo4j/neo4j-javascript-driver/blob/1.1/test/v1/tck/steps/util.js#L383

PR #60, opened 2016-04-15
example code doesn't work without v1

whatSocks:
Having two different ways to create a driver instance is confusing. `examples/node.js` does not use v1, but the only way I could get it to work was with v1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ bower install neo4j-driver
 ```
 
 ```javascript
-var neo4j = require('neo4j-driver');
+var neo4j = require('neo4j-driver').v1;
 ```
 
 ## Include in web browser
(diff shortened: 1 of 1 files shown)


pontusmelke:
@whatSocks Thanks a lot for fixing this! For me to merge this I need you to sign our CLA, http://neo4j.com/developer/cla/ . It is just a single email so it should be pretty quick.

whatSocks:
I'm trying to get the examples in the readme to work at all and am struggling. I'll update the readme I find any other little things.

whatSocks:
Updated the three usage examples (they were creating a node with no age parameter and trying to return the age parameter). Now readers of the three examples can run all three and create three different person nodes, Alice, Bob and James.

pontusmelke:
@whatSocks Did you sign the CLA?

whatSocks:
yep I mailed it in

pontusmelke:
Awesome!

PR #65, opened 2016-05-03
sync after reset on session.close

pontusmelke:
In order to preserve roundtrips we didn't call `sync` after `reset` in
`session.close`. However this means that if `session.close` is invoked
in order to close a long-running query nothing will happen until you send
the next query.

### DIFF
--- a/lib/v1/driver.js
+++ b/lib/v1/driver.js
@@ -158,10 +158,9 @@ var Driver = (function () {
         // wrapper around Connection anyway, so it makes little difference.
 
         // Queue up a 'reset', to ensure the next user gets a clean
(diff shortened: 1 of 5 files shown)


pontusmelke:
I don't know how to write a test for this

oskarhane:
I think we would need a test for this though.  
When trying this feature out I sometimes get the error message `Neo.ClientError.Request.Invalid: 'acknowledging a failure' cannot be done when a session is in the 'IDLE' state.` when executing a query right after the closed one.

oskarhane:
Oh, the query is terminated directly btw, so the termination works.

pontusmelke:
@oskarhane Ok I think I fixed it now.

Issue #64, opened 2016-04-28
Mangled records returned from bolt driver.


holmok:
local neo4j version: 3.0.0-M05
driver version 1.0.0

code:

```
'use strict';
var neo4j = require('neo4j-driver').v1;
var driver = neo4j.driver("bolt://localhost", neo4j.auth.basic("neo4j", "-----"));

var session = driver.session();

session.run("MATCH (n:ATTRACTION) RETURN n LIMIT 20;")
    .subscribe(
        {
            onNext: function (record) {
                var node = record.get('n');
                console.log(node);
            },
            onCompleted: function () {
                // Completed!
                session.close();
            },
            onError: function (error) {
                console.log(error);
            }
        }
    );
```

all the nodes look fine, but one looks like this:

```
Node {
  identity: Integer { low: 4373, high: 0 },
  labels: 
   [ '�q��N�\u0011\u0015��',
     Integer { low: 65, high: 0 },
     Integer { low: 84, high: 0 },
     Integer { low: 84, high: 0 } ],
  properties: Integer { low: 82, high: 0 } }
```

it's a fairly random smattering.  any ideas?

pontusmelke:
Hi @holmok,
We have never encountered this before, is it always the same node (`identity: 4373`) or does it happen to different nodes? If it only happens to that node could you add `WHERE id(n) = 4373` and also try to do `console.log(n.labels[0])`?

Thanks for reporting this!

vmarovic:
Hi,
I seem to have same issue:

Record {
  keys: [ 'u._id' ],
  length: 1,
  _fields: [ '�q��$d5ceb4ea-e6f9-4397-823b-103cf4a' ],
  _fieldLookup: { 'u._id': 0 }
}

It does seem to happen on the same nodes always, and only when returning large number of records. When i fetch this record only it is returned properly. Also if i skip to location near this record and limit results to 10 it is returned ok also.

pontusmelke:
Ok thanks, we'll look into this issue

vmarovic:
Hi just one more thing
If i write query like 
match (u:User) return u.name
I get some names mangled, but if i typecast it to string in cypher
match (u:User) return toString(u.name)
results are returned fine.

pontusmelke:
That doesn't look right. With _some names_, do you mean that is it the same as in the other example that it only happens when there is a large number of records?

vmarovic:
[...]

PR #68, opened 2016-05-09
Mangled records hotfix

kanban-agency:
Combined buffers is not read from correctly so we just copy all data to
one buffer.

Issue #64

### DIFF
--- a/lib/v1/internal/chunking.js
+++ b/lib/v1/internal/chunking.js
@@ -224,7 +224,29 @@ var Dechunker = (function () {
         if (this._currentMessage.length == 1) {
           message = this._currentMessage[0];
         } else {
-          message = new _buf2['default'].CombinedBuffer(this._currentMessage);
+          // This returns mangled records
+          //message = new _buf2['default'].CombinedBuffer(this._currentMessage);
+
+          var length = 0;
+
+          this._currentMessage.forEach(function (m) {
+              length += m.length;
+          })
+
+          var buf = Buffer.alloc(length);
+
+          var p = 0;
+
+          this._currentMessage.forEach(function (m) {
+            for(var i = 0; i < m.length; i++,p++) {
+              var t = m._buffer.readUInt8(i);
+              buf.writeUInt8(t, p);
+            }
+          })
+
+          this._currentMessage[0]._buffer = buf;
+
+          message = this._currentMessage[0];
         }
         this._currentMessage = [];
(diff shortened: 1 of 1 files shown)


pontusmelke:
Hi, thanks a lot for helping out. I was just looking at this issue but you beat me to it :)

In order for me to merge you need to sign our CLA, http://neo4j.com/developer/cla/ . It is just a single email so it should be pretty quick.

Furthermore a couple of issues need to be adressed: 
- the fix should go into the `src` not `lib`, the `lib` folder contains generated files from the `src` directory
- the PR should be opened against 1.1 instead of 1.0
- If you could add a failing test in for example something like this to `session.test.js`

```
it('should return lots of data', function (done) {
    session.run("UNWIND range(1,1000) AS x CREATE (:ATTRACTION)")
      .then(function () {
        session.run("MATCH (n) RETURN n")
          .subscribe({
              onNext: function (record) {
                var node = record.get('n');
                expect(node.labels[0]).toEqual("ATTRACTION");
              },
              onCompleted: function () {
                // Completed!
                session.close();
                done()
              },
              onError: function (error) {
                console.log(error);
              }
            }
          )
      });
  });
```
- Lastly your fix simply adds everything to one buffer, I would prefer a fix of the `CombinedBuffer`.

Thanks again for helping out.

kanban-agency:
Ok. We'll see if we can sort out reading from CombinedBuffer, and we'll send CLA by the time we submit that.

gurkerl83:
Hi, I don't know if its related but we ran into a similar buffer issues a while ago. We are using neo4j-driver within an electron application and therefore the correct transport channel does not be resolved. Even if the net lib is available (within electron) the NodeChannel does never get selected. We have modified the selection behaviour setting nodes net lib as default (if statement).

``` javascript
//modified order - default channel = node, fallback channel = web-socket

if( NodeChannel.available ) {
    Channel = NodeChannel.channel;
}
else if( WebSocketChannel.available ) {
    Channel = WebSocketChannel.channel;
}
else {
    throw newError("Fatal: No compatible transport available. Need to run on a platform with the WebSocket API.");
}
```

You can find a fork of those modifications including the replacement of text-encoding with text-encoding-utf-8 (cjs enabled) in order to resolve bundling issues with browserify / webpack available at https://github.com/gurkerl83/neo4j-javascript-driver.

Thanks,

PR #76, opened 2016-05-20
Make js driver to use external python script to control neo4j server

zhenlineo:


gulpfile.js
```diff
@@ -247,28 +195,42 @@ gulp.task('set', function() {
...
-    ]));
-  }
+
+var neo4jHome  = path.resolve('./build/neo4j');
+var neorunPath = path.resolve('./neokit/neorun.py');
+
+gulp.task('start-neo4j', function() {
+    var neorunArgs;
+    //var neorunArgs = minimist(process.argv.slice(2), { string: 'opts' }).opts;
+    if (neorunArgs == null)
```
pontusmelke: How can `neoRunArgs` not be `null` here? Typically the more ideomatic way is doing

```
neoRunArgs = neoRunArgs || '-v 3.0.1 -p neo4j'
```

neokit
```diff
@@ -0,0 +1 @@
+Subproject commit d2059f1a4bfc8775697937eea5285b69864f2b7f
```
pontusmelke: You should probably not commit this?

test/v1/tck/steps/util.js
```diff
@@ -354,48 +354,38 @@ function changeCertificates(keyFile, certFile) {
   fs.writeFileSync(neo4jCert, cert);
 }
 
-function restart(callback) {
+function restart() {
+  // how to sequentially execute these two methods????
```
pontusmelke: I think the easiest way is to use timeouts as we did before.

PR #83, opened 2016-06-15
Fix #82 - Added Record.has(key)

### LINKED ISSUE #82: Checking if a value exists in a Record
There is no _"proper"_ way to check if a value exists or not in a Record without trying to access and catching for errors.

Take this oversimplified example:

``` javascript
const queries = [
   'MATCH (a:Apple) RETURN COUNT(a) AS apples' ,
   'MATCH (a:Apple), (o:Orange) RETURN COUNT(a) AS apples, COUNT(o) AS oranges'
];

const query = queries[Math.random() > .5 ? 1 : 0];

session.run(query).then((data) => {
    console.log(data[0].get('apples'));

    // will throw if first query is chosen
    console.log(data[0].get('oranges')); 
});
```

There should be a way of checking for fields before trying to use them:

``` javascript
session.run(query).then((data) => {
    console.log(data[0].get('apples'));

    // will throw if first query is chosen
    if(data[0].has('oranges')) {
        console.log(data[0].get('oranges')); 
    }
});
```

legraphista:
Added: has method of Record to check if a value exists before using it
Added: tests for the has method

### DIFF
--- a/src/v1/record.js
+++ b/src/v1/record.js
@@ -100,6 +100,22 @@ class Record {
 
     return this._fields[index];
   }
+
+  /**
+   * Check if a value from this record, either by index or by field key, exists.
+   *
+   * @param {string|Number} key Field key, or the index of the field.
+   * @returns {boolean}
+   */
+  has( key ) {
+    // if key is a number, we check if it is in the _fields array
+    if( typeof key === "number" ) {
+      return ( key >= 0 && key < this._fields.length );
+    }
+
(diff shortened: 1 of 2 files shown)


pontusmelke:
Thank @legraphista for fixing this!

In order for me to merge this you need to sign our [CLA](http://neo4j.com/developer/cla/)

Regards,
Pontus

legraphista:
Thank you @pontusmelke. I have signed up for Neo4j's CLA.

Issue #91, opened 2016-06-21
Rollback calls failing silently after invalid pack error


kpdecker:
When rolling back after this root error:

```
Cannot pack this value: undefined
```

the promise fails to reject or resolve:

```
      return tx.rollback()
        .then(() => {
          console.log('resolve!');
          session.close();
          operations = tx = session = undefined;
        },
        (err) => {
          console.log('reject!', err);
          session.close();
          operations = tx = session = undefined;

          log(['error', 'neo', 'transaction', 'rollback'], {sourceId, transactionId, error: err.stack});
          return Promise.reject(err);
        });
```

Trying to log through this a bit, I found the following:

```
packStruct 10
Error
    at Packer.packStruct (~/node_modules/neo4j-driver/lib/v1/internal/packstream.js:160:21)
    at Connection.run (~/node_modules/neo4j-driver/lib/v1/internal/connector.js:384:20)
    at _runDiscardAll (~/node_modules/neo4j-driver/lib/v1/transaction.js:236:8)
    at Object.rollback (~/node_modules/neo4j-driver/lib/v1/transaction.js:168:24)
    at Transaction.rollback (~/node_modules/neo4j-driver/lib/v1/transaction.js:115:35)
    at Object.rollback (~/src/neo4j.js:144:25)
    at ~/src/neo4j.js:103:31
    at lib$es6$promise$$internal$$tryCatch (~/node_modules/neo4j-driver/lib/external/es6-promise.js:338:14)
    at lib$es6$promise$$internal$$invokeCallback (~/node_modules/neo4j-driver/lib/external/es6-promise.js:353:15)
    at lib$es6$promise$$internal$$publish (~/node_modules/neo4j-driver/lib/external/es6-promise.js:321:9)
    at lib$es6$promise$$internal$$publishRejection (~/node_modules/neo4j-driver/lib/external/es6-promise.js:263:5)
    at lib$es6$promise$asap$$flush (~/node_modules/neo4j-driver/lib/external/es6-promise.js:120:7)
    at doNTCallback0 (node.js:430:9)
    at process._tickDomainCallback (node.js:400:13)
pack ROLLBACK string
pack {} object
packStruct 2f
```

To which the server responded:

```
message! Structure { signature: 112, fields: [ { fields: [] } ] }
message! Structure { signature: 112, fields: [ {} ] }
message! Structure {
  signature: 127,
  fields:
   [ { code: 'Neo.ClientError.Request.InvalidFormat',
[...]

PR #92, opened 2016-06-22
Add support for sparse javascript objects

kpdecker:
For objects, silently drop undefined fields. For arrays, convert undefined entries to null. (Supporting true undefined values in arrays would require protocol changes, so null seems good enough for now)

Related to #91


pontusmelke:
Awesome! I was just looking at this.

In order for me to merge this you need to sign our [CLA](https://neo4j.com/developer/cla/)

kpdecker:
Sent

pontusmelke:
Thanks, and just one more thing :)
Would you mind re-open the PR but make it against 1.0? (select 1.0 as base when you create the PR).

PR #94, opened 2016-06-22
Add preferNativeInt for unpack operation

kpdecker:
This forces database results to be read as language-native number values unless they are out of range, which reduces the code overhead for applications that do not need to worry about the extended range.

### DIFF
--- a/src/v1/integer.js
+++ b/src/v1/integer.js
@@ -719,6 +719,23 @@ Integer.fromValue = function(val) {
 };
 
 /**
+ * Returns a number or Integer for a given pair of low and high bits, preferring number when in scope.
+ * @access private
+ * @param {number} lowBits The low 32 bits
+ * @param {number} highBits The high 32 bits
+ * @returns {!Integer|number} The corresponding Integer value
+ * @expose
+ */
+Integer.asNative = function(lowBits, highBits) {
+  var val = new Integer(lowBits, highBits);
+  if (val.greaterThan(MIN_NATIVE) && val.lessThan(MAX_NATIVE)) {
+    return val.toNumber();
+  } else {
+    return val;
+  }
+};
+
+/**
  * @type {number}
  * @const
  * @inner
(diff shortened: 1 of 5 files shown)


pontusmelke:
Sorry for not getting back to you on this. We like this feature but we maybe like to support in a slightly different manner. There may be several ways a user want to affect serialization and we are discussing adding this kind of support across all drivers. That is what taking us so long to decide on this PR.

whatSocks:
this would be useful

src/v1/integer.js
```diff
@@ -802,6 +819,24 @@ Integer.MAX_VALUE = Integer.fromBits(0xFFFFFFFF|0, 0x7FFFFFFF|0, false);
...
+
+/**
+ * Maximum safe native value.
+ * @type {!Integer}
+ * @const
+ * @inner
+ * @private
+ */
+var MAX_NATIVE = Integer.fromValue(Number.MAX_SAFE_INTEGER);
+
```
swist, line 838: Is there a reason you have added these as variables in module scope? Perhaps you could attach them to the `Integer` constructor like so:

```
Integer.MIN_NATIVE = Integer.fromValue(Number.MIN_SAFE_INTEGER)
Integer.MAX_NATIVE = Integer.fromValue(Number.MAX_SAFE_INTEGER)
```

I'd do it myself, but I'm not sure if mucking about with someone's PR is frowned upon :)

It would be nice to have these available in the same place like other MIN_MAX values, perhaps you could even reuse the others, but I don't have the driver on hand to test myself. Might try over lunch

[edit] d'oh, just saw @pontusmelke's comment...

brian-gates:
I wrote a function that recursively maps Result objects to their Javascript native equivalents.

https://github.com/codex-digital/cypher-stream/blob/master/util/to-native.js

Figured it might be useful to someone.

zhenlineo:
Hi, sorry for not coming back to this PR for a long time. We finally agreed to keep `neo4j.int` but provide a method `InSafeRange` to help to decide whether to convert a `neo4j.int` to `string` or `int`. The PR is https://github.com/neo4j/neo4j-javascript-driver/pull/153

laurieboyes:
Is there any chance of reopening the discussion on this or a similar PR? The integer functions provided are helpful, but not a complete solution, as using them requires that you know exactly the format of the data being returned, as highlighted in this issue raised in May: https://github.com/neo4j/neo4j-javascript-driver/issues/245

We're currently in the process of moving some of our apps over from the thingdom client so that they can take advantage of bolt, but this issue feels like a major disadvantage of the new client

PR #95, opened 2016-06-23
Slighly more robust tofu detection

pontusmelke:
The checking of if platform supports trust-on-first-use is broken on node 6.
We need a more robust way of detecting this, but until then this will have to do.

### DIFF
--- a/src/v1/internal/features.js
+++ b/src/v1/internal/features.js
@@ -26,8 +26,8 @@ const FEATURES = {
       // This is insane. We are verifying that we have a version of getPeerCertificate
       // that supports reading the whole certificate, eg this commit:
       // https://github.com/nodejs/node/commit/345c40b6
-      let desc = require('tls').TLSSocket.prototype.getPeerCertificate.toString();
-      return desc.startsWith("function getPeerCertificate(detailed)");
+      let desc = require('tls').TLSSocket.prototype.getPeerCertificate;
+      return desc.length() >= 1;
     } catch( e ) {
       return false;
     }


src/v1/internal/features.js
```diff
@@ -27,7 +27,7 @@ const FEATURES = {
       // that supports reading the whole certificate, eg this commit:
       // https://github.com/nodejs/node/commit/345c40b6
       let desc = require('tls').TLSSocket.prototype.getPeerCertificate.toString();
-      return desc.startsWith("function getPeerCertificate(detailed)");
+      return desc.indexOf("function getPeerCertificate(detailed)") >= 0;
```
kpdecker: `require('tls').TLSSocket.prototype.getPeerCertificate.length >= 1` is going to be more efficient and less prone to breakages.

src/v1/internal/features.js
```diff
@@ -26,8 +26,8 @@ const FEATURES = {
       // This is insane. We are verifying that we have a version of getPeerCertificate
       // that supports reading the whole certificate, eg this commit:
       // https://github.com/nodejs/node/commit/345c40b6
-      let desc = require('tls').TLSSocket.prototype.getPeerCertificate.toString();
-      return desc.startsWith("function getPeerCertificate(detailed)");
+      let desc = require('tls').TLSSocket.prototype.getPeerCertificate;
+      return desc.length() >= 1;
```
kpdecker, line 30: That's `desc.length >= 1`, it's a property not a function.

pontusmelke, line 30: oops, thanks

PR #100, opened 2016-06-26
Added: Record.toObject() to transform a Record into an Object

legraphista:
There are cases when you want to:

``` cypher
RETURN { a: 1, b: true, c: 'Hello World' } AS data 
```

and you don't want to 

``` javascript
const data =  {
   a: record.get('a'),
   b: record.get('b'),
   c: record.get('c')
};
```

so you can just

``` javascript
const data = record.toObject();
```

### DIFF
--- a/src/v1/record.js
+++ b/src/v1/record.js
@@ -77,6 +77,20 @@ class Record {
   }
 
   /**
+   * Generates an object out of the current Record
+   *
+   * @returns {Object}
+   */
+  toObject() {
+    const object = {};
+    this.forEach((value, key) => {
+      object[key] = value
+    });
+
+    return object;
+  }
+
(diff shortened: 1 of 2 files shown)


legraphista:
This feature may not make sense in other languages, but accessing data fields via a `.get()` method is more _Java_-esc and not so _Javascript_ friendly.

oskarhane:
+1 on this. Good stuff.

brian-gates:
I love this feature, but I suspect it requires (or would greatly benefit by) recursion.

e.g. I suspect this query will still produce Node and Rel objects:

```
MATCH (n)-[r]-()
with n, r limit 1
return { node: n, rel: r } as data
```

PR #110, opened 2016-07-22
Fix/known hosts fingerprint duplication error/upstream 1.0

legraphista:
Fixes issue #107 & #108 
Fix: `loadFingerprint` callback only called once
Fix: `storeFingerprint` does not duplicate information in the `known_hosts` file
Add: test cases

PR #111, opened 2016-07-25
Make known-host file to tolerate duplicated serverId

zhenlineo:
If there is more than one dup in the known-host file, then the first one will be picked.
Based on #110

### DIFF
--- a/test/internal/tls.test.js
+++ b/test/internal/tls.test.js
@@ -76,6 +76,41 @@ describe('trust-on-first-use', function() {
 
   var driver;
 
+  it('should not throw an error if the host file contains two host duplicates', function(done) {
+    'use strict';
+    // Assuming we only run this test on NodeJS with TOFU support
+    if( !hasFeature("trust_on_first_use") ) {
+      done();
+      return;
+    }
+
+    // Given
+    var knownHostsPath = "build/known_hosts";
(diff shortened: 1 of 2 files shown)


test/internal/tls.test.js
```diff
@@ -90,22 +90,23 @@ describe('trust-on-first-use', function() {
...
+    driver.session(); // write into the knownHost file
+
+    // duplicate the same serverId twice
+    setTimeout(function() {
+      var text = fs.readFileSync(knownHostsPath,'utf8');
+      fs.writeFileSync(knownHostsPath, text+text);
+    }, 1000);
+
     // When
     driver.session().run("RETURN true AS a").then( function(data) {
```
legraphista, line 108: This should be encapsulated in a setTimeout(2000) in order to wait for the above one to duplicate the file data

zhenlineo, line 108: This test passed on test server. So maybe the timeout is okay? Though currently we just tested on linux, our windows test server might require longer to see the change. Currently the test failed at a known tck failing test :D I am trying to fix it now so get this PR ready.

legraphista, line 108: Yes, but since `driver.session().run()` will fire off the queue at an arbitrary time, you cannot know if the file has been modified to have duplicated lines yet.

PR #124, opened 2016-08-26
Deprecate TRUST_SIGNED_CERTIFICATES

pontusmelke:
`TRUST_SIGNED_CERTIFICATES` is behaving differently on different platforms. In
order to clean this up we deprecate it and replace with
- `TRUST_SYSTEM_CA_SIGNED_CERTIFICATES` which trusts the system defaults on the
  platform
- `TRUST_CUSTOM_CA_SIGNED_CERTIFICATES` which trusts a particular root
  certificate (can be a self-signed certificate for example)


oskarhane:
lgtm but I cannot merge since github says it has conflicts.

PR #127, opened 2016-08-31
Expose onComplete method from Driver.

eve-bright:
Provides quick feedback to client that credentials are correct.


pontusmelke:
We usually don't commit the generated files in `lib`

PR #130, opened 2016-09-02
Expose version and timings in summary of driver

pontusmelke:
Note that in the js driver it makes more sense to just expose version in the
metadata in

``` javascript
driver.onCompleted = function(meta) {
    console.log(meta.server);
}
```

instead of putting it on session since it is the place where we are certain that the
information has arrived.


oskarhane:
Yay!  
Looks good to me, but why not on the 1.0 branch?

pontusmelke:
@oskarhane The only reason was that it requires server 3.1 but that reason is not really valid since 1.0 will also work on 3.1. I'll discuss with @nigelsmall, we did it on 1.1 in the java driver so we need to be consistent at least and if so also backport there.

pontusmelke:
will be reopened against 1.0

PR #134, opened 2016-09-05
Added support for custom auth tokens

pontusmelke:


src/v1/index.js
```diff
@@ -31,8 +31,20 @@ export default {
...
-    basic: (username, password) => {
-      return {scheme: "basic", principal: username, credentials: password};
+    basic: (username, password, realm = undefined) => {
+      if (realm) {
+        return {scheme: "basic", principal: username, credentials: password, realm: realm};
+      } else {
+        return {scheme: "basic", principal: username, credentials: password};
+      }
+    },
+    custom: (principal, credentials, realm, scheme = "basic", parameters = undefined ) => {
```
craigtaverner: Should scheme be optional here?

test/v1/driver.test.js
```diff
@@ -76,6 +76,32 @@ describe('driver', function() {
     driver.session();
   });
 
+  it('should be possible to create custom auth tokens', function(done) {
```
craigtaverner, line 92: Perhaps also a test for basic with realm?

PR #141, opened 2016-09-20
Routing driver

pontusmelke:
Support for routing operations in the js driver:

Usage:

``` javascript
let driver = neo4j.driver("bolt+routing://127.0.0.1:9001", neo4j.auth.basic("neo4j", "neo4j"));
let session = driver.session(neo4j.session.WRITE);
let tx = session.beginTransaction();

tx.run('CREATE ()');

tx.commit()
    .then((results) => {/*handle results*/})
    .catch((err) => {
       if (err.code === neo4j.error.SESSION_EXPIRED) {
         //session expired, ask the driver for a new one and replay
       } else if (err.code === neo4j.error.SERVICE_UNAVAILABLE) {
         //driver has no contact with the cluster, recreate the driver and replay
       } else {
         //a "normal error" has occurred handle it appropriately
      }
```

By using the scheme `bolt+routing` the driver will ask the seed server to provide a view of the current cluster, which contains information on where to direct read requests, write requests and where to direct new routing requests. The driver maintains this information and will automatically spread the load among these servers. If servers go offline the driver will remove them and when running out of servers a new routing request will be performed to refresh the view of the cluster. The return from the routing request also contains a `ttl`, when this time has expired the driver will also refresh the cluster view.

On running requests two types of errors can happen:
- `SESSION_EXPIRED`: Means that the server is unable to perform the current action, the application should ask for a new session from the driver and retry the operation.
- `SERVICE_UNAVAILABLE`: The driver has completely lost contact with the cluster, the driver needs to be recreated in order for any new operations to be possible.


src/v1/driver.js
```diff
@@ -119,17 +130,32 @@ class Driver {
...
-      if(cb) { cb(); }
+      if (cb) {
+        cb();
+      }
     });
   }
 
+  //Extension point
+  _acquireConnection(mode) {
+   return Promise.resolve(this._pool.acquire(this._url));
```
nigelsmall: Where does `mode` get used?

src/v1/internal/round-robin-array.js
```diff
@@ -0,0 +1,82 @@
...
+      //make sure we are in range
+      if (this._items.length === 0) {
+        this._index = 0;
+      } else {
+        this._index %= this._items.length;
+      }
+      index = this._items.indexOf(item, index);
+    }
+  }
+}
```
technige, line 80: Alternatively you could move the `_index` manipulation logic into `next` so that it does bounds checking and adjustment before using the value.

pontusmelke: `splice` does in-place rewrite of the array I want to be sure that the index is up-to-date after each removal in the loop

PR #137, opened 2016-09-06
remove terminated connections from pool

GertSallaerts:
Related to #136

PR #149, opened 2016-09-30
improve closed connection handling

GertSallaerts:
Related to my previous PR (#137).
- Not sure why it seemed to work, but the `_handleConnectionTerminated` handler never really worked because it wasn't bound correctly and `'this.onerror` is always undefined. So I bound the function.
- Upon reflection, I don't think that the server closing a connection should be considered as an error, the connection is simply not open anymore for whatever reason. Once you try to open a new connection, you'll get a more accurate error message like the server not being reachable anymore for example or no error is there is a failover setup configured, ... So instead of throwing an error, I close the connection.
- The pool did not call the `_destroy` method on resources that are detected as invalid during acquisition, so they weren't being cleaned up correctly.

### DIFF
--- a/src/v1/internal/ch-node.js
+++ b/src/v1/internal/ch-node.js
@@ -266,6 +266,7 @@ class NodeChannel {
     this._open = true;
     this._error = null;
     this._handleConnectionError = this._handleConnectionError.bind(this);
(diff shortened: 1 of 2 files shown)


pontusmelke:
Awesome, thanks! I am currently looking into this. I think I will end up putting the `onerror` call back.

PR #153, opened 2016-10-26
Expose new integer methods

pontusmelke:
- Expose `inSafeRange`, `toNumber`, and `toString` in order to
  facilitate integer handling.
- Deprecate exposing `Integer`: It does not make sense to expose the `Integer` type to users. Users should use a framework of their liking instead and we only expose `inSafeRange`, `toNumber` and `toString` functions.


pontusmelke:
NOTE: is built on top of https://github.com/neo4j/neo4j-javascript-driver/pull/141, so merge that one first

PR #163, opened 2016-11-24
Added params and made format consistent for all queries

petraselmer:
This must be fwd-merged to 1.1


test/v1/examples.test.js
```diff
@@ -86,7 +87,7 @@ describe('examples', function() {
     //end::configuration[]
 
     var s = driver.session();
-    s.run( "CREATE (p:Person { name: {name} })", {name: "The One"} )
+    s.run( "CREATE (p:Person {name: {name}})", {"name": "The One"} )
```
oskarhane: No need to quote js object keys, unless the key includes breaking characters like space etc.

petraselmer: Interesting - I didn't know that. Cool - will change

PR #172, opened 2016-12-14
Expose server info in result summary from sessions and transactions

oskarhane:
Exposed as a map in `result.summary.server`:

```
{
  address: 'localhost:7687',
  version: 'Neo4j/3.1.0'
}
```

### DIFF
--- a/src/v1/driver.js
+++ b/src/v1/driver.js
@@ -67,8 +67,8 @@ class Driver {
    */
   _createConnection(url, release) {
     let sessionId = this._sessionIdGenerator++;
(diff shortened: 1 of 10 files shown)


zhenlineo:
Need rebase to fix the driver version

zhenlineo:
What about unhappy case where we failed to run a query and how does a user still access `summary.server`?

lutovich:
Rebased. Currently there exists no API to retrieve summary after failure. I think it should be added in a separate PR. Also added couple tests for `summary.server` with observer API.

PR #178, opened 2016-12-19
Prefer node channel over websocket channel when both available

### LINKED ISSUE #154: neo4j-driver in electron apps
Hello,

I tried to use the (very cool) neo4j-driver within an electron app. [http://electron.atom.io/](http://electron.atom.io/) and faced this error: 

`Uncaught Error: Don't know how to send buffer: NodeBuffer( position=0 )60 60 b0 17 00 00 00 01 00 00 00 00 00 00 00 00 00 00 00 00`

See also: [Stackoverflow](http://stackoverflow.com/questions/37465563/neo4j-official-javascript-driver-with-electron)

I found a possible solution for that:

If you change the order of the first two if blocks here:
[L65 in connector.js](https://github.com/neo4j/neo4j-javascript-driver/blob/1.1/lib/v1/internal/connector.js#L65)

to:

``` javascript
var Channel = undefined;
if (_chNode2["default"].available) {
  Channel = _chNode2["default"].channel;
}else if (_chWebsocket2["default"].available) {
  Channel = _chWebsocket2["default"].channel;
}else {
  throw (0, _error.newError)("Fatal: No compatible transport available. Need to run on a platform with the WebSocket API.");
}
```

It works perfectly.

If changing the order of the if-blocks doesn't interfere with anything else, I would suggest to change them in ordner to make the driver working in electron apps.

lutovich:
Fixes #154

### DIFF
--- a/src/v1/internal/connector.js
+++ b/src/v1/internal/connector.js
@@ -26,11 +26,11 @@ import {Node, Relationship, UnboundRelationship, Path, PathSegment} from '../gra
 import {newError} from './../error';
 
 let Channel;
-if( WebSocketChannel.available ) {
-    Channel = WebSocketChannel.channel;
+if( NodeChannel.available ) {
+  Channel = NodeChannel.channel;
 }
-else if( NodeChannel.available ) {
-    Channel = NodeChannel.channel;
+else if( WebSocketChannel.available ) {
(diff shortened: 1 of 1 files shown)


rvazarkar:
This is a huge fix for Electron based apps using neo4j-driver

PR #183, opened 2016-12-29
Started Adding Type Definitions

TimothyMischief:
Covered All Externals I think. First time writing Type Definitions so anyone else feel free to take over and do a better job. Also my first PR, so let me know if there's anything I could do better. This will definitely need a few iterations it's far from complete but I need more experienced eyes over it.

There's a trailing comma that wound up in one of the javascript files that I'm about to delete with another commit.


lutovich:
Thanks for this contribution @SnappyCroissant!
We need to learn a bit more about type definitions to be able to evaluate this change.
I'll create a task for this and we'll get back to you in a while.

pocesar:
thanks @SnappyCroissant that was a god send, already using the typings on my branch. If I find anything I'll make a PR to you

pocesar:
I've made the typings functional from the initial work from @SnappyCroissant with type safe results from `session.run`. there are still some things missing that are currently *non type safe* that uses `Object` or `Function` as types, but that might be implemented gradually, neo4j lib is huge

I've made the mistake to place the typings in the working branch though... https://github.com/pocesar/neo4j-javascript-driver

nbransby:
Any reason is has not been merged?

lutovich:
@nbransby lack of time to assess, review and learn about type definitions is the only reason. It is definitely on our list of priorities and this PR will not be lost.

nbransby:
Maybe I can help with that

lutovich:
@nbransby help would be appreciated. We are currently busy with 1.3 features. I hope to complete them this week. Then there are couple small issues to fix and then I'll try to catch up with type definitions. How would you like to help? Maybe you could review the PR first?

At first glance, changes look good but I'm not sure type definitions should live in `lib` directory. Currently only generated sources live there and hand-written code lives in `src`.

janwo:
Any updates on this?

nbransby:
Not from me, getting by without the definitions as API for basic use is very small, I want to work on this but haven't found the time

lutovich:
Small question to everyone involved. Is it true that with separate type definitions, like in this PR, we will have to manually keep them consistent with actual JS code? I.e. when API is changed developer needs to remember to change/update corresponding type definition?

Would it be possible to put type definitions in https://github.com/DefinitelyTyped/DefinitelyTyped?

I wonder if we should actually gradually rewrite this driver in TypeScript.

nbransby:
@lutovich this manual process is typical, the benefit of generated type definitions alone doesn't usually suffice as a reason to rewrite a library in typescript.

Once completed, the type definitions won't need to be added to DefinitelyTyped because they will be included in the package and picked up by typescript automagically

lutovich:
@nbransby I think TypeScript could be valuable in general, not only for type definitions. Could we put type definitions in DefinitelyTyped only? I.e. forward this PR to DefinitelyTyped and not include it in this driver? What are the benefits of including type definitions in the project directly and not in DefinitelyTyped?

nbransby:
TypeScript  is super valuable especially If you come from a strongly typed language background such as Java.

The benefits are for the end user - they only have to npm install neo4j-javascript-driver, otherwise they would also need to npm install @types/neo4j-javascript-driver - separately hosted type definitions on DefinitelyTyped exist for libraries where authors arent willing to add definitions to the source.

agalazis:
+1 for this @SnappyCroissant  maybe it would be faster if you put it all together and do a PR at  
https://github.com/DefinitelyTyped/DefinitelyTyped

janwo:
Am I right that the definitions in result-summary.d.ts are outdated? The type `Integer` is used along the whole file whereas in the js code it is converted into a `int`?

lutovich:
@janwo code was converted to use `int` which is a function that returns `Integer`. Here is it's definition https://github.com/neo4j/neo4j-javascript-driver/blob/1.3.0/src/v1/integer.js#L863.

I will spend some time on type definitions this week. It seems adding type definitions to the driver code base is a good approach. DefinitelyTyped recommends doing it this way and type definitions can serve as additional "documentation" for the code. I'll rebase and cleanup the PR while preserving all current commits.

@SnappyCroissant, @pocesar to be able to merge your changes we will need a signed CLA from you. See http://neo4j.com/developer/cla/ for further information. Additional information on contributing code is found at http://neo4j.com/developer/contributing-code/.
Thanks a lot for your help!

janwo:
@lutovich This needs to be changed in the types as well than. Currently it is referenced as Integer, thus I can call all Integer methods at the moment.

pocesar:
@lutovich sent the CLA email

TimothyMischief:
Been on holiday only got the notification just now. I'll send through the CLA shortly!

UPDATE: Sent.

lutovich:
Closing this PR in favour of https://github.com/neo4j/neo4j-javascript-driver/pull/251

PR #184, opened 2017-01-03
Make sure session does not have bookmark after unsuccessful tx

lutovich:
Session's last bookmark should not be defined after explicit transaction fails or is rolled back.


oskarhane:
lgtm

PR #193, opened 2017-01-19
Improve transaction tests

lutovich:
This PR makes all tests in `transaction.test.js` more predictable by using promise callbacks and moves file to es6 syntax.


oskarhane:
Lgtm

Issue #201, opened 2017-02-05
Unit testing neo4j-driver errors


pito-svk:
I would like to unit test my function which takes Neo4jError as first argument for example.

I have a function to properly format neo4j driver errors:

```javascript
exports.formatErrors = () => {
  return (err, req, res, next) => {
    let error

    if (err.fields) {
      const singleError = err.fields[0]
      const errorName = _.last(singleError.code.split('.'))

      if (errorName === 'ConstraintValidationFailed') {
        const field = singleError.message.split('"')[1]
        error = new ValidationError(`${field} already exists`)
      }
    }

    if (!error) error = err

    next(error)
  }
}
```

The point is, that I would like to mock this error for example:

```
Structure {
  signature: 127,
  fields:
   [ { code: 'Neo.ClientError.Schema.ConstraintValidationFailed',
       message: 'Node 0 already exists with label User and property "email"=[john@doe.com]' } ] }
```

Is it possible to get Neo4jError class from this module and create error mock from this?

Thanks!

lutovich:
Hi @pito-svk,

I think `Neo4jError` is exported alongside functions like `driver()` and `auth()`. Following code works for me:

```js
var neo4j = require('neo4j-driver').v1;
var error = new neo4j.Neo4jError('Hi, I am an error', 'This is my error code');
console.log(error);
```

Do you think it is enough to test `formatErrors` function?

Thanks.

pito-svk:
What about struct with signature and fields? This is the error object I get when try to create node with uniqueness constraint.

lutovich:
@pito-svk I think the whole `Structure` object is exposed by mistake. What you see here is just a raw Bolt protocol message. We should construct a `Neo4jError` from it and pass it to the `catch()`/`onError()` callbacks. I'll create a task for us to fix this.

Thanks a lot for reporting the problem!

lutovich:
Hi @pito-svk, we've just released 1.1.1 version of the driver with the mentioned errors fix. Now all thrown errors should be of type `Neo4jError`. Hope the fix works for you, please feel free to create a new issue or update this one if it does not.

PR #203, opened 2017-02-15
Fix: Structure error has bad .toString()

legraphista:
Fixes issue where Structure.toString() outputs: `Structure(127, [[object Object]])`
Now it should properly stringify the error:  `Structure(127, [{"code":"Neo.ClientError.Statement.ArithmeticError","message":"/ by zero"}])`

### DIFF
--- a/src/v1/internal/packstream.js
+++ b/src/v1/internal/packstream.js
@@ -65,9 +65,9 @@ class Structure {
     let fieldStr = "";
     for (var i = 0; i < this.fields.length; i++) {
       if(i > 0) { fieldStr+=", " }
-      fieldStr += this.fields[i];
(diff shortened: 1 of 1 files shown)


legraphista:
I've noticed this issue when coupling a neo4j error with `boom.internal` from the [boom package](https://github.com/hapijs/boom) resulting in a bad error message (code found [here](https://github.com/hapijs/boom/blob/master/lib/index.js#L123-L125)).

lutovich:
@legraphista thanks for this PR! I think it is a good change. We'll also change errors to always be `Neo4jError` and not contain `Structure` thingy. See https://github.com/neo4j/neo4j-javascript-driver/issues/201.

PR #204, opened 2017-02-16
Unpack failure messages to Neo4jErrors

### LINKED ISSUE #201: Unit testing neo4j-driver errors

lutovich:
Previously raw FAILURE message was passed to `onError`-like callbacks and thrown as is. This means thrown error contained packstream related fields and actually was a packstream structure. Example:

```
Structure {
  signature: 127,
  fields: [
    {
      code: 'Neo.ClientError.Schema.ConstraintValidationFailed',
      message: 'Node 0 already exists with label A and property "a"=[42]'
    }
  ]
}
```

This commit makes sure we unpack/convert FAILURE messages to `Neo4jError`s and pass those further to callbacks.

Fixes #201

### DIFF
--- a/src/v1/internal/buf.js
+++ b/src/v1/internal/buf.js
@@ -22,7 +22,6 @@
   *(via Buffer API).
   */
 
-import {newError} from './../error';
 let _node = require("buffer");
 /**
   * Common base with default implementation for most buffer methods.
@@ -551,7 +550,7 @@ class NodeBuffer extends BaseBuffer {
         val.position + bytesToCopy );
       val.position += bytesToCopy;
     } else {
-      throw newError("Copying not yet implemented.");
+      super.putBytes(position, val);
     }
   };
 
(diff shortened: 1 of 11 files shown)


oskarhane:
Thanks for this! 
lgtm

PR #219, opened 2017-03-15
Add transaction with retries API

### LINKED ISSUE #207: Ability to read a transaction's state
It would be useful to be able to check what state a transaction is in (or at least whether it is still usable or not). For example if I do something like
```js
let tx = session.beginTransaction();
tx.run(...).then(result =>{
//dostuff
return tx.commit();
}).catch(err => {
return tx.rollback(); //How to tell if I need to do this
});
```
Then how to I tell if I need to rollback my transaction in the catch block - the error could have been caused by some code other than the run statement failing, so I can't assume the transaction has been rolled back. At the moment it seems like the only solution is to add an extra catch to the tx.rollback() call, but this seems a bit messy (and looks even more messy using async await).

lutovich:
3 main commits of this PR:
	
 * __Introduce transaction with retries API__
 This commit adds two new API functions:
 `Session#readTransaction()`
 `Session#writeTransaction()`
 Both take a single function as input. This function takes a single argument of type `Transaction` and returns a promise. It can be used to perform regular async operations like query execution. Introduced functions will commit/rollback transaction depending on the returned promise, so user code does not need to call `Transaction#commit()` explicitly. They also perform retries if given transaction fails with network errors (`ServiceUnavaliable` or `SessionExpired`) or with transient errors (`DeadlockDetected`, etc.). Retries are one with exponential backoff with initial delay of 1 second and total cap of 30 seconds.
 These API functions are useful to hide tolerable network problems and transient errors for both single and causal cluster deployments.
 Commit also fixes a problem in transaction error handling where `_onError` executed a rollback but did not properly wait for the returned promise to complete.

 * __TransactionExecutor checks if tx is active__
 To make sure it does not try to commit transaction that has already been committed or rolled back.
 This commit also adds a new API function `Transaction#isOpen()` which checks internal transaction state and returns `true` when transaction has not been committed/rolled back and `false` otherwise.

 * __Make maximum transaction retry time configurable__
 Commit makes driver aware of a new config property `maxTransactionRetryTime` to tune amount of time transactions executed via `Session#readTransaction(function(Transaction))` and `Session#writeTransaction(function(Transaction))` can be retried. It also adds JSDoc for `connectionPoolSize` property.

Fixes #207

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -164,23 +164,17 @@ gulp.task('test', function(cb){
 
 gulp.task('test-nodejs', ['install-driver-into-sandbox'], function () {
   return gulp.src('test/**/*.test.js')
-        .pipe(jasmine({
-            // reporter: new reporters.JUnitXmlReporter({
-            //   savePath: "build/nodejs-test-reports",
-            //   consolidateAll: false
-            // }),
-            includeStackTrace: true
-        }));
+    .pipe(jasmine({
+      includeStackTrace: true,
+      verbose: true
+    }));
 });
 
 gulp.task('test-boltkit', ['nodejs'], function () {
   return gulp.src('test/**/*.boltkit.it.js')
     .pipe(jasmine({
-      // reporter: new reporters.JUnitXmlReporter({
-      //   savePath: "build/nodejs-test-reports",
-      //   consolidateAll: false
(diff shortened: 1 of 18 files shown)


oskarhane:
Really nice! lgtm

PR #223, opened 2017-03-22
Do not retry transaction when it is terminated

lutovich:
This PR makes transaction functions `Session.readTransaction()` and `Session#writeTransaction()` not retry when executed transaction was explicitly terminated by the user. Such termination can happen when `Session#close()` is called or `killQuery` procedure is used.

Transaction termination can result in two different status codes on the server, both classified as transient errors which is not entirely correct. So we need additional status code checks in the retying code.


zhenlineo:
LGTM.

PR #230, opened 2017-04-19
Adding kerberos auth support

zhenlineo:


src/v1/index.js
```diff
@@ -38,6 +38,9 @@ const auth ={
       return {scheme: "basic", principal: username, credentials: password};
     }
   },
+    kerberos:(base64EncodedTicket)=> {
+    return {scheme:"kerberos", credentials: base64EncodedTicket};
```
lutovich, line 42: Formatting is a bit off here

lutovich: Please add tests for this method to avoid breaking with a typo or smth like it. We should also type check given `base64EncodedTicket` with `util.assertString()` function.

lutovich (changes requested):
@zhenlineo looks good but I think we need to add some type checking and tests. JS is really fragile wrt to typos :)

lutovich:
Closing in favour of https://github.com/neo4j/neo4j-javascript-driver/pull/236

PR #241, opened 2017-05-08
Added simple stress test

lutovich:
It executes read and write queries randomly against a given or default (`bolt://localhost`) URI. Queries are executed in both explicit and implicit transactions, both with and without bookmarks. Test can read environment variables to switch mode (fast or extended), turn on/off logging and change database URI. This makes it possible to run test against real cluster using `bolt+routing` scheme. New npm script is added for this `run-stress-tests`

### DIFF
--- a/test/v1/stress.test.js
+++ b/test/v1/stress.test.js
@@ -23,7 +23,7 @@ import parallelLimit from 'async/parallelLimit';
 import _ from 'lodash';
 import sharedNeo4j from '../internal/shared-neo4j';
 
-fdescribe('stress tests', () => {
+describe('stress tests', () => {
 
   const TEST_MODES = {
     fast: {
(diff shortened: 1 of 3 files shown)


lutovich:
Set of seen servers should be checked as well to make sure routing (if any) works.

test/v1/stress.test.js
```diff
@@ -0,0 +1,337 @@
...
+ * limitations under the License.
+ */
+
+import neo4j from '../../src/v1';
+import {READ, WRITE} from '../../src/v1/driver';
+import parallelLimit from 'async/parallelLimit';
+import _ from 'lodash';
+import sharedNeo4j from '../internal/shared-neo4j';
+
+fdescribe('stress tests', () => {
```
adnasa: typo?

lutovich: Oh yes :)
Thanks for spotting this.

PR #251, opened 2017-06-18
 TypeScript declaration files

### LINKED ISSUE #77: typescript definition file
Hi,

Is it possible to get a typescript definition file so this could be used in an Angular2 project?

Thanks

Alex

lutovich:
Added TypeScript declaration files for all public driver APIs. These should make it easier to use driver in TypeScript projects.

PR is heavily based on https://github.com/neo4j/neo4j-javascript-driver/pull/183. Thanks for contribution @SnappyCroissant and @pocesar.

Closed #77 

**Not ready for merge:** all declaration files should be covered with tests. Currently on `transaction.d.ts` is.


lutovich:
Hi @SnappyCroissant and @pocesar.
Could you guys please review this PR? I made some changes to your initial PR https://github.com/neo4j/neo4j-javascript-driver/issues/183. Most significant changes include:
 * removed definitions for non-public classes and functions
 * used interfaces as much as possible to define API shapes
 * moved type declarations to a dedicated folder "types"
 * started adding tests

@oskarhane could you please review as well? and share your general feeling about this?

oskarhane:
I'm positive, this is a good step. 
As far as reviewing I put my trust in @pocesar and @SnappyCroissant 😃

pocesar:
@lutovich looks good! I will eventually help with the generics (by using your own created interfaces to create safe code through the driver), but that's the cherry on top, the base definitions LGTM

EDIT: for example, when using subscribe, you'd want to have a `subscribe<T>(observer: Observer<T>)` so your records reflect whatever you pass in for the `T` generic, or a strongly typed generic for `Result`

lutovich:
@pocesar thanks for the review!
Actually, I removed generics from `Result`, `Record` and `Observer` on purpose. One can get back a `Record` of mixed types so I did not quite understand why should it have a generic signature. For example this code is valid:

```javascript
const result: Result = session.run("RETURN 42, "42", ["42"]");
const record: Record = result.records[0];
const value1: number = record.get(0);
const value2: string = record.get(1);
const value3: string[] = record.get(2);
```

Am I missing something?

pocesar:
in TS you can have union types, like `run<Type1 & Type2 & Type3>()`, so you can actually have one that has them all, all you can use `run<Type1 | Type2 | Type3>()` then after checking for their types, you get the correct type in a scope, like:

```ts
tx.run<Partial<Type1 & Type2 & Type3>>().records[0].// has members of Type1, 2 and 3, but if you use |, you can refine the type by checking for it's members
```

besides writing less code and not having to rely on variables, it makes composition easier (even more when dealing with async / promises code), so when you're writing wrappers around neo4j driver, and consuming in your code, it becomes much more evident, since you don't need to check the code to understand what will come out of the function / method (talking consumers / team members here)

check `Discriminated Unions` in https://www.typescriptlang.org/docs/handbook/advanced-types.html

TimothyMischief:
Everything within my understanding looks good. Discriminated Unions are a bit above my level of understanding but I'm educating myself now!

lutovich:
Nice, thanks for review!
I need to read more about TS type system as well.
Will update PR with more tests before merging.

cojack:
Im counting on you @lutovich

cojack (approved):

cojack:
@lutovich please don't forget to bump version and release it at npm :heart: 
Thanks in advance.

lutovich:
@cojack these should be released with 1.4.0-beta01 early next week. Pre-release versions are available with `next` npm tag. I.e. `npm install --save neo4j-driver@next`.

lutovich:
Pre-release version 1.4.0-beta01 is now available on npm, please feel free to try it out, give feedback and contribute.

cojack:
@lutovich I had a problem with explicitly property type definition.
When I try something like that:
```typescript
import * as Neo4j from 'neo4j-driver';
// stuff
private driver: Neo4j.v1.Driver;
```
Typescript said:
```
Error:(11, 18) TS2694:Namespace '"/home/cojack/Projects/neo4j-test/node_modules/neo4j-driver/types/v1/index"' has no exported member 'Driver'.
```

So I resigned with type annotation for this field, but I guess it should works.

lutovich:
@cojack non-default declaration does not export `Driver`. Could you please try:
```javascript
import Neo4j from 'neo4j-driver';
// stuff
private driver: Neo4j.Driver;
```

I'm not quite sure if we need to export **all** types from `v1/index.d.ts`. You could as well import `Driver` using `import {Driver} from "neo4j-driver/types/v1/driver";`.  Some input here is very welcome.

cojack:
@lutovich nope, it doesn't not work with your proposal. Anyway, doesn't have to :+1: 
Thanks for types :smiley:

Issue #218, opened 2017-03-14
Cannot check if error is instance of Neo4jError


GertSallaerts:
I would like to check if an error is an instance of `Neo4jError` but using ìnstanceof` does not work.

```js
const { Neo4jError } = require('neo4j-driver').v1;
const e = new Neo4jError('foo');
console.log(e instanceof Neo4jError); // false
```

This seems to be a known issue in babel. For now I've got a workaround that checks if there is a `code` property on the error that starts with `Neo.`.

This babel-transform might be a solution? https://github.com/loganfsmyth/babel-plugin-transform-builtin-extend/blob/master/src/index.js

GertSallaerts:
I also use this library on a few projects but I'm not sure how it will interact with babel: https://github.com/bjyoungblood/es6-error

lutovich:
Hi @Gertt,

Sorry for the delayed response. I tried using babel-plugin-transform-builtin-extend plugin and it seems to work in NodeJS tests but does not work for same tests in PhantomJS. Library es6-error does not seem to work in both.

There exists a standard `Error.name` property https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name. We do nothing with it at the moment but it is probably a good idea to set it to `Neo4jError`. D you think it is worthwhile to do this?

PR #254, opened 2017-06-22
fix neo4jerror name issue

praveenag:
`instanceof Neo4jError` will return false but you can use the property `name` instead to determine the error.
Fixes [218](https://github.com/neo4j/neo4j-javascript-driver/issues/218)

### DIFF
--- a/src/v1/error.js
+++ b/src/v1/error.js
@@ -44,6 +44,7 @@ class Neo4jError extends Error {
     super( message );
     this.message = message;
     this.code = code;
+    this.name = "Neo4jError"
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/v1/error.js
```diff
@@ -44,6 +44,12 @@ class Neo4jError extends Error {
     super( message );
     this.message = message;
     this.code = code;
+    Object.defineProperty(this, 'name', {
```
lutovich: Why not just `this.name = 'Neo4jError'`?

lutovich (commented):
@praveenag changes look good, one small question.

src/v1/error.js
praveenag: There was some reading on property definition vs assignment and by defining the property we are making it non enumerable.
http://2ality.com/2012/08/property-definition-assignment.html

lutovich: Yeah, code sets `enumerable: false`. All other properties are enumerable. Can we have `name` similar?

praveenag: Sure I will make the change and push it through.

lutovich (approved):

PR #265, opened 2017-07-08
Extract load balancing logic in a separate class

lutovich:
Previously load balancing logic for read and write servers was coupled with storage of server addresses in `RoundRobinArray` class. Such approach was problematic because it made hard to use different load balancing strategy.

This PR pulls load balancing in a separate class and removes `RoundRobinArray`. Instead addresses are stored in simple arrays and load balancing logic is applied to those arrays. Such approach makes it easy to plug a different load balancing strategy.


src/v1/internal/load-balancing-strategies.js
```diff
@@ -0,0 +1,72 @@
...
+   * Select next most appropriate writer from the list of given writers.
+   * @param {string[]} knownWriters an array of currently known writers to select from.
+   * @return {string} most appropriate writer or <code>null</code> if given array is empty.
+   */
+  selectWriter(knownWriters) {
+    throw new Error('Abstract function');
+  }
+}
+
+export class RoundRobinLoadBalancingStrategy extends LoadBalancingStrategy {
```
praveenag: Could we maybe extract this to a file of it's own. It has it's own purpose and it makes sense in imports to import it as 
```import {RoundRobinLoadBalancingStrategy} from './round-robin-load-balancing-strategy';```

praveenag (changes requested):

praveenag (approved):

PR #267, opened 2017-07-11
Least connected load balancing strategy

lutovich:
This replaces the existing round-robin. It gives us better performance with clusters composed of different machine capacities. Lease connected load balancing strategy selects start index in round robin fashion to avoid always selecting same machine when cluster does not have any running transactions or all machines have same number of active connections.

It is possible to go back to previous round-robin load balancing strategy using an experimental config setting:
```
  {loadBalancingStrategy: 'round_robin'}
```

### DIFF
--- a/test/v1/routing-driver.test.js
+++ b/test/v1/routing-driver.test.js
@@ -1,0 +1,42 @@
+/**
+ * Copyright (c) 2002-2017 "Neo Technology,","
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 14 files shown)


praveenag (approved):

test/v1/routing-driver.test.js
```diff
@@ -0,0 +1,42 @@
...
+  it('should create least connected when it is configured', () => {
+    const strategy = RoutingDriver._createLoadBalancingStrategy({loadBalancingStrategy: 'least_connected'}, new Pool());
+    expect(strategy instanceof LeastConnectedLoadBalancingStrategy).toBeTruthy();
+  });
+
+  it('should create round robin when it is configured', () => {
+    const strategy = RoutingDriver._createLoadBalancingStrategy({loadBalancingStrategy: 'round_robin'}, new Pool());
+    expect(strategy instanceof RoundRobinLoadBalancingStrategy).toBeTruthy();
+  });
+
```
praveenag: This is missing a test case for `'Unknown load balancing strategy`

praveenag (changes requested):

lutovich:
@praveenag test added

Issue #270, opened 2017-07-14
Integer.d.ts


janwo:
I just upgraded to the latest npm package and changed the importings accordingly. However...

    import Integer from "neo4j-driver/types/v1/integer";
    ...
    if (obj instanceof Integer) return Integer.toNumber(obj);

...results in a error...

    Cannot find module 'neo4j-driver/types/v1/integer

Why?

lutovich:
Hi @janwo,

I do not think you need to change imports. Those TypeScript declaration files are just type declarations, right? They do not contain any actual executable JavaScript code and modules. TypeScript knows how to extract types from NPM package and should use those defined in `types` folder. Your IDE should also just pick existing TypeScript declarations and make better code completion suggestions.

Hope this helps.

janwo:
@lutovich When I make no changes to my imports, I get this error message:

    Could not find a declaration file for module 'neo4j-driver/lib/v1/record'. 
    '/root/app/node_modules/neo4j-driver/lib/v1/record.js' implicitly has an 'any' type.
    `npm install @types/neo4j-driver/lib/v1/record` if it exists or add a new declaration (.d.ts) file containing `declare module 'neo4j-driver/lib/v1/record'`

Typescript 2.4

lutovich:
@janwo this definitely does not sound right. Could you please share an example project to reproduce this problem?

janwo:
@lutovich I´ve extracted an example: https://github.com/janwo/neo4j-test

`npm install && npm start`

lutovich:
@janwo I'm able to make your code compile by changing imports to:

```typescript
import neo4j from "neo4j-driver";
import {Driver} from "neo4j-driver/types/v1/driver";
import Integer from "neo4j-driver/types/v1/integer";
import Record from "neo4j-driver/types/v1/record";
import Transaction from "neo4j-driver/types/v1/transaction";
import Session from "neo4j-driver/types/v1/session";
import {AuthToken} from "neo4j-driver/types/v1";
```

and `DatabaseManager#get()` function to:

```typescript
public static get(): Driver {
    const authToken: AuthToken = neo4j.auth.basic("user", "pwd");
    const driver: Driver = neo4j.driver(`bolt://`, authToken);
    if(!DatabaseManager.neo4jClient) DatabaseManager.neo4jClient = driver;
    return this.neo4jClient;
}
```

[...]

PR #272, opened 2017-07-15
Wrap integer.d.ts with module

janwo:
Wrapping a module around integer to solve https://github.com/neo4j/neo4j-javascript-driver/issues/270.

### DIFF
--- a/types/v1/integer.d.ts
+++ b/types/v1/integer.d.ts
@@ -114,22 +114,25 @@ declare class Integer {
   static MAX_SAFE_VALUE: Integer;
 }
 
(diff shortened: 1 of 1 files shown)


pocesar:
this isn't correct, since you're mixing module declaration types and the rest of the files uses exports

lutovich:
@janwo I tried this change on your `neo4j-test` project and it does not seem to help. Am I missing something?

janwo:
@lutovich It did work for me. Whats the error?

lutovich:
@janwo never mind, I did not change the import. It works. I'm not entirely sure what exactly does this change mean. Maybe all declaration files should define modules like this?

@pocesar what do you think?

PR #280, opened 2017-08-07
Export all available types and interfaces

lutovich:
This PR makes index TypeScript declaration export possible user-facing types/interfaces so that they can be used without directly importing files form `types` folder.

Example:

```
import neo4j from "neo4j-driver";

const driver: neo4j.Driver = neo4j.driver("bolt://localhost");
const session: neo4j.Session = driver.session();
...
```

Fixes https://github.com/neo4j/neo4j-javascript-driver/issues/270.

@janwo, @pocesar your review would be super valuable :)

### DIFF
--- a/types/v1/index.d.ts
+++ b/types/v1/index.d.ts
@@ -17,14 +17,21 @@
  * limitations under the License.
  */
 
-import {inSafeRange, int, isInt, toNumber, toString} from "./integer";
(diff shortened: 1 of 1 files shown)


pocesar:
LGTM. this PR empowers library creators that use TS and neo4j internals 👍

PR #282, opened 2017-08-11
Added ability to enforce max connection lifetime

lutovich:
Driver keeps idle network connections in a connection pool. These connections can get invalidated while resting idle in the pool. They can be killed by network equipment like load balancer, proxy or firewall. It is also safer to refresh connections once in a while so that database can renew corresponding resources.

This PR adds `maxConnectionLifetime` setting and makes driver close too old connections. Checking and closing happens during connection acquisition.

Also removed manual `Date.now()` mocking in favour of existing library - Lolex.

### DIFF
--- a/test/v1/driver.test.js
+++ b/test/v1/driver.test.js
@@ -19,18 +19,22 @@
 
 import neo4j from '../../src/v1';
 import sharedNeo4j from '../internal/shared-neo4j';
+import FakeConnection from '../internal/fake-connection';
+import lolex from 'lolex';
 
 describe('driver', () => {
 
+  let clock;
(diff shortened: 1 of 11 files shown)


test/v1/driver.test.js
```diff
@@ -19,18 +19,22 @@
 
 import neo4j from '../../src/v1';
 import sharedNeo4j from '../internal/shared-neo4j';
+import FakeConnection from '../internal/fake-connection';
+import lolex from 'lolex';
```
praveenag, line 23: This is a good addition 👍

praveenag (approved):

PR #288, opened 2017-09-19
check if statement text is empty on Session.run and Transaction.run

ali-ince:


src/v1/session.js
```diff
@@ -65,6 +65,9 @@ class Session {
       statement = statement.text;
     }
     assertString(statement, 'Cypher statement');
+    if (statement.length == 0) {
+      throw new TypeError('Cypher statement is expected to be a non-empty string.');
```
lutovich: Maybe worth adding a `assertCypherStatement()` function in `util.js` that will verify given cypher is a non-empty string? It can be used in `session` and `transaction` just to avoid minor code duplication

lutovich (approved):
@ali-ince looks good! one minor comment

PR #289, opened 2017-09-20
implemented maxConnectionPoolSize logic

ali-ince:


src/v1/driver.js
```diff
@@ -59,7 +59,11 @@ class Driver {
       this._createConnection.bind(this),
       this._destroyConnection.bind(this),
       this._validateConnection.bind(this),
-      config.connectionPoolSize
+      {
```
lutovich, line 62: What do you think about adding a dedicated `PoolConfig` class for this object?
Then it will be possible to specify type of `config` constructor parameter in `Pool` in JSDoc and get autocompletion from Intellij.

src/v1/internal/pool.js
```diff
@@ -46,26 +52,35 @@ class Pool {
    * @return {object} resource that is ready to use.
    */
   acquire(key) {
-    let pool = this._pools[key];
-    if (!pool) {
-      pool = [];
-      this._pools[key] = pool;
+    let resource = this._acquire(key);
```
lutovich: Maybe change `let`s and `var`s to `const`s in this class, where possible.

src/v1/internal/util.js
```diff
@@ -114,6 +116,22 @@ function trimAndVerify(string, name, url) {
   return result;
 }
 
+function promiseOrTimeout(timeout, otherPromise, onTimeout) {
+  let timeoutPromise = new Promise((resolve, reject) => {
```
lutovich: Maybe use `const`

lutovich (commented):
@ali-ince changes look good. Made couple minor comments. Maybe also worth adding an integration test for max connections. All this can be done in a separate PR.

PR #291, opened 2017-09-22
Refactor boltstub, always run related tests

lutovich:
Previously driver allowed boltkit to be an optional dependency and relied on neokit (submodule) to start/stop neo4j database. That's why all tests that use stub server from boltkit were executed as a separate step. These days boltkit is always required and neokit is gone.

This commit refactors stub server handling and turns all boltkit tests into normal tests that run as part of `npm test` and `runTests` script. These tests are however skipped in browser testing because it is not possible to spawn child processes when in phantomjs environment.


test/v1/direct.driver.boltkit.test.js
```diff
@@ -19,30 +19,28 @@
...
-import boltkit from './boltkit';
-import sharedNeo4j from '../internal/shared-neo4j';
+import boltStub from '../internal/bolt-stub';
 
-describe('direct driver', () => {
+describe('direct driver with stub server', () => {
 
   it('should run query', done => {
-    if (!boltkit.BoltKitSupport) {
+    if (!boltStub.supported) {
```
ali-ince, line 27: What do you think about switching between describe or xdescribe based on the value of boltStub.supported before L24? Something like

```javascript
const fdescribe = boltStub.supported ? describe : xdesribe;

fdescribe('direct driver with stub server', () => ....
```

ali-ince (approved):
LGTM!

test/v1/direct.driver.boltkit.test.js
lutovich: Interesting point, if will save us from all these `boltStub.supported` checks. Let's maybe do it in a separate PR. I'll create a card.

PR #297, opened 2017-10-03
Deprecate connectionPoolSize config setting

lutovich:
In favour of `maxConnectionPoolSize`. Extracted pool config in a dedicated class which contains logic of handling now deprecated `connectionPoolSize` and new `maxConnectionPoolSize`. Added `maxConnectionPoolSize` to TypeScript declaration. Made code use `console.warn` instead of `console.log` for deprecation warnings. Updated test neo4j version to 3.2.5 and allowed stress test run a bit longer.

Currently default value of `maxConnectionPoolSize` is set to 50. This corresponds to previously default value of `connectionPoolSize`. We might tune this default in future.

### DIFF
--- a/src/v1/internal/pool-config.js
+++ b/src/v1/internal/pool-config.js
@@ -1,0 +1,82 @@
+/**
+ * Copyright (c) 2002-2017 "Neo Technology,","
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const DEFAULT_SIZE = 50;
+const DEFAULT_ACQUISITION_TIMEOUT = 60000;
+
+export default class PoolConfig {
+
+  constructor(maxIdleSize, maxSize, acquisitionTimeout) {
+    this.maxIdleSize = valueOrDefault(maxIdleSize, DEFAULT_SIZE);
+    this.maxSize = valueOrDefault(maxSize, DEFAULT_SIZE);
+    this.acquisitionTimeout = valueOrDefault(acquisitionTimeout, DEFAULT_ACQUISITION_TIMEOUT);
+  }
+
+  static defaultConfig() {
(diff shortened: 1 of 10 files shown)


ali-ince (commented):
Changes seem ok for me, but are we deprecating connectionPoolSize (i.e. maxIdleConnectionPoolSize in meaning)?

src/v1/internal/pool-config.js
```diff
@@ -0,0 +1,82 @@
...
+    if (maxIdleSizeConfigured && maxSizeConfigured) {
+      // both settings are configured - use configured values
+      maxIdleSize = config.connectionPoolSize;
+      maxSize = config.maxConnectionPoolSize;
+    } else if (!maxIdleSizeConfigured && maxSizeConfigured) {
+      // only maxSize is configured - use it's value for both
+      maxIdleSize = config.maxConnectionPoolSize;
+      maxSize = config.maxConnectionPoolSize;
+    } else if (maxIdleSizeConfigured && !maxSizeConfigured) {
+      // only maxIdleSize is configured - use it's value for both
```
zhenlineo: Shall we just ignore `maxIdle` rather than use it for `maxSize`?
As we changed the meaning of the config. Maybe people would be surprised to know their config meaning was changed.

src/v1/internal/pool-config.js
```diff
@@ -0,0 +1,82 @@
...
+  static defaultConfig() {
+    return new PoolConfig(DEFAULT_SIZE, DEFAULT_SIZE, DEFAULT_ACQUISITION_TIMEOUT);
+  }
+
+  static fromDriverConfig(config) {
+    const maxIdleSizeConfigured = isConfigured(config.connectionPoolSize);
+    const maxSizeConfigured = isConfigured(config.maxConnectionPoolSize);
+
+    if (maxIdleSizeConfigured) {
+      console.warn('WARNING: neo4j-driver setting "connectionPoolSize" is deprecated, please use "maxConnectionPoolSize" instead');
```
zhenlineo: WARNING: neo4j-driver setting "connectionPoolSize" is deprecated, the values set to this config will be ignored. To limit the maximum size of the pool, use `maxConnectionPoolSize` instead.

src/v1/internal/pool-config.js
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    let maxIdleSize;
+    let maxSize;
+
+    if (maxIdleSizeConfigured && maxSizeConfigured) {
+      // both settings are configured - use configured values
+      maxIdleSize = config.connectionPoolSize;
```
zhenlineo: I get same confused as @ali-ince 
I thought we just deprecate `maxIdle`, why we still have two values?
If we want to keep `maxIdle`, then we should also add `maxIdleConnectionPoolSize` into config, right? otherwise, we have no reason to keep both two values here, right?

lutovich:
@ali-ince @zhenlineo you are right, we should probably just point `connectionPoolSize` to `maxConnectionPoolSize` since it's name does not say anything about "idle" and remove `maxIdleSize` from the pool. I'll update the PR.

PR #298, opened 2017-10-04
Fix Driver.onCompleted callback

### LINKED ISSUE #234: driver.onCompleted not executing; driver.onError not executing too.
So, I've started using this driver after a few months of not using, and everything was crashing - it looked like driver is not instantating.
After some investigations I've discovered that my app was waiting for driver.onCompleted to execute, but it never executed.

However, if I would remove line of waiting driver.onCompleted to run, everything starts to work fine.
Same for driver.onError. I cannot detect whether credentials are bad or I can perform the requests properly.
IDK why this is happening as I've installed older (1.1.0 version) which was working fine with this API. Also I've used neo4j 3.0.x (cannot definitely recall the exact version) and then upgraded to 3.1.4, thinking that it can be newer driver-older db issue, still did not help.

```  
▶ node    
> const {v1: neo4j} = require('neo4j-driver')
undefined
> driver = neo4j.driver("bolt://localhost", neo4j.auth.basic("neo4j", "password1")); driver.onError = console.log.bind(console, 'err');
[Function: bound bound log]
> driver = neo4j.driver("bolt://localhost", neo4j.auth.basic("neo4j", "password")); driver.onCompleted = console.log.bind(console, 'completed');
[Function: bound bound log]
> 
```

lutovich:
Setting this callback will now trigger connectivity verification. Provided callback function will be invoked if verification succeeds. When `onCompleted` callback is not set no verification will happen. This is achieved using getter/setter on `onCompleted` property.

This callback has previously been broken when session was decoupled from connection.

Removed `Driver.onCompleted` and `Driver.onError` from README. These callbacks were introduced specifically for neo4j-browser that needs to verify connectivity and be notified about broken WebSockets. So they were always a "special" not fully thought through part of driver's public API. It is preferable to do connectivity verification by running a dummy query like `RETURN 1`. Note that it's not always required.

Fixes #234

### DIFF
--- a/src/v1/driver.js
+++ b/src/v1/driver.js
@@ -24,6 +24,7 @@ import StreamObserver from './internal/stream-observer';
 import {newError, SERVICE_UNAVAILABLE} from './error';
 import {DirectConnectionProvider} from './internal/connection-providers';
 import Bookmark from './internal/bookmark';
+import ConnectivityVerifier from './internal/connectivity-verifier';
 
 const READ = 'READ', WRITE = 'WRITE';
 /**
@@ -72,6 +73,21 @@ class Driver {
      * @protected
      */
     this._connectionProvider = null;
+
+    this._onCompleted = null;
+  }
+
+  get onCompleted() {
+    return this._onCompleted;
+  }
+
+  set onCompleted(callback) {
+    this._onCompleted = callback;
+    if (this._onCompleted) {
+      const connectionProvider = this._getOrCreateConnectionProvider();
+      const connectivityVerifier = new ConnectivityVerifier(connectionProvider, this._onCompleted);
+      connectivityVerifier.verify();
+    }
   }
 
   /**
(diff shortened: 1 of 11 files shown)


src/v1/driver.js
```diff
@@ -72,6 +73,21 @@ class Driver {
...
     this._connectionProvider = null;
+
+    this._onCompleted = null;
+  }
+
+  get onCompleted() {
+    return this._onCompleted;
+  }
+
+  set onCompleted(callback) {
```
ali-ince, line 84: Setting an event handler invoking actual verification confused me a bit. What happens if verification fails? Maybe an explicit method `verifyConnectivity()` returning a promise would be a better option, can be elaborated for upcoming versions?

ali-ince (commented):
looks good. One comment though about a confusion.

src/v1/driver.js
lutovich: Failed verification will result in `Driver.onCompleted` not being invoked. Instead `Driver.onError` will be invoked. I agree, these callbacks are not ideal and rather ugly. That's why they are removed from README. Users can verify connectivity by running a dummy query like `RETURN 1`. So this fix is only for backwards compatibility.

ali-ince (approved):

PR #307, opened 2017-11-18
use neo4j.v1 in browser code snippet

johnymontana:


lutovich:
Thanks @johnymontana!

PR #313, opened 2017-12-18
Fixed timeout when acquiring connection from pool

### LINKED ISSUE #304: Cannot read property 'filter' of undefined
Using version 1.5.0
Node 8.1.4
Neo4j 3.1.1 (singel instance) & 3.2.3 (Casual cluster)
Linux / Debian

A couple of time per day we get this error, since it happens outside the normal process chain we're unable to catch it and the stack trace doesn't point to anything obvious.

```
TypeError: Cannot read property 'filter' of undefined
    at node_modules/neo4j-driver/lib/v1/internal/pool.js:115:44
    at Timeout._onTimeout (node_modules/neo4j-driver/lib/v1/internal/util.js:150:9)
    at ontimeout (timers.js:488:11)
    at tryOnTimeout (timers.js:323:5)
    at Timer.listOnTimeout (timers.js:283:5)
    at listOnTimeoutNT (timers.js:359:26)
    at _combinedTickCallback (internal/process/next_tick.js:99:11)
    at process._tickDomainCallback (internal/process/next_tick.js:198:9)
```

lutovich:
Connection pool has limited size and returns `Promise<Connection>` instead of `Connection` directly. Returned promise can be failed when connection acquisition timeout has fired. Pool uses helper function `util#promiseOrTimeout()` to race acquisition and timeout promises.

This function allowed both timeout callback to be invoked and connection to be acquired. It cleared the timeout outside of the raced promises. As result connection was acquired but timeout callback failed in background. It tried to clear non-existing pending acquisition attempts.

This PR fixes the problem by including `#clearTimeout()` call in the chain of raced promises.

Fixes #304

### DIFF
--- a/src/v1/internal/pool.js
+++ b/src/v1/internal/pool.js
@@ -162,10 +162,10 @@ class Pool {
     // check if there are any pending requests
     const requests = this._acquireRequests[key];
     if (requests) {
-      var pending = requests.shift();
+      const pending = requests.shift();
 
       if (pending) {
-        var resource = this._acquire(key);
+        const resource = this._acquire(key);
         if (resource) {
           pending.resolve(resource);
 
--- a/src/v1/internal/util.js
+++ b/src/v1/internal/util.js
@@ -17,7 +17,7 @@
  * limitations under the License.
  */
 
-import { newError } from "../error";
+import {newError} from '../error';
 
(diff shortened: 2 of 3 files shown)


ali-ince (approved):
:+1:

PR #317, opened 2017-12-29
IPv6 support

lutovich:
PR makes driver able to connect to Neo4j databases that listen on IPv6 addresses. Routing driver will now be able to receive IPv6 addresses in routing procedure responses and resolve host names to IPv6 addresses.

Parsing of URLs was the biggest piece here. New dependency on [url-parse](https://github.com/unshiftio/url-parse) is added for this.


lutovich:
@oskarhane could you please take a look at this PR?

It adds a new runtime dependency, which I assume is fine and will not cause any problems for users. Driver will use it's own copy of url-parse. Is my understanding correct?

When this change is merged and released we should consider using it in neo4j-browser.

ali-ince (approved):
Looks good!

Issue #245, opened 2017-05-22
Handling integers


hierozbosch:
The handling of integers is referenced in previous issues (153, 122, 106, etc.).  These issues have been closed and point to the Bolt driver methods (InSafeRange, int, etc.) as the resolution.  However, using these methods implies hard-coding conversions against a known set of data entities, and doing it on the server.  It's a highly inefficient solution:  the driver should only convert 64-bit numbers to {high: low:} objects as a last resort.  Converting all integers into these objects and then requiring a server to locate and re-convert most of them back into JS numbers is not an ideal solution.  I'm using the neo4j-javascript-driver for a web app, and I would be very surprised if serious 64-bit applications considered JS their language of choice.  In my case:  a.)  all my numbers are JS numbers, both in and out of neo4j;  b.) I do not want to use my back-end neo4j server and front-line web servers to convert JS integers into objects and then convert those objects back into JS integers; and c.) I do not want to hard-code my server or client scripts to the granularities of a data model.  I'm raising this as a new issue:  I think the neo4j-JS-driver should have an optional directive to convert 64-bit integers directly to JS numbers and not waste time converting them all into objects.  Thanks!

lutovich:
Hi @hierozbosch,

Thanks for reporting this.

I feel that converting to `Integer` objects as a last resort might be strange. Some numbers will then be regular javascript numbers and some will be `Integer`s which could be even more confusing. You are right saying that we could have a switch like `{useNativeNumbersOnly: boolean}` but what should driver do if it receives a 64-bit number from the database? It is possible for multiple applications written in different languages to talk to the same database. Stored numbers could have any number of bits in such case.

All those mentioned issues and this one definitely show that `Integer` class with `{high: ..., log: ...}` does not feel very native and convenient. We'll think about improving number handing in upcoming releases.

hierozbosch:
Thanks Lutovich, Just by way of background, I've been running on neo4j for 5 years, since v 0.7.  We use the graph for supply chain management--relationship tracking, access control, and social media stuff.  We deal in smallish numbers--JS dates are probably the only time we exceed the ^32 range, and we don't see a case where we'd get over 2^53 with this app.  

I put off the upgrade to neo4j 3.x for over a year due to the breakages, but I'm deep in it now.  I had to change a lot of cypher, and almost all the result-handling.  That's when the issue with the integers came up.  

[...]

Issue #225, opened 2017-03-30
Properties(node) with integer values returned with high/low object


fancellu:
e.g.

```match (n) return properties(n), labels(n), id(n)```

So, for a node with properties: name: "Dino", age: 50, I'll get back the following JS object

```
{
          "name": "Dino",
          "age": {
            "low": 50,
            "high": 0
          }
        }```

Is there a way to get Neo4J to return something less surprising? 

The browser does this when displaying the value. If not, is there a recommended way to take the returned
object and recurse through it turning it back to a JS integer?

The properties could of course be N layers deep, and I just want to JSON.stringify() on it, 
but the high/low stuff looks odd.

Thanks

lutovich:
Hi @fancellu,

The problem here is that Neo4j database supports wider integer type than JS. So to make sure large integer property, that is stored in the database, can be safely (without loss of precision) retrieved using the JS driver we had to introduce `Integer` class. This is the thing you see with `high` & `low` properties. A bit more info about this can be found here in README: https://github.com/neo4j/neo4j-javascript-driver#a-note-on-numbers-and-the-integer-type.

You could use `Record#toObject()` function and then recursively traverse the returned object checking properties with `neo4j.isInt()` function and then reassigning them with result of `Integer.toString()`.

Following sample code:

```javascript
it('transform integers to strings', () => {
  const object = {
    name: 'Dino',
    age: int(50),
    info: {
      salary: int(1000),
      car: {
        brand: 'BMW',
        price: int(1000)
      }
    }
  };

  console.log(object);
  transform(object);
  console.log(object);
});

function transform(object) {
  for (let property in object) {
    if (object.hasOwnProperty(property)) {
      const propertyValue = object[property];
      if (isInt(propertyValue)) {
        object[property] = propertyValue.toString();
      } else if (typeof propertyValue === 'object') {
        transform(propertyValue);
      }
    }
  }
}
```

outputs this before transformation:

```json
{"name":"Dino","age":{"low":50,"high":0},"info":{"salary":{"low":1000,"high":0},"car":{"brand":"BMW","price":{"low":1000,"high":0}}}}
```

and this after transformation:

```json
{"name":"Dino","age":"50","info":{"salary":"1000","car":{"brand":"BMW","price":"1000"}}}
```

note the use of `isInt()` function.

Hope this helps.

fancellu:
Perfect. That code snippet could do with going in the docs. Very useful.

[...]

Issue #122, opened 2016-08-25
Strange behavior with integer number literals being returned as objects {high: 23, low: 0}


adrianseeley:
```
package: "neo4j-driver": "^1.1.0-M01"
node: v6.2.1
```

I'm creating a new node in Neo4J using the following from NodeJS, noting that the property 'datapoints' is a number literal:

``` neo4j
CREATE (a:pipeline)
    SET a.uuid       = {uuid},
        a.type       = "datasource",
        a.datapoints = 0,
        a.normalized = {normalized},
        a.headers    = {headers},
        a.types      = {types},
        a.t_created  = {now}
RETURN a
```

Now when I go to lookup that node using the following statement I get different results from the Neo4J browser and NodeJS driver:

``` neo4j
MATCH (a:pipeline)
RETURN a
```

If I run this statement from the Neo4J browser I get a response as expected where datapoints is indeed a number:

``` json
{
  "records": [
    {
      "keys": [
        "a"
      ],
      "length": 1,
      "_fields": [
        {
          "identity": "37225",
          "labels": [
            "pipeline"
          ],
          "properties": {
            "headers": [
              "dt",
              "h"
            ],
            "datapoints": "0",
            "types": [
              "Date",
              "Float"
            ],
            "normalized": false,
            "type": "datasource",
            "t_created": 1472143439959,
            "uuid": "1472143439959.562064"
          },
          "id": "37225"
        }
      ],
      "_fieldLookup": {
        "a": 0
      }
    }
  ],
  "summary": {
    "statement": {
      "text": "MATCH (a:pipeline)\nRETURN a",
      "parameters": {}
    },
    "statementType": "r",
    "updateStatistics": {
      "_stats": {
        "nodesCreated": 0,
        "nodesDeleted": 0,
        "relationshipsCreated": 0,
        "relationshipsDeleted": 0,
        "propertiesSet": 0,
        "labelsAdded": 0,
        "labelsRemoved": 0,
        "indexesAdded": 0,
        "indexesRemoved": 0,
        "constraintsAdded": 0,
        "constraintsRemoved": 0
      }
    },
    "plan": false,
    "profile": false,
    "notifications": []
  },
  "responseTime": 316
}
```

However when I run the same statement using the NodeJS package I get an unexpected object for datapoints:

``` json
{
   "records": [
      {
         "keys": [
            "a"
         ],
         "length": 1,
         "_fields": [
            {
               "identity": {
                  "low": 37225,
[...]

PR #323, opened 2018-01-17
Introduce 'disableLosslessIntegers' config option

### LINKED ISSUE #322: Make the library more javascript friendly
One of the principles of a js client library should be to make it easier to use the underlying platfrom (in this case cypher/neo4j) from a js/nodejs application. I think there are a number of design patterns in this lib which make it quite impractical

- Handling of integers -
Has already been raised on numerous occasions #245 #225 #122 
Edge cases shouldn't make the default behavior impractical

- Response format - currently returns a representation of a table
Was referred in #189 and the resolution provided for this issue should IMO be the default format that the promise returns.
As a consumer inside a javascript/nodejs application I would expect a json representation of the data.

This forces me to map the result in _100 pc_ of the cases - which IMO is one of the jobs that a client library should handle.

I don't mean to suggest that the other cases should be ignored, but these cases which are used less frequently and only in advanced/edge scenarios, should not be the default behavior.

lutovich:
This boolean option allows users to make driver accept and represent integer values as native JavaScript `Number`s instead of driver's special `Integer`s. When this option is set to `true` driver will fail to encode `Integer` values passed as query parameters. All returned `Record`, `Node`, `Relationship`, etc. will have their integer fields as native `Number`s. Object returned by `Record#toObject()` will also contain native numbers.

**Warning:** enabling this setting can potentially make driver return lossy integer values. This would be the case when database contain integer values which do not fit in `[Number.MIN_SAFE_INTEGER, Number.MAX_SAFE_INTEGER]` range. Such integers will then be represented as `Number.NEGATIVE_INFINITY` or `Number.POSITIVE_INFINITY`, which is lossy and different from what is actually stored in the database. That is why this option is set to `false` by default. Driver's `Integer` class is able to represent integer values beyond `[Number.MIN_SAFE_INTEGER, Number.MAX_SAFE_INTEGER]` and thus is a much safer option in the general case.

Fixes #322 
Related to #245, #225, #122

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "neo4j-driver",
-  "version": "1.5.0-dev",
+  "version": "1.6.0-dev",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/v1/index.js
+++ b/src/v1/index.js
@@ -157,6 +157,17 @@ const USER_AGENT = "neo4j-javascript/" + VERSION;
  *       // result in no timeout being applied. Connection establishment will be then bound by the timeout configured
  *       // on the operating system level. Default value is 5000, which is 5 seconds.
  *       connectionTimeout: 5000, // 5 seconds
+ *
+ *       // Make this driver always return and accept native JavaScript number for integer values, instead of the
+ *       // dedicated {@link Integer} class. Values that do not fit in native number bit range will be represented as
+ *       // <code>Number.NEGATIVE_INFINITY</code> or <code>Number.POSITIVE_INFINITY</code>. Driver will fail to encode
+ *       // {@link Integer} values passed as query parameters when this setting is set to <code>true</code>.
+ *       // <b>Warning:</b> It is not safe to enable this setting when JavaScript applications are not the only ones
+ *       // interacting with the database. Stored numbers might in such case be not representable by native
+ *       // {@link Number} type and thus driver will return lossy values. For example, this might happen when data was
+ *       // initially imported using neo4j import tool and contained numbers larger than
+ *       // <code>Number.MAX_SAFE_INTEGER</code>. Driver will then return positive infinity, which is lossy.
+ *       useNativeNumbers: false,
  *     }
  *
  * @param {string} url The URL for the Neo4j database, for instance "bolt://localhost"
--- a/src/v1/index.js
+++ b/src/v1/index.js
@@ -158,16 +158,17 @@ const USER_AGENT = "neo4j-javascript/" + VERSION;
  *       // on the operating system level. Default value is 5000, which is 5 seconds.
  *       connectionTimeout: 5000, // 5 seconds
  *
- *       // Make this driver always return and accept native JavaScript number for integer values, instead of the
+ *       // Make this driver always return native JavaScript numbers for integer values, instead of the
  *       // dedicated {@link Integer} class. Values that do not fit in native number bit range will be represented as
- *       // <code>Number.NEGATIVE_INFINITY</code> or <code>Number.POSITIVE_INFINITY</code>. Driver will fail to encode
- *       // {@link Integer} values passed as query parameters when this setting is set to <code>true</code>.
- *       // <b>Warning:</b> It is not safe to enable this setting when JavaScript applications are not the only ones
+ *       // <code>Number.NEGATIVE_INFINITY</code> or <code>Number.POSITIVE_INFINITY</code>.
+ *       // <b>Warning:</b> It is not always safe to enable this setting when JavaScript applications are not the only ones
  *       // interacting with the database. Stored numbers might in such case be not representable by native
- *       // {@link Number} type and thus driver will return lossy values. For example, this might happen when data was
+ *       // {@link Number} type and thus driver will return lossy values. This might also happen when data was
  *       // initially imported using neo4j import tool and contained numbers larger than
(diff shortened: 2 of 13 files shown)


lutovich:
I'm unsure about two things:

1. Is `useNativeNumbers` a good name for this setting? @nigelsmall could you please assess this?
2. What to do with TypeScript declaration files? This setting basically changes what kind of type driver uses for integer values. So TS type info depends on this setting. Things like:
    ```typescript
    declare class Node {
      identity: Integer;
      ...
    }
    ```

    will have to change somehow to represent this. Maybe it's enough to use union types like this:
    ```typescript
    declare class Node {
      identity: Integer | number;
      ...
    }
    ```
    and add comments? @pocesar, @TimothyDespair, @cojack could you please suggest?

ali-ince:
As per your points;

1. I would prefer `forceNativeNumbers` because of its emphasise on forcing something that could go wrong (be lossy).

2. AFAIK, we can either specify `any` or `union types` as you've already suggested. Going with union types feel the way to go.

ali-ince (commented):
Changes look good to me.

What would you think about adding an `toObject()` overload which will convert `Integer` instances to `number` types although Driver is not configured to use native numbers?

oskarhane (approved):
This looks good to me as well. 

And I agree with the changes that if you opt out of using the integer type when receiving results you should also not send integer types as params.

oskarhane:
btw, I'm +1 on @ali-ince param name suggestion `forceNativeNumbers`

sarincasm:
agree with what @ali-ince has suggested - this config option could also be received by `Record.toObject()` (and maybe also by `Record.get()`)

Lets say majority of my cases can be handled by JS Numbers and in a few cases I need the `Integer` -

constructor -
`neo4j.driver("bolt://localhost", neo4j.auth.basic("neo4j", "neo4j"), {forceJSNumbers: true})`
handle edge case with the same driver
`record.toObject({forceJSNumbers: false})`

If, its the other way around,

constructor -
`neo4j.driver("bolt://localhost", neo4j.auth.basic("neo4j", "neo4j"))`
handle smaller cases with the same driver
`record.toObject({forceJSNumbers: true})`

also, the param could specify that we are forcing JS Numbers. Native might be confusing as in - native to javascript or native to neo4j - hence `forceJSNumbers`

cojack (approved):

pocesar:
@lutovich you could make it a generic, so the person decides what type it should use, as such:

```ts
declare class Node<T = Integer> {
  identity: T;
  // ...
}
```

that means, if the person decides to not explicitly decide which type `Node` should be, it will be `Integer` (the default argument). If the person wants something else, he can then `const node: Node<number>` or `type MyNode = Node<number>`, and the types will be infered

lutovich:
@pocesar thanks a lot for your reply! The only problem I see with `Node<T = Integer>` is that users will be able to use `Node<string>` and try to access `identity` as string, while it's in fact `Integer | number`. 

Here is a possible solution:

```typescript
declare class Node<T extends (Integer | number) = Integer> {
  identity: T;
  // ...
}
```

it's essentially the same thing but adds an additional generic type bound. So no code change for existing users, since `Integer` is still the default generic type. Users who enable JS numbers will have to use `Node<number>`.

What do you think?

lutovich:
PR updated and should now be ready for a final review. Changes:

 * renamed `useNativeNumbers` to `disableLosslessIntegers`
 * lifted restriction on storing `Integer` when `disableLosslessIntegers=true` because it allows for better backwards compatibility
 * updated TS declarations to support `number` instead of `Integer`

See commit messages for more details.

@eelsweb we've decided to only expose `disableLosslessIntegers` config on the driver level for now. Having it on both driver and record level makes things more complicated. What if it's set to `true` on the driver level but user calls `record.toObject({disableLosslessIntegers: false})`? It's definitely implementable but weird.

This change will hopefully be merged in 1.6 and go through alpha, beta and RC stages. Feedback is very valuable!

pocesar:
@lutovich if the `identity` member will ALWAYS be either `Integer` or `number`, then it's the best choice to do that 👍 
just a nit-pick, you can also explicitly decouple that from the declaration by creating a new type, like 
```ts
declare type NumberInteger = Integer | number;
declare class Node<T extends NumberInteger = Integer> {
}
```
so it can be reused elsewhere

sarincasm:
@lutovich this kind of over-riding is actually quite standard practice

For instance - [Google's Nodejs Lib](https://github.com/google/google-api-nodejs-client) for their APIs

View the section [Options](https://github.com/google/google-api-nodejs-client#options)

> You may specify additional options either in the global google object or on a service client basis.

> You can specify an auth object to be used per request. Each request also inherits the options specified at the service level and global level.

For this neo4j library, consider a case where 80 pc of my functions work with `disableLosslessIntegers=true`, but there are some cases where i need `Integer`, so I create a driver with `{disableLosslessIntegers:true}` and for the remaining few cases, I don't need a separate driver, I can simply override that by calling `toObject({disableLosslessIntegers: false})` since local config takes precedence over global options.

lutovich:
@eelsweb is this an actual use-case you have or just a theoretical concern?

Imho `disableLosslessIntegers` is a good first step. It solves the problem many people complained about. Overrides on the `Record` level are definitely doable but will make API more complicated. It is now unclear if they are needed or solve real problems. That is why I think we can defer them for future, at least right now.

sarincasm:
Agreed 😄

PR #330, opened 2018-02-23
Fixed timeout when acquiring connection from the pool

### LINKED ISSUE #304: Cannot read property 'filter' of undefined

lutovich:
Previously it was possible for timeout to kick in after connection acquisition promise is resolved, but before timeout itself is cleared.

Pseudocode would be:

```
const acquisitionTimeout = xxx;

const f1 = () => {/* fail and clear the acquisition attempt */};
const acquisitionTimeoutId = setTimeout(f1, acquisitionTimeout);

const f2 = () => clearTimeout(acquisitionTimeoutId);
connectionAcquisitionPromise.then(f1);
```

This code allows `connectionAcquisitionPromise` to get resolved and `f1` to get executed before `f2`. It is possible because installed callbacks are only scheduled, not executed when promise is resolved.

This PR fixes the problem by moving `clearTimeout` to the place that resolves the acquisition request, i.e. `Pool#release()` call.

Fixes #304


ali-ince (approved):
:+1:

PR #335, opened 2018-03-09
Bolt V2 with point support

lutovich:


src/v1/internal/packstream-v1.js
```diff
@@ -504,17 +506,88 @@ class Unpacker {
   }
 
   _unpackStructWithSize(size, buffer) {
-    let signature = buffer.readUInt8();
-    let mapper = this.structMappers[signature];
-    if (mapper) {
-      return mapper(this, buffer);
+    const signature = buffer.readUInt8();
```
ali-ince: Does it make sense if we add size checks in existing struct unpacking logic?

src/v1/spatial-types.js
```diff
@@ -0,0 +1,57 @@
...
+export class Point {
+
+  /**
+   * @constructor
+   * @param {number|Integer} srid the coordinate reference system identifier.
+   * @param {number} x the <code>x</code> coordinate of the point.
+   * @param {number} y the <code>y</code> coordinate of the point.
+   * @param {number|undefined} z the <code>y</code> coordinate of the point or <code>undefined</code> if point has 2 dimensions.
+   */
+  constructor(srid, x, y, z) {
```
ali-ince: If we have `p1 = new Point(1, 1, 0)`, a 2D point after construction and later on do something like `p1.z = 4` we'll end up with a 3D point. I think it would be better if we could keep point's dimensions immutable, basically deciding it in constructor and not exposing it as a public property. Does it make sense?

types/v1/spatial-types.d.ts
```diff
@@ -0,0 +1,37 @@
...
+import {NumberOrInteger} from './graph-types';
+import Integer from "./integer";
+
+declare class Point<T extends NumberOrInteger = Integer> {
+  srid: T;
+  x: number;
+  y: number;
+  z: number | undefined;
+
+  constructor(srid: T, x: number, y: number, z?: number | undefined)
```
ali-ince: Having `z?`, do we still need `| undefined`?

ali-ince (commented):
Changes look good. There's already a build error related to `TypeScript` definitions.

I made a couple of small comments.

lutovich:
@ali-ince thanks for the review! Comments should now be addressed.

PR #341, opened 2018-03-27
Support for temporal types

lutovich:
This PR makes driver able to receive temporal types as query results and send them as query parameters. All types are represented by custom date-time classes because native `Date` type is quite limited. It can't represent needed millisecond range and does not have support for things like timezone.

New types are:
 * `Duration` - represents `Duration` Cypher type
 * `LocalTime` - represents `LocalTime` Cypher type
 * `Time` - represents `Time` Cypher type
 * `Date` - represents `Date` Cypher type
 * `LocalDateTime` - represents `LocalDateTime` Cypher type
 * `DateTimeWithZoneOffset` - represents `DateTime` Cypher type with time zone as offset in seconds
 * `DateTimeWithZoneOffset` - represents `DateTime` Cypher type with time zone as ID

All temporal types expose understandable properties. For example `Date` has `year`, `month` and `day` properties. Driver uses code adapted from https://github.com/ThreeTen/threetenbp to perform the conversions. Main difference is that driver operates on `Integer` objects and not native JavaScript numbers.

Temporal types also override `toString()` to return ISO 8601 strings.

### DIFF
--- a/src/v1/internal/packstream-v2.js
+++ b/src/v1/internal/packstream-v2.js
@@ -19,13 +19,56 @@
 
 import * as v1 from './packstream-v1';
 import {isPoint, Point} from '../spatial-types';
+import {
+  CypherDate,
+  CypherDateTimeWithZoneId,
+  CypherDateTimeWithZoneOffset,
+  CypherDuration,
+  CypherTime,
+  isCypherDate,
+  isCypherDateTimeWithZoneId,
+  isCypherDateTimeWithZoneOffset,
+  isCypherDuration,
+  isCypherLocalDateTime,
+  isCypherLocalTime,
+  isCypherTime
+} from '../temporal-types';
 import {int} from '../integer';
+import {
+  cypherDateToEpochDay,
+  cypherLocalDateTimeToEpochSecond,
+  cypherLocalTimeToNanoOfDay,
+  epochDayToCypherDate,
+  epochSecondAndNanoToCypherLocalDateTime,
+  nanoOfDayToCypherLocalTime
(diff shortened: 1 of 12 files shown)


src/v1/internal/packstream-v2.js
```diff
@@ -121,3 +192,215 @@ function unpackPoint3D(unpacker, structSize, buffer) {
...
+  const months = unpacker.unpack(buffer);
+  const days = unpacker.unpack(buffer);
+  const seconds = unpacker.unpack(buffer);
+  const nanoseconds = unpacker.unpack(buffer);
+
+  return new CypherDuration(months, days, seconds, nanoseconds);
+}
+
+function packLocalTime(value, packer, onError) {
+  const totalNanos = cypherLocalTimeToNanoOfDay(value);
```
ali-ince: Maybe name this as `nanoOfDay` instead of `totalNanos`?

src/v1/internal/packstream-v2.js
```diff
@@ -121,3 +192,215 @@ function unpackPoint3D(unpacker, structSize, buffer) {
...
+}
+
+/**
+ * Pack given cypher time.
+ * @param {CypherTime} value the time value to pack.
+ * @param {Packer} packer the packer to use.
+ * @param {function} onError the error callback.
+ */
+function packTime(value, packer, onError) {
+  const totalNanos = cypherLocalTimeToNanoOfDay(value.localTime);
```
ali-ince: same here?

ali-ince (approved):
LGTM. 

Only one very small comment on variable naming.

PR #345, opened 2018-04-05
Flatten all temporal types

lutovich:
And unit tests for temporal-util.

### DIFF
--- a/src/v1/internal/packstream-v2.js
+++ b/src/v1/internal/packstream-v2.js
@@ -24,22 +24,22 @@ import {
   DateTimeWithZoneId,
   DateTimeWithZoneOffset,
   Duration,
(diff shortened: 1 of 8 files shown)


ali-ince (approved):
LGTM.

I've only one question. What's the motivation behind adding `string` as a possible input type to the functions in `temporal-util.js`? Does it provide the user to pass in a map like this as a temporal value?

```
{
  year: "1990",
  month: "12",
  day: "5"
}
```

lutovich:
@ali-ince thanks for the review!
Internally all arguments are converted to `Integer` using `int()` function. It supports strings as parameters. That is why JSDocs mention strings. `temporal-util` is not public and public docs say nothing about strings.

PR #350, opened 2018-04-11
Support number props in spatial and temporal types

lutovich:
By default, driver exposes integer values as instances of `Integer` class. This class can represent integer values larger than `Number.MAX_SAFE_INTEGER` without loss of precision. Driver can be forced to return native numbers using `disableLosslessIntegers` config option.

This PR makes spatial and temporal types support `disableLosslessIntegers`. All exposed integer properties will be of type `number` when this config option is set to `true`. Internal conversions for temporal types still use `Integer` to not lose precision.


ali-ince (approved):
👍

PR #352, opened 2018-04-11
Represent Cypher DateTime with a single type

lutovich:
DateTime contains time zone information. It might either be an offset in seconds (that corresponds to ±HH:MM) or the time zone id (like 'Europe/London'). Previously driver exposed DateTime as two different types: `DateTimeWithZoneOffset` and `DateTimeWithZoneId`.

This PR combines them into a single type:

```
class DateTime {
  year: Integer|number,
  month: Integer|number,
  day: Integer|number,
  hour: Integer|number,
  minute: Integer|number,
  second: Integer|number,
  nanosecond: Integer|number,
  timeZoneOffsetSeconds: Integer|number,
  timeZoneId: string
}
```

where either `timeZoneOffsetSeconds` or `timeZoneId` is defined. Such single type better corresponds to the Cypher DateTime type and feels more idiomatic to JavaScript. Shape of objects (set of their properties) feels more important than the class/type.

Also renamed `Time.offsetSeconds` to `Time.timeZoneOffsetSeconds` to better represent the meaning of the property.

### DIFF
--- a/src/v1/index.js
+++ b/src/v1/index.js
@@ -30,22 +30,7 @@ import {assertString, isEmptyObjectOrNull} from './internal/util';
 import urlUtil from './internal/url-util';
 import HttpDriver from './internal/http/http-driver';
 import {isPoint, Point} from './spatial-types';
-import {
-  Date,
-  DateTimeWithZoneId,
-  DateTimeWithZoneOffset,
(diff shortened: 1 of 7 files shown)


ali-ince (approved):
👍

Issue #340, opened 2018-03-25
Unclear error on bad input


linusmartensson:
Hey guys! Love your work!

I was running a query and accidentally passed a string into the query parameters, rather than the actual parameters. Instead of a descriptive error message, all I got was "error: koa-onerror Neo4jError: Connection was closed by server" after a query timeout - which wasn't too helpful at 0430am. :dancer: 
Could we get an input type check and an "erroneous query parameter type" or something when passing bad input on the client side? :1st_place_medal:

lutovich:
Hi @linusmartensson,

Thank you for raising this issue. It definitely makes sense to validate params in the driver. Otherwise, it's a protocol violation and database closes connection. We will most likely fix this problem for the upcoming 1.6 release.

PR #354, opened 2018-04-16
More validation for query parameters

lutovich:
Driver expects query parameters to either be undefined/null or an object. This was previously not enforced and illegal parameters, like strings or arrays, were sent to the database. This resulted in a protocol violation and database closed the connection. Users were only able to see the actual error/stacktrace in the database logs. Driver received a `ServiceUnavailable` or `SessionExpired` error.

This PR adds validation of query parameters. It also prohibits nodes, relationships, and paths from being used as query parameters in the driver.

Resolves https://github.com/neo4j/neo4j-javascript-driver/issues/340

### DIFF
--- a/src/v1/internal/util.js
+++ b/src/v1/internal/util.js
@@ -39,7 +39,7 @@ function isEmptyObjectOrNull(obj) {
 }
 
 function isObject(obj) {
-  return typeof obj === 'object' && !Array.isArray(obj) && Boolean(obj);
+  return typeof obj === 'object' && !Array.isArray(obj) && obj !== null;
 }
 
 /**
@@ -79,7 +79,7 @@ function assertCypherStatement(obj) {
 }
 
 function assertQueryParameters(obj) {
-  if (!isObject(obj) && Boolean(obj)) {
+  if (!isObject(obj)) {
     // objects created with `Object.create(null)` do not have a constructor property
     const constructor = obj.constructor ? ' ' + obj.constructor.name : '';
     throw new TypeError(`Query parameters are expected to either be undefined/null or an object, given:${constructor} ${obj}`);
--- a/src/v1/internal/util.js
+++ b/src/v1/internal/util.js
@@ -39,8 +39,29 @@ function isEmptyObjectOrNull(obj) {
 }
 
 function isObject(obj) {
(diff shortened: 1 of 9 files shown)


src/v1/internal/util.js
```diff
@@ -52,10 +73,17 @@ function assertString(obj, objName) {
...
   assertString(obj, 'Cypher statement');
-  if (obj.trim().length == 0) {
+  if (obj.trim().length === 0) {
     throw new TypeError('Cypher statement is expected to be a non-empty string.');
   }
-  return obj;
+}
+
+function assertQueryParameters(obj) {
+  if (!isObject(obj) && Boolean(obj)) {
```
ali-ince: `isObject(obj)` already includes the condition `Boolean(obj)`, maybe just remove `&& Boolean(obj)` here?

src/v1/internal/util.js
```diff
@@ -39,8 +39,29 @@ function isEmptyObjectOrNull(obj) {
 }
 
 function isObject(obj) {
-  const type = typeof obj;
-  return type === 'function' || type === 'object' && Boolean(obj);
+  return typeof obj === 'object' && !Array.isArray(obj) && Boolean(obj);
```
ali-ince: I don't know an immediate replacement but not sure if `Boolean(obj)` is the correct thing to check here. Here are a couple of console tests;

`isObject(new Date())` returns `true`
`isObject(new Number())` returns `true`

Do you think we can have a better way of checking for actual object types?

ali-ince (commented):
Changes look good to me. Only added a couple of comments.

src/v1/internal/util.js
lutovich: I will replace `Boolean(obj)` with `obj !== null` which is much less cryptic.

Technically, `Date` and `Number` are objects so maybe it's okay to accept them. We should probably support parameters given like this:

```javascript
class PersonParams {
  constructor(name, age) {
    this.name = name;
    this.age = age;  
  }
}

session.run('CREATE (:Peson {name: $name, age: $age})', new PersonParams('Bob', 42))
```

and that is why I'm not sure how to allow these but disallow `Date` and `Number`.

PR #355, opened 2018-04-16
Improve exports of spatial and temporal types

lutovich:
Previously driver exported all spatial and temporal types and functions from both default and non-default export in v1 index.js file.

This PR makes it export all spatial and temporal types only from the `types` constant. Functions are still exported from both default and non-default exports to be consistent with existing `isInt`. Created two grouping constants `spatial` and `temporal` to export corresponding functions in a nicer, grouped way.

### DIFF
--- a/src/v1/index.js
+++ b/src/v1/index.js
@@ -235,7 +235,7 @@ const error = {
 };
 
 /**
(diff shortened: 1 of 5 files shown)


oskarhane (approved):
LGTM

PR #356, opened 2018-04-17
Added Point.toString()

lutovich:
It returns a string representation of the `Point` value object. All coordinates are formatted with at least one decimal place to emphasize they are not integers. So coordinate value `42` will be formatted as `42.0`. Optional `z` coordinate is included only if it is defined.


oskarhane (approved):
Perfect, thank you!

PR #361, opened 2018-04-26
Omit zero nanoseconds in `toString()` for temporal types

lutovich:
This changes implementations of `toString()` function for `Duration`, `Time`, `LocalTime`, `LocalDateTime` and `DateTime`. Returned strings still remain valid ISO strings and represent same temporal values.


ali-ince (approved):

oskarhane:
Thank you for this!

PR #371, opened 2018-05-13
Automatically use secure WebSocket on HTTPS pages

### LINKED ISSUE #255: https connection error
Hi,

I tried to connect the neo4j from my application it worked well while an application running in **http** protocol, but while changing the protocol to **https**. I am getting the following error.

`neo4j-web.js:29960 WebSocket connection to 'ws://neo4j.host.io:7687/' failed: Connection closed before receiving a handshake response`

I know we need to connect through wss while in https. But don't know how to configure the driver to access https

how to pass **opts.encrypted= true**, can't it switch automatically passed on the application protocol. something like //expample.com

```
    //Allow boolean for backwards compatibility
    if (opts.encrypted === true || opts.encrypted === _util.ENCRYPTION_ON) {
      if (!opts.trust || opts.trust === "TRUST_CUSTOM_CA_SIGNED_CERTIFICATES") {
        scheme = "wss";
      } else {
        this._error = (0, _error.newError)("The browser version of this driver only supports one trust " + "strategy, 'TRUST_CUSTOM_CA_SIGNED_CERTIFICATES'. " + opts.trust + " is not supported. Please " + "either use TRUST_CUSTOM_CA_SIGNED_CERTIFICATES or disable encryption by setting " + "`encrypted:\"" + _util.ENCRYPTION_OFF + "\"` in the driver configuration.");
        return;
      }
    }
    this._url = scheme + "://" + opts.host + ":" + opts.port;

```

lutovich:
Previously, driver used in browser used secure WebSocket with 'wss' scheme only when encryption was explicitly turned on in the driver config. On HTTP web pages it is perfectly fine to use either plane or secure WebSocket ('ws' and 'wss' respectively). On HTTPS web pages only 'wss' is allowed. So explicit `{encrypted: true}` configuration was always required for drivers used on HTTPS web pages.

This PR makes driver automatically use secure WebSocket when driver is used on a HTTPS web page.

Resolves #255

### DIFF
--- a/src/v1/internal/ch-websocket.js
+++ b/src/v1/internal/ch-websocket.js
@@ -228,21 +228,23 @@ function asWindowsFriendlyIPv6Address(scheme, parsedUrl) {
  * @return {{scheme: string|null, error: Neo4jError|null}} object containing either scheme or error.
  */
 function determineWebSocketScheme(config, protocolSupplier) {
-  const encrypted = config.encrypted;
+  const encryptionOn = isEncryptionExplicitlyTurnedOn(config);
+  const encryptionOff = isEncryptionExplicitlyTurnedOff(config);
   const trust = config.trust;
+  const secureProtocol = isProtocolSecure(protocolSupplier);
+  verifyEncryptionSettings(encryptionOn, encryptionOff, secureProtocol);
 
-  if (encrypted === false || encrypted === ENCRYPTION_OFF) {
+  if (encryptionOff) {
     // encryption explicitly turned off in the config
     return {scheme: 'ws', error: null};
   }
 
-  const protocol = typeof protocolSupplier === 'function' ? protocolSupplier() : '';
-  if (protocol && protocol.toLowerCase().indexOf('https') >= 0) {
+  if (secureProtocol) {
     // driver is used in a secure https web page, use 'wss'
     return {scheme: 'wss', error: null};
   }
 
-  if (encrypted === true || encrypted === ENCRYPTION_ON) {
+  if (encryptionOn) {
     // encryption explicitly requested in the config
     if (!trust || trust === 'TRUST_CUSTOM_CA_SIGNED_CERTIFICATES') {
       // trust strategy not specified or the only supported strategy is specified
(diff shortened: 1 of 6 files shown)


src/v1/internal/ch-websocket.js
```diff
@@ -233,4 +222,46 @@ function asWindowsFriendlyIPv6Address(scheme, parsedUrl) {
   return `${scheme}://${ipv6Host}:${parsedUrl.port}`;
 }
 
+/**
+ * @param {ChannelConfig} config - configuration for the channel.
+ * @param {function(): string} protocolSupplier - function that detects protocol of the web page.
+ * @return {{scheme: string|null, error: Neo4jError|null}} object containing either scheme or error.
+ */
+function determineWebSocketScheme(config, protocolSupplier) {
```
ali-ince, line 230: Based on the information given in _Security Considerations_ section [here](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_client_applications), it may make sense generating warning message about mixing protocols may not work - if encryption is explicitly turned off or on (i.e. https/ws or http/wss).

src/v1/internal/ch-websocket.js
```diff
@@ -233,4 +222,46 @@ function asWindowsFriendlyIPv6Address(scheme, parsedUrl) {
...
+
+  const protocol = typeof protocolSupplier === 'function' ? protocolSupplier() : '';
+  if (protocol && protocol.toLowerCase().indexOf('https') >= 0) {
+    // driver is used in a secure https web page, use 'wss'
+    return {scheme: 'wss', error: null};
+  }
+
+  if (encrypted === true || encrypted === ENCRYPTION_ON) {
+    // encryption explicitly requested in the config
+    if (!trust || trust === 'TRUST_CUSTOM_CA_SIGNED_CERTIFICATES') {
```
ali-ince: This is totally out of scope of this PR, but I find `TRUST_CUSTOM_CA_SIGNED_CERTIFICATES` setting to be confusing and would prefer `TRUST_SYSTEM_CA_SIGNED_CERTIFICATES` as the default. We may consider revising naming / behaviour in upcoming releases?

ali-ince (approved):
LGTM along with a couple of comments.

src/v1/internal/ch-websocket.js
lutovich: Added a warn print in https://github.com/neo4j/neo4j-javascript-driver/pull/371/commits/c623d8a93962d0633c53c8deb0d9536f918aa883

src/v1/internal/ch-websocket.js
lutovich: I'll create a card for this

ali-ince (approved):

PR #380, opened 2018-05-30
Support configurable logging

lutovich:
This PR makes driver able to log internal events to the configured logging destination. Produced log messages should give an insight into what operations are happening inside the driver. Logging level and destination can be configured for the driver in the config object:

```javascript
const config = {
  logging: {
    level: 'info',
    logger: (level, message) => console.log(level + ' ' + message)
  }
};

const driver = neo4j.driver(
 'bolt://localhost',
  neo4j.auth.basic('neo4j', 'password'),
  config
);
```

Property `level` should be a string. Currently supported levels are: 'error', 'warn', 'info' and 'debug'. Levels have the following priorities:

```
error: 0
warn: 1
info: 2
debug: 3
```

and enabling a certain level also enables all levels with lower priorities.

Property `logger` should be a function that takes two arguments `level: string` and `message: string`. Log level can now be 'error', 'warn', 'info' or 'debug'. Function should not execute any blocking or long-running operations because it is often invoked on a hot path.

There exists a predefined logger implementation which outputs timestamp, log level and message. Example:

```
1527703465096 INFO Direct driver 0 created for server address localhost:7687
```

It is defined in the `neo4j.logging` object and can be used like this:

```javascript
const config = {
  logger: neo4j.logging.console('info') // optional level parameter
};

const driver = neo4j.driver(
 'bolt://localhost',
  neo4j.auth.basic('neo4j', 'password'),
  config
);
```

Users of the driver should be able to plug in an existing logging framework like https://github.com/winstonjs/winston.

Also fixed double closing of connections in `Driver#close()` and added current timestamp to the `RoutingTable#toString()`.

### DIFF
--- a/src/v1/driver.js
+++ b/src/v1/driver.js
@@ -238,11 +238,17 @@ class Driver {
   close() {
     this._log.info(`Driver ${this._id} closing`);
 
-    for (let connectionId in this._openConnections) {
-      if (this._openConnections.hasOwnProperty(connectionId)) {
-        this._openConnections[connectionId].close();
-      }
+    try {
+      // purge all idle connections in the connection pool
       this._pool.purgeAll();
+    } finally {
+      // then close all connections driver has ever created
+      // it is needed to close connections that are active right now and are acquired from the pool
+      for (let connectionId in this._openConnections) {
+        if (this._openConnections.hasOwnProperty(connectionId)) {
+          this._openConnections[connectionId].close();
+        }
+      }
     }
   }
 }
--- a/src/v1/driver.js
+++ b/src/v1/driver.js
@@ -26,6 +26,7 @@ import {DirectConnectionProvider} from './internal/connection-providers';
 import Bookmark from './internal/bookmark';
 import ConnectivityVerifier from './internal/connectivity-verifier';
 import PoolConfig, {DEFAULT_ACQUISITION_TIMEOUT, DEFAULT_MAX_SIZE} from './internal/pool-config';
+import Logger from './internal/logger';
 
 const DEFAULT_MAX_CONNECTION_LIFETIME = 60 * 60 * 1000; // 1 hour
 
@@ -42,6 +43,8 @@ const READ = 'READ';
  * @type {string}
  */
 const WRITE = 'WRITE';
+
+let idGenerator = 0;
 
 /**
  * A driver maintains one or more {@link Session}s with a remote
@@ -66,17 +69,19 @@ class Driver {
   constructor(hostPort, userAgent, token = {}, config = {}) {
     sanitizeConfig(config);
 
+    this._id = idGenerator++;
     this._hostPort = hostPort;
     this._userAgent = userAgent;
-    this._openSessions = {};
-    this._sessionIdGenerator = 0;
+    this._openConnections = {};
     this._token = token;
     this._config = config;
+    this._log = Logger.create(config);
     this._pool = new Pool(
       this._createConnection.bind(this),
       this._destroyConnection.bind(this),
       this._validateConnection.bind(this),
-      PoolConfig.fromDriverConfig(config)
+      PoolConfig.fromDriverConfig(config),
+      this._log
     );
 
     /**
@@ -87,6 +92,15 @@ class Driver {
     this._connectionProvider = null;
 
     this._onCompleted = null;
+
+    this._afterConstruction();
+  }
+
+  /**
+   * @protected
+   */
+  _afterConstruction() {
+    this._log.info(`Direct driver ${this._id} created for server address ${this._hostPort}`);
   }
 
   /**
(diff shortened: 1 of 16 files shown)


src/v1/internal/connector.js
```diff
@@ -348,7 +365,9 @@ class Connection {
       return;
     }
 
-    log('C', 'ACK_FAILURE');
+    if (this._log.isEnabled()) {
```
oskarhane: Are these `.isEnabled()` calls really needed?
They create some noise and makes it harder to read when most of these functions only have a single row of logic.

Wouldn't a noop logger just do nothing with the level calls anyway?

src/v1/internal/logger.js
```diff
@@ -0,0 +1,126 @@
...
+   */
+  static noOp() {
+    return noOpLogger;
+  }
+
+  /**
+   * Check if logger is enabled, i.e. it is not a no-op implementation.
+   * @return {boolean} <code>true</code> when logger is not a no-op, <code>false</code> otherwise.
+   */
+  isEnabled() {
```
oskarhane: See prev comment about this. Is this needed?

src/v1/driver.js
```diff
@@ -224,11 +236,19 @@ class Driver {
...
-    for (let sessionId in this._openSessions) {
-      if (this._openSessions.hasOwnProperty(sessionId)) {
-        this._openSessions[sessionId].close();
-      }
+    this._log.info(`Driver ${this._id} closing`);
+
+    try {
+      // purge all idle connections in the connection pool
       this._pool.purgeAll();
+    } finally {
```
oskarhane, line 244: Why are these changes part of this PR?

test/internal/connection-providers.test.js
```diff
@@ -137,7 +138,7 @@ describe('LoadBalancer', () => {
   it('initializes routing table with the given router', () => {
     const connectionPool = newPool();
     const loadBalancingStrategy = new LeastConnectedLoadBalancingStrategy(connectionPool);
-    const loadBalancer = new LoadBalancer('server-ABC', {}, connectionPool, loadBalancingStrategy, NO_OP_DRIVER_CALLBACK);
+    const loadBalancer = new LoadBalancer('server-ABC', {}, connectionPool, loadBalancingStrategy, NO_OP_DRIVER_CALLBACK, Logger.noOp());
```
oskarhane, line 141: What if the logger argument isn't defined, does it break?

test/internal/connector.test.js
```diff
@@ -140,7 +141,7 @@ describe('connector', () => {
 
   it('should convert failure messages to errors', done => {
     const channel = new DummyChannel.channel;
-    connection = new Connection(channel, 'bolt://localhost');
+    connection = new Connection(channel, 'bolt://localhost', Logger.noOp());
```
oskarhane, line 144: What if the logger argument isn't defined, does it break?

oskarhane (commented):
I think this is a very welcome feature that'll help with the quality of reported issues and more.

Just a few comments.

src/v1/internal/connector.js
lutovich: These checks prevent code from generating a message string when logging is disabled. For example:

```javascript
this._log.debug(`${this} S: RECORD ${JSON.stringify(msg)}`);
```

will stringify a record and create a message string with `this` and record content even if `this._log` is a no-op logger. The `isEnabled()` check makes code not even allocate all those strings.

This is an optimization that might be relevant in the connection code because it is very hot. I did not use it in other places that are not performance-critical.

Does this make sense?

src/v1/driver.js
lutovich: I noticed a double-close of connections because of logging and decided to fix it in-place :)
Will move it to a different PR.

test/internal/connector.test.js
lutovich: Yes, this test breaks. This is expected, right?

test/internal/connection-providers.test.js
lutovich: Yes, `LoadBalancer` tests in `connection-providers.test.js` break. This is expected, right?

src/v1/driver.js
lutovich: Extracted two separate commits - one for `RoutingTable#toString()` change and one about this `Driver#close()`

src/v1/internal/connector.js
oskarhane: This makes sense, great!

src/v1/driver.js
oskarhane: 👍

test/internal/connection-providers.test.js
oskarhane: Maybe. Or maybe a noop function could be the default value. 
Otoh, these are internal stuff that can have breaking changes, so I guess this is totally fine.

ali-ince (commented):
What do you think about bringing a feature that makes log level configurable and filters messages based on the configured level?

I believe it will better align with other logging frameworks and also will have less internal overhead for unwanted log calls (compared to having the user implemented his own filtering logic).

lutovich:
@ali-ince this is definitely a valuable thing to have. Updated the PR with it.

ali-ince (approved):

PR #390, opened 2018-06-19
Allow to create temporal objects from standard JS Date

lutovich:
This PR adds `#fromStandardDate()` functions to all temporal types except `Duration`. Such functions allow to create temporal objects from the provided standard JavaScript `Date`.

### DIFF
--- a/src/v1/internal/temporal-util.js
+++ b/src/v1/internal/temporal-util.js
@@ -278,9 +278,9 @@ export function totalNanoseconds(standardDate, nanoseconds) {
 }
 
 /**
- * Get the total number of nanoseconds from the given standard JavaScript date.
+ * Get the time zone offset in seconds from the given standard JavaScript date.
  * @param {global.Date} standardDate the standard JavaScript date.
- * @return {number} the total amount of nanoseconds.
+ * @return {number} the time zone offset in seconds.
  */
 export function timeZoneOffsetInSeconds(standardDate) {
   return standardDate.getTimezoneOffset() * SECONDS_PER_MINUTE;
--- a/src/v1/internal/temporal-util.js
+++ b/src/v1/internal/temporal-util.js
@@ -17,7 +17,7 @@
  * limitations under the License.
  */
 
-import {int} from '../integer';
+import {int, isInt} from '../integer';
 import {Date, LocalDateTime, LocalTime} from '../temporal-types';
 
 /*
@@ -35,6 +35,7 @@ const MINUTES_PER_HOUR = 60;
 const SECONDS_PER_MINUTE = 60;
 const SECONDS_PER_HOUR = SECONDS_PER_MINUTE * MINUTES_PER_HOUR;
 const NANOS_PER_SECOND = 1000000000;
+const NANOS_PER_MILLISECOND = 1000000;
 const NANOS_PER_MINUTE = NANOS_PER_SECOND * SECONDS_PER_MINUTE;
 const NANOS_PER_HOUR = NANOS_PER_MINUTE * MINUTES_PER_HOUR;
 const DAYS_0000_TO_1970 = 719528;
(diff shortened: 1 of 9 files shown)


src/v1/temporal-types.js
```diff
@@ -168,6 +203,21 @@ export class Date {
...
+   * Hour, minute, second, millisecond and time zone offset components of the given date are ignored.
+   * @param {global.Date} standardDate the standard JavaScript date to convert.
+   * @return {Date} new date.
+   */
+  static fromStandardDate(standardDate) {
+    verifyStandardDateAndNanos(standardDate, null);
+
+    return new Date(
+      standardDate.getFullYear(),
+      standardDate.getMonth(),
```
ali-ince: I think there's a problem about conversion of `month` value from/to standard date as javascript `Date` object stores months as `0` based values - with `0` = `January` and `11` = `December`. It seems to me that we do not consider this difference in conversions.

Maybe add value range checks on temporal type constructors?

src/v1/internal/temporal-util.js
```diff
@@ -264,6 +265,27 @@ export function dateToIsoString(year, month, day) {
...
+ * @return {Integer|number} the total amount of nanoseconds.
+ */
+export function totalNanoseconds(standardDate, nanoseconds) {
+  nanoseconds = (nanoseconds || 0);
+  const nanosFromMillis = standardDate.getMilliseconds() * NANOS_PER_MILLISECOND;
+  return isInt(nanoseconds) ? nanoseconds.add(nanosFromMillis) : nanoseconds + nanosFromMillis;
+}
+
+/**
+ * Get the total number of nanoseconds from the given standard JavaScript date.
```
ali-ince: It seems that the comment is outdated.

src/v1/temporal-types.js
```diff
@@ -211,6 +261,26 @@ export class LocalDateTime {
...
+   * @param {global.Date} standardDate the standard JavaScript date to convert.
+   * @param {Integer|number|undefined} nanosecond the optional amount of nanoseconds.
+   * @return {LocalDateTime} new local date-time.
+   */
+  static fromStandardDate(standardDate, nanosecond) {
+    verifyStandardDateAndNanos(standardDate, nanosecond);
+
+    return new LocalDateTime(
+      standardDate.getFullYear(),
+      standardDate.getMonth(),
```
ali-ince: Same here.

src/v1/temporal-types.js
```diff
@@ -261,6 +331,27 @@ export class DateTime {
...
+   * @param {global.Date} standardDate the standard JavaScript date to convert.
+   * @param {Integer|number|undefined} nanosecond the optional amount of nanoseconds.
+   * @return {DateTime} new date-time.
+   */
+  static fromStandardDate(standardDate, nanosecond) {
+    verifyStandardDateAndNanos(standardDate, nanosecond);
+
+    return new DateTime(
+      standardDate.getFullYear(),
+      standardDate.getMonth(),
```
ali-ince: Same here.

test/v1/temporal-types.test.js
```diff
@@ -658,6 +659,147 @@ describe('temporal-types', () => {
     expect(() => new neo4j.types.DateTime(1, 2, 3, 4, 5, 6, 7, 8, 'UK')).toThrow();
   });
 
+  it('should convert standard Date to neo4j LocalTime', () => {
```
ali-ince, line 662: Maybe add some integration tests that incorporate standard date -> temporal type conversions?

ali-ince (approved):

oskarhane (approved):
Good stuff, I have nothing to add!

ivan-kleshnin:
Any reason to not apply this to native JS dates automatically?

lutovich:
@ivan-kleshnin the problem is that driver can't know which Neo4j type should the native JS date be converted to. It can be converted to either `LocalDateTime` or `DateTime` but driver can't know if time zone information should be preserved or not. Technically, native JS dates can also get converted to `Date`, `LocalTime` or `Time` by using a subset of available getters. Only use `getFullYear()`, `getMonth()` and `getDate()` to construct Neo4j Date from a native date, ignore `getHours()`, etc.

That's why this PR adds explicit conversion methods so application code can tell driver which Neo4j temporal type is required.

Does this make sense to you?

ivan-kleshnin:
@lutovich yes, thanks. That what I thought, but as a Neo4j noob wanted to ensure :)

PR #393, opened 2018-06-21
Value range checks for temporal types

lutovich:
Added checks for all temporal types to assert that values used to create them are in expected numeric ranges. This should disallow negative months, days more than 31 and things like that.

Based on https://github.com/neo4j/neo4j-javascript-driver/pull/390, only last commit is new.


ali-ince (approved):
👍

PR #397, opened 2018-07-20
Rollback transaction after failure in transaction functions

lutovich:
Previously, when transaction function failed with a retryable error transaction was not explicitly rolled back. This could cause driver not return connections back to the pool.

This PR fixes the problem by adding an explicit rollback after a failed transaction function.


ali-ince (approved):
👍

PR #399, opened 2018-08-20
Replace ACK_FAILURE with RESET

lutovich:


src/v1/internal/connector.js
```diff
@@ -347,50 +341,37 @@ class Connection {
...
-    if (this._ackFailureMuted) {
-      return;
-    }
-
-    if (this._log.isDebugEnabled()) {
-      this._log.debug(`${this} C: ACK_FAILURE`);
-    }
-
-    const observer = {
+  _resetOnFailure() {
```
ali-ince, line 350: Wdyt about calling `resetAndFlush` instead, like `resetAndFlush().finally(() => { /* clear failure */ })`?

ali-ince (approved):
Changes look good for me. I only had one small comment.

src/v1/internal/connector.js
lutovich: It would definitely make the code a bit more concise but then failure will be cleared in a tick after the one that processes the RESET response. **I think** it might result in some queued callbacks to be failed unnecessarily.

PR #400, opened 2018-08-22
Refactoring to better support new protocol versions

lutovich:
- Extract protocol negotiation into a dedicated class. It encapsulates all the logic related to the Bolt protocol negotiation and versioned packstream components
- Connection is now less coupled with protocol messages. It is able to write an arbitrary `RequestMessage` instead of exposing functions to write messages of a specific type, like `Connection#pullAll()`. Also introduced a `BoltProtocol` abstraction which exposes only logical operations, like "initialize the connection" or "run a query". It should make it simpler to introduce new protocol version that uses different messages to execute same logical operations.
- Expose transaction operations on BoltProtocol. Make it expose logical operations for explicit transactions like begin, commit and rollback. These operations require different set of messages in Bolt V3. Having operations exposed like this should make it easier to override them.


src/v1/internal/protocol-handshaker.js
```diff
@@ -0,0 +1,115 @@
...
+    this._channel.write(newHandshakeBuffer());
+  }
+
+  /**
+   * Read and interpret the Bolt handshake response from the given buffer.
+   * @param {BaseBuffer} buffer byte buffer containing the handshake response.
+   * @return {BoltProtocol} bolt protocol corresponding to the version suggested by the database.
+   * @throws {Neo4jError} when bolt protocol can't be instantiated.
+   */
+  readHandshakeResponse(buffer) {
```
ali-ince: Would it be better to name this to something like `createNegotiatedProtocol`? Wdyt?

src/v1/internal/protocol-handshaker.js
```diff
@@ -0,0 +1,115 @@
...
+    const proposedVersion = buffer.readInt32();
```
ali-ince: and this maybe to `negotiatedVersion`?

src/v1/transaction.js
```diff
@@ -236,16 +228,20 @@ let _states = {
...
+      } else {
+        return connection.protocol().rollbackTransaction(observer);
+      }
+    })
+    .catch(error => observer.onError(error));
 
   // for commit & rollback we need result that uses real connection holder and notifies it when
   // connection is not needed and can be safely released to the pool
-  return new Result(observer, msg, {}, emptyMetadataSupplier, connectionHolder);
+  return new Result(observer, '', {}, emptyMetadataSupplier, connectionHolder);
```
ali-ince: Does it make sense to use something other than `''` here - just for clarity?

ali-ince (approved):
LGTM, only had a couple of comments about naming.

PR #412, opened 2018-09-14
Allow rollback transaction after failure

lutovich:
Previously, it was not possible to rollback transaction after failure. Doing this resulted in an error. This PR turns it into a no-op. Transaction is automatically rolled back after failure in the database because failure is acknowledged by sending a RESET message. So this is just a driver API level change.


ali-ince (approved):
LGTM.

PR #414, opened 2018-09-20
Improve packaging to better support module bundlers

### LINKED ISSUE #396: v1 driver cannot resolve internal modules
Installed most recent stable version of neo4j-driver today and when I try to launch my app I get the following error message: `Module not found: Can't resolve 'dns' in '/app/node_modules/neo4j-driver/lib/v1/internal'` (due note I'm using create-react-app and running the command `npm run start`)

I'll look into this issue and see if its just a bad publish or something else. Other recommendations would be helpful though.

lutovich:
- Replace local copy of [text-encoding](https://github.com/inexorabletash/text-encoding) library with a dependency
- Split Node and browser components into separate modules and export them from `src/v1/internal/node/index.js` and `src/v1/internal/browser/index.js`
- Make all files import `node/index.js` by default
- Make UMD module build skip Node components
- Split Node and browser specific tests into separate folders `test/internal/node` and `test/internal/browser`
- Make UMD module tests skip tests in `test/node`
- Make Node tests skip `test/browser`
- Use "browser" field in package.json to make bundlers replace Node components with browser components
- Use browserify transform to replace `require('./node')` with `require('./browser')` when building UMD modules
- Remove runtime feature detection in favor of "main" and "browser" fields in package.json and require rewriting for UMD module
- Add integration test with Webpack
- Exclude test UMD bundle `neo4j-web.test.js` from the npm package
- Migrate couple tests to ES6 syntax
- Move `DummyChannel` to test scope

Resolves #396


ali-ince (approved):
LGTM, I've also verified it with a very dumb test drive with `create-react-app`.

oskarhane (approved):
Much appreciated change!
This looks good to me, thank you.

I've tried it in neo4j-browser which is using webpack 3 and the `browser` field overrides in package.json works fine. After this change we can remove our (neo4j-browser) webpack alias workaround:

```
'neo4j-driver': 'neo4j-driver/lib/browser/neo4j-web.min.js',
```

which in turn should make the bundled code for neo4j-browser better structured.

So 👍

PR #425, opened 2018-11-01
Fix TTL parsing when driver configured to use JS numbers

### LINKED ISSUE #424: Causal Cluster Unable To Parse TTL Entry From Router
Hi,

I'm using the js driver version 1.7 and haven't been able to connect to my causal cluster (same issue with previous versions). I dont understand why it can't parse it because it appears the TTL (300) is right there . Any insight on what i'm doing wrong?

I use bolt+routing:// neo-core-1.al-dev.ecs.local.

```
const driver = neo4j.driver(NEO4J_URI, neo4j.auth.basic(NEO4J_USER, NEO4J_PASSWORD), {
	disableLosslessIntegers: true,
	encrypted: 'ENCRYPTION_OFF'
});

runCypherQuery(driver.session(), 'MATCH (n) RETURN COUNT(n)')
	.then(r => {
		console.log('--> NODE COUNT: ', r);
	})
	.catch(console.log);

```

{ Neo4jError: Unable to parse TTL entry from router neo-core-1.al-dev.ecs.local:7687 from record:
{"keys":["ttl","servers"],"length":2,"_fields":[300,[{"addresses":["neo-core-1.al-dev.ecs.local:7687"],"role":"WRITE"},{"addresses":["neo-core-3.al-dev.ecs.local:7687","neo-core-2.al-dev.ecs.local:7687"],"role":"READ"},{"addresses":["neo-core-2.al-dev.ecs.local:7687","neo-core-1.al-dev.ecs.local:7687","neo-core-3.al-dev.ecs.local:7687"],"role":"ROUTE"}]],"_fieldLookup":{"ttl":0,"servers":1}}

bolt without routing works, but i need to be able to take advantage of the cluster

Doing some debugging -- I see the rror returned from routing-util.js is 

```
 TypeError: record.get(...).multiply is not a function
    at RoutingUtil.parseTtl (/Users/albertfrimpong/Documents/Development/sourcify-api/node_modules/neo4j-driver/lib/v1/internal/routing-util.js:112:38)
    at /Users/albertfrimpong/Documents/Development/sourcify-api/node_modules/neo4j-driver/lib/v1/internal/rediscovery.js:83:49
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

The ttl returned from record.get is of type number but its seems thats not what we expect?

lutovich:
Driver can be created with `{disableLosslessIntegers: true}` config. Then all integer numbers returned in query results will be `Number` instead of `neo4j.Integer`.

Routing table response parsing code did not expect `Number`s and failed to parse the TTL entry. This PR fixes parsing by always wrapping returned value of TTL in a `neo4j.Integer`.

Also improved error messages around routing response parsing to include the original error message.

Fixes #424

### DIFF
--- a/src/v1/internal/routing-util.js
+++ b/src/v1/internal/routing-util.js
@@ -61,7 +61,7 @@ export default class RoutingUtil {
   parseTtl(record, routerAddress) {
     try {
       const now = int(Date.now());
-      const expires = record.get('ttl').multiply(1000).add(now);
+      const expires = int(record.get('ttl')).multiply(1000).add(now);
       // if the server uses a really big expire time like Long.MAX_VALUE this may have overflowed
       if (expires.lessThan(now)) {
         return Integer.MAX_VALUE;
@@ -69,7 +69,7 @@ export default class RoutingUtil {
       return expires;
     } catch (error) {
       throw newError(
-        'Unable to parse TTL entry from router ' + routerAddress + ' from record:\n' + JSON.stringify(record),
+        `Unable to parse TTL entry from router ${routerAddress} from record:\n${JSON.stringify(record)}\nError message: ${error.message}`,
         PROTOCOL_ERROR);
     }
   }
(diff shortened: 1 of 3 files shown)


ali-ince (approved):
LGTM.

PR #428, opened 2018-11-14
Fix TypeScript exports for classes

### LINKED ISSUE #426: Cannot use Point with TypeScript
I'm using this driver in a typescript-node project and came across this issue where TS prevents me from using the Point because the type is missing a signature for `new`.

`Cannot use 'new' with an expression whose type lacks a call or construct signature.`
screen shot 2018-11-08 at 18 51 00

It's like it doesn't understand that it is a class. I solved the type issue, although hacky, by extending the Point type myself:
```
import { v1 as neo4j } from 'neo4j-driver';
import { Point as PointT } from 'neo4j-driver/types/v1/spatial-types';

const Point = neo4j.types.Point as PointT & {
  new (srid: number, x: number, y: number, z?: number): PointT;
};
```
screen shot 2018-11-08 at 19 01 52

This works and I'm now able to create Points. However, if I try to replicate the fix into `spatial-types.d.ts` with `new(srid: T, x: number, y: number, z?: number): Point;` it does not work. Would love to figure out how this can be fixed for everyone.

lutovich:
They were previously exported in a way that did not allow user code to create instances using `new`. For example:

```typescript
const date = new neo4j.types.Date(1, 2, 3);
```

did not compile.

Added tests for all types exported from `neo4j.types` object.

Fixes #426

### DIFF
--- a/test/types/index.test.ts
+++ b/test/types/index.test.ts
@@ -79,18 +79,12 @@ const protocolError: string = neo4j.error.PROTOCOL_ERROR;
 const error1: neo4j.Neo4jError = new neo4j.Neo4jError("Error message");
 const error2: neo4j.Neo4jError = new neo4j.Neo4jError("Error message", "Error code");
 
-const node: neo4j.Node = new neo4j.types.Node(int1, [], {});
-const relationship: neo4j.Relationship = new neo4j.types.Relationship(int1, int1, int1, "", {});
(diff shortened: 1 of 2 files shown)


ali-ince (commented):
I have a small comment around exporting `Node`, `Relationship`, `UnboundRelationship`, `PathSegment`, `Path` and `Record` with constructors as this change will make them part of the public API. Would it be better to keep them as they were?

PR #429, opened 2018-11-14
Fix conversion from standard Date to Time & DateTime

### LINKED ISSUE #427: Incorrect offset format for DateTime
## Expected behaviour
`new Date().getTimezoneOffset()` returns -180 for UTC+3.
`DateTime.fromStandardDate(new Date()).toString()` should return the value `YYYY-MM-DDThh:mm:ss.ms+03:00`.

## Unexpected behaviour
`DateTime.fromStandardDate(new Date()).toString()` returns the value `YYYY-MM-DDThh:mm:ss.ms-03:00`.
 
JS [Date.getTimezoneOffset()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset#Description) returns a negative offset for the positive local timezone.

lutovich:
Time zone offset was not converted correctly. Offset exposed by the standard JavaScript `Date` is the difference, in minutes, from local time to UTC. So positive offset means local time is behind UTC and negative means it's ahead. This is different from Neo4j temporal types that support time zone offsets - `Time` and `DateTime`. They define offset as the difference, in seconds, from UTC to local time. Previous code converted offset in minutes to offset in seconds but did not change the sign of the value.

Fixes #427

### DIFF
--- a/src/v1/internal/temporal-util.js
+++ b/src/v1/internal/temporal-util.js
@@ -311,11 +311,22 @@ export function totalNanoseconds(standardDate, nanoseconds) {
 
 /**
  * Get the time zone offset in seconds from the given standard JavaScript date.
+ *
+ * <b>Implementation note:</b>
+ * Time zone offset returned by the standard JavaScript date is the difference, in minutes, from local time to UTC.
+ * So positive value means offset is behind UTC and negative value means it is ahead.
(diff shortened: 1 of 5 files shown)


ali-ince (approved):
LGTM.

PR #430, opened 2018-11-15
Add support for simple integer-like numbers

bobvanderlinden:
Currently it is possible to pass integers to queries using the [`Integer` type](https://github.com/neo4j/neo4j-javascript-driver/blob/0d259a1611c628aa62a916f6040e5015f2e4f844/src/v1/integer.js). Normal numbers are interpreted as `float` by this driver. This can happen unintentionally.

This happened for a query parameter where a normal number was passed, expecting it to be interpreted as an ID and thus integer. In the query the parameter was compared with an ID from the database and the equality always ended up being `false`. This was because this driver interpreted the value as being `float` and thus the query would check equality of an float with an actual integer, always resulting in false.

This PR adds another case for the packer to check whether a Javascript number can be interpreted as an integer and if so, pack it as an integer. In other cases the number will still be packed as a float.

### DIFF
--- a/src/v1/internal/packstream-v1.js
+++ b/src/v1/internal/packstream-v1.js
@@ -112,6 +112,8 @@ class Packer {
       return () => this._ch.writeUInt8( TRUE );
     } else if (x === false) {
       return () => this._ch.writeUInt8( FALSE );
+    } else if (Number.isInteger(x)) {
+      return () => this.packInteger(Integer.fromInt(x))
     } else if (typeof(x) == "number") {
       return () => this.packFloat(x);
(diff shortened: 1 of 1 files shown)


bobvanderlinden:
Hmm, I looked around some more and found the current patch will not be compatible with floating numbers that happen to be rounded and thus interpreted as integer. The current driver implementation is correct, but I stumbled upon this issue while using `neo4j-graphql-js`. The current behaviour is very sneaky when working with integers. Will look into patching this in neo4j-graphql-js, but suggestions are welcome!

PR #435, opened 2018-12-05
Export Integer type

joknelid:


lutovich (approved):
LGTM! Having a test in `types.test.js` that accesses the exported type as `neo4j.types.Integer` would be nice. Just to make sure it's not removed accidentally during a refactoring.

joknelid:
@lutovich Added a test now

PR #438, opened 2019-01-09
Remove tck

zhenlineo:
Removing tck.
Thinking of merging to 2.0.


test/v1/summary.test.js
```diff
@@ -0,0 +1,138 @@
...
+  it('should get profile from summary', done => {
+    session.run("PROFILE RETURN 1").then(result => {
+      var summary = result.summary;
+      expect(summary.plan).toBeDefined();
+      expect(summary.profile).toBeDefined();
+
+      var profile = summary.profile;
+      var plan = summary.plan;
+
+      verifyPlansAreEqual(profile, plan);
```
lutovich: Can `isEqual` from lodash be used instead of this function?
I.e. `expect(_.isEqual(profile, plan)).toBeTruthy()`

test/v1/summary.test.js
```diff
@@ -0,0 +1,138 @@
...
+  });
+
+  afterEach(() => {
+    driver.close();
+  });
+
+  it('should get result summary', done => {
+    // When & Then
+    session.run("CREATE (p:Person { Name: 'Test'})").then(result => {
+      var summary = result.summary;
```
lutovich: Imho it is better to always use `let` or `const` instead of `var` because they have a narrower scope.

test/v1/summary.test.js
zhenlineo: No, my bad, I am actually compare profile and plan. The profile has two extra fields, so the isEqual will not work. Change the method name to `verifyProfileAndPlanEqual` instead.

PR #444, opened 2019-03-05
Pass AccessMode in BEGIN and RUN messages

ali-ince:
This PR makes the `AccessMode` set on client side available on messages BEGIN and RUN in Bolt V3.

No `mode` parameter in statement metadata means that the access mode set by client is `WRITE`, whereas a value of `r` means it is set as `READ`.

It also addresses an issue about API docs where it mentioned session pools instead of connection pools.


PetrJanouch:
Looks good

PR #447, opened 2019-04-04
Remove deprecated API

ali-ince:
This PR removes previously deprecated API. These are;

1. `TRUST_ON_FIRST_USE` and `TRUST_SIGNED_CERTIFICATES` trust strategies are removed along with `knownHosts` configuration option that is related with `TRUST_ON_FIRST_USE`.
2. `loadBalancingStrategy` configuration option is removed and `least_connected` is the only option going forward.
3. `connectionPoolSize` configuration option is removed and `maxConnectionPoolSize` is the only option that has affect on the connection pool sizing.
4. `Driver.onCompleted` event property is now removed and replaced by `Driver.verifyConnectivity` which returns a `Promise`.


ali-ince:
@oskarhane would you please take a look this PR, especially how we've replaced `Driver.onCompleted`?

oskarhane (approved):
LGTM @ali-ince

PR #448, opened 2019-04-05
Enable ESLint on driver project

ali-ince:
This PR enables ESLint on the driver project to enforce standard code styling and checks for problematic patterns.

Based on #447.


lutovich (approved):
LGTM. Noticed a bit weird formatting in two files:
 * `integer.js` has `var`s on the same line and quite long if-else on the same line
 * `connection.js` has very long if-else lines

Would be nice to also enable linting on TypeScript files in future.

oskarhane:
While you're at it, I would recommend adding https://prettier.io as well.
It's _very_ common in js projects these days.
Auto formatting, in addition to linting.
We use it in browser, desktop etc.

I figured I'd mention it while you're poking around with formatting.

ali-ince:
@lutovich thanks for the review, with @oskarhane's suggestion I've also integrated prettier and it resulted in the formatting errors you've pointed out to be resolved.

I'll take a look at typescript linting in the future.

thanks!

oskarhane:
Nice! I'd suggest you add the lint-staged step for formatting with prettier on commits, to keep the code well formatted (it only formats staged files).
See: https://prettier.io/docs/en/precommit.html

PR #451, opened 2019-04-17
Changed `bolt+routing` scheme to `neo4j`

zhenlineo:
Changed the routing scheme from `bolt+routing` to `neo4j` for routing everywhere.


ali-ince (approved):
:+1:

PR #453, opened 2019-04-30
Fixes around initial discovery

ali-ince:
This PR introduces a couple of fixes around initial/fallback routing discovery through seed router.

1. Introduce a dedicated `ServerAddress` class to hold server addresses, including resolved IP addresses, to be used across the driver to establish connections. This is specifically done to fix a failure where we forget host name (required for `SNI` to work) and rely on IP addresses to establish TLS connections where TLS connection is towards an endpoint that expects `SNI`. The introduced class will always include the initial host name even after it is resolved through DNS.
2. The routing driver, when constructed, attempts discovery through the provided seed URI and then falled back to configured `custom` resolver. This was not the intended behaviour with initial discovery and it is now defaulted to use seed URI applied to provided custom resolver results of which are passed on to DNS resolution. This is the behaviour which will apply on initial and fall-back to initial discovery.

### DIFF
--- a/src/v1/internal/connection-providers.js
+++ b/src/v1/internal/connection-providers.js
@@ -76,10 +76,9 @@ export class LoadBalancer extends ConnectionProvider {
   }
 
   acquireConnection(accessMode) {
-    let that = this;
     const connectionPromise = this._freshRoutingTable(accessMode).then(routingTable => {
       if (accessMode === READ) {
-        const address = that._loadBalancingStrategy.selectReader(routingTable.readers);
+        const address = this._loadBalancingStrategy.selectReader(routingTable.readers);
         return this._acquireConnectionToServer(address, 'read');
       } else if (accessMode === WRITE) {
         const address = this._loadBalancingStrategy.selectWriter(routingTable.writers);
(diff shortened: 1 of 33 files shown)


src/v1/internal/connection-providers.js
```diff
@@ -45,38 +46,40 @@ class ConnectionProvider {
...
-    this._useSeedRouter = false;
+    this._useSeedRouter = true;
   }
 
   acquireConnection(accessMode) {
+    let that = this;
     const connectionPromise = this._freshRoutingTable(accessMode).then(routingTable => {
       if (accessMode === READ) {
-        const address = this._loadBalancingStrategy.selectReader(routingTable.readers);
+        const address = that._loadBalancingStrategy.selectReader(routingTable.readers);
```
zhenlineo: Is variable `that` needed here? It looks like a reference copy to me.

zhenlineo (approved):

src/v1/internal/connection-providers.js
ali-ince: definitely not :) thanks for the catch.

PR #471, opened 2019-07-10
Multi-database routing

ali-ince:
This PR brings multi-database routing support to the javascript driver. Besides this core feature, it includes a couple of small other changes as well.

1. The `db` parameter specified during session construction (for selecting the target database) is updated as `database` to make it aligned with other drivers.
2. Removed `driver.onError` callback completely.
2. Stub tests are now grouped into bolt versions and most test scripts are based on Bolt V3.
3. Jasmine setup is simplified so that it's now possible to directly run tests without going through the packaging step - which saves lots of time!
4. Most tests are now classified as `unit` and `integration` and are now possible to selectively execute them using `gulp test-nodejs-unit` or `gulp test-nodejs-integration`. Or it is possible to directly use filter argument to `jasmine` executable to filter desired tests to be run (i.e. `jasmine --filter="#unit*"`).


ali-ince:
@zhenlineo I'm merging this PR. Please feel free to add your comments and I'll address them in a seperate PR.

PR #473, opened 2019-07-23
enable connection URI contains path, so can use proxy to hide the backend db

chuter:


technige:
Hi @chuter. Please explain what you are trying to achieve with this PR.

chuter:
> Hi @chuter. Please explain what you are trying to achieve with this PR.

I used driver to connect to my backend db, need rewrite rule to route db request to the db server because the safe stuff.

But the default address parse in driver so strict that can not accept the address without port, so i change the parse behavior to accept that address.

technige:
@chuter Sorry but we can't accept this PR as the URL path is reserved for future internal use.

PR #480, opened 2019-08-23
Improve error messages and objects around Transaction commit/rollback

ali-ince:
This PR simplifies the error messages thrown in transaction commit / rollback cases. It also makes the thrown error objects to be actual `Neo4jError` objects. 

Updated related tests and re-enabled previously skipped tests.

Based on #479.


src/driver.js
```diff
@@ -209,6 +266,9 @@ function sanitizeConfig (config) {
   )
 }
 
+/**
+ * @private
+ */
 function sanitizeIntValue (rawValue, defaultWhenAbsent) {
```
zhenlineo, line 272: Do we follow the practice to name private methods start with `_`?

src/internal/browser/browser-channel.js
```diff
@@ -247,16 +264,16 @@ function determineWebSocketScheme (config, protocolSupplier) {
 
   if (encryptionOn) {
     // encryption explicitly requested in the config
-    if (!trust || trust === 'TRUST_CUSTOM_CA_SIGNED_CERTIFICATES') {
+    if (!trust || trust === 'TRUST_SYSTEM_CA_SIGNED_CERTIFICATES') {
```
zhenlineo, line 267: Does this imply that if a customer properly set up the server cert, then when a user connecting, the user might get the cert trust chain from the browser directly?

src/internal/connection-provider-direct.js
```diff
@@ -27,7 +27,7 @@ export default class DirectConnectionProvider extends PooledConnectionProvider {
     this._address = address
   }
 
-  acquireConnection (accessMode, database) {
+  acquireConnection ({ accessMode, database, bookmarks } = {}) {
```
zhenlineo, line 34: Maybe some comments about how the bookmarks are used in this acquireConn method. This method reads like `acquire a connection with access mode xx for database yy and at least up-to bookmarks zz.` However the bookmarks are only passes to getRT calls. So maybe a small comment here for people who is not familiar with our internals.

test/internal/bookmark.test.js
```diff
@@ -34,35 +34,60 @@ describe('#unit Bookmark', () => {
...
 
   it('should be possible to construct bookmark from empty array', () => {
     const bookmark = new Bookmark([])
 
     expect(bookmark.isEmpty()).toBeTruthy()
-    expect(bookmark.maxBookmarkAsString()).toBeNull()
+    expect(bookmark.values()).toEqual([])
+  })
+
+  it('should be possible to construct bookmark from nested arrays', () => {
```
zhenlineo, line 72: Is it legal to have `session.lastBookmark` to return `null` or `undefined`?
If so, maybe also add a test for testing `null` and/or `undefined` inside bookmark arrays?
Basically to simulate how a user would give bookmark arrays to session.

zhenlineo (approved):
Looks awesome. Just have a few small comments.

src/driver.js
ali-ince: For class fields and methods, that naming convention holds. However, that method is not a class member and we don't follow that naming convention for non-exported/internal functions which are not accessible by end users.

src/internal/browser/browser-channel.js
ali-ince: For web sockets, no luck of any kind. 

You don't have any control on trust settings, etc. If the browser decides that the server is trusted then you get the connection, otherwise you don't.

This name change, specifically, is because `CUSTOM` word implies that it would be possible to connect if you provide trusted certificates through the config (which is the case for nodejs driver), however it's not so and I believe it's misleading. By naming it as `SYSTEM` now implies that we do not have any control on it.

src/internal/connection-provider-direct.js
ali-ince: Added doc comments in `connection-provider.js` and added a note to the sub-class overrides.

test/internal/bookmark.test.js
ali-ince: There wouldn't be any `null` or `undefined` returns from `session.lastBookmark()`. However, for the case if a user would somehow pass; I've added a new test case that would test nested arrays with `null` and `undefined` inside.

PR #491, opened 2019-09-30
Add supportsMultiDb function

ali-ince:
This PR adds a new `supportsMultiDb` method on Driver, that when called negotiates a protocol for the remote database (or one of the seed routers) and returns a boolean Promise telling whether it supports multi database features _based on the bolt protocol version only_.

Based on #490.


oskarhane:
Thanks @ali-ince!

zhenlineo (approved):
Just wondering if we can just do a `return 1` and return "support multi-db" if the server version in result summary is greater than 4.0?

ali-ince:
@zhenlineo we wanted it to happen even without authentication (due to the fact that on initial login the password could be reported as expired and nothing other than change password functionality is allowed) so picked the handshake path.

Issue #469, opened 2019-06-06
[Feature Request] Record map


thebestnom:
hi, I would like to add a suggestion that Record will have a map function, it's already have forEach so I think its not too difference.
my use case is doing async stuff inside the forEach and I want to await there

``` javascript
const promises = recored.map(async (v) => {
   // ... stuff
   await stuff(v)
   //... more stuff
})
await Promise.all(promises)
``` 
which I cant do with forEach,
what I did instead for now is
``` javascript
const promises = Object.values(record.toObject()).map(async (v) => {
   // ... stuff
   await stuff(v)
   //... more stuff
})
await Promise.all(promises)
```

which is kinda hack

zhenlineo:
Hi @thebestnom , thanks very much for you input. We've added this as a new feature in 4.0 drivers. You can try this feature out with our latest 4.0.0-beta01 and later drivers.

PR #494, opened 2019-10-10
Implemented map function for Record.

martin-neotech:
1. Implemented map function for Record.

2. Also included is functions for graph Objects.

3. Made the IDENTIFIER_PROPERTY_ATTRIBUTES more explicit.

4. Need to implement tests for isPath, isNode, etc

# record.map

https://github.com/neo4j/neo4j-javascript-driver/issues/469

hi, I would like to add a suggestion that Record will have a map function, it's already have forEach so I think its not too difference.
my use case is doing async stuff inside the forEach and I want to await there

```
const promises = record.map(async (v) => {
   // ... stuff
   await stuff(v)
   //... more stuff
})
```
await Promise.all(promises)
which I cant do with forEach,
what I did instead for now is

```
const promises = Object.values(record.toObject()).map(async (v) => {
   // ... stuff
   await stuff(v)
   //... more stuff
})
```
await Promise.all(promises)
which is kinda hack

### DIFF
--- a/src/graph-types.js
+++ b/src/graph-types.js
@@ -17,6 +17,23 @@
  * limitations under the License.
  */
 
+const IDENTIFIER_PROPERTY_ATTRIBUTES = {
+  value: true,
+  enumerable: false,
+  configurable: false,
+  writable: false
+}
+
+const NODE_IDENTIFIER_PROPERTY = '__isNode__'
+const RELATIONSHIP_IDENTIFIER_PROPERTY = '__isRelationship__'
+const UNBOUND_RELATIONSHIP_IDENTIFIER_PROPERTY = '__isUnboundRelationship__'
+const PATH_IDENTIFIER_PROPERTY = '__isPath__'
+const PATH_SEGMENT_IDENTIFIER_PROPERTY = '__isPathSegment__'
+
+function hasIdentifierProperty (obj, property) {
+  return (obj && obj[property]) === true
+}
+
 /**
  * Class for Node Type.
  */
@@ -54,7 +71,7 @@ class Node {
     for (let i = 0; i < this.labels.length; i++) {
       s += ':' + this.labels[i]
     }
-    let keys = Object.keys(this.properties)
+    const keys = Object.keys(this.properties)
     if (keys.length > 0) {
       s += ' {'
       for (let i = 0; i < keys.length; i++) {
@@ -66,6 +83,21 @@ class Node {
     s += ')'
     return s
   }
+}
+
+Object.defineProperty(
+  Node.prototype,
+  NODE_IDENTIFIER_PROPERTY,
+  IDENTIFIER_PROPERTY_ATTRIBUTES
+)
+
+/**
+ * Test if given object is an instance of {@link Node} class.
+ * @param {Object} obj the object to test.
+ * @return {boolean} `true` if given object is a {@link Node}, `false` otherwise.
+ */
+function isNode (obj) {
+  return hasIdentifierProperty(obj, NODE_IDENTIFIER_PROPERTY)
 }
 
 /**
@@ -114,7 +146,7 @@ class Relationship {
    */
   toString () {
     let s = '(' + this.start + ')-[:' + this.type
-    let keys = Object.keys(this.properties)
+    const keys = Object.keys(this.properties)
     if (keys.length > 0) {
       s += ' {'
       for (let i = 0; i < keys.length; i++) {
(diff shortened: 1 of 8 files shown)


zhenlineo:
@lutovich Can you help to review this PR :)

test/record.test.js
```diff
@@ -113,4 +113,17 @@ describe('#unit Record', () => {
     // Then
     expect(result).toEqual([['Bob', 'name', record], [45, 'age', record]])
   })
+
+  it('should allow map function for the record', () => {
+    // Given
+    const record = new Record(['name', 'age'], ['Bob', 45])
+
+    // When
+    const result = record.map(function (value, key, rec) {
```
lutovich: Maybe use an arrow function instead `(value, key, rec) => [value, key, rec]`

src/graph-types.js
```diff
@@ -260,4 +337,30 @@ class Path {
...
+/**
+ * Test if given object is an instance of {@link Path} class.
+ * @param {Object} obj the object to test.
+ * @return {boolean} `true` if given object is a {@link Path}, `false` otherwise.
+ */
+function isPath (obj) {
+  return hasIdentifierProperty(obj, PATH_IDENTIFIER_PROPERTY)
+}
+
+export {
```
lutovich, line 355: TypeScript declarations also need to be updated to include these new functions

lutovich (commented):
Looks good, made two small comments.
It would also be nice to add tests for `isNode`, `isRelationship`, etc.

types/record.d.ts
```diff
@@ -19,6 +19,8 @@
 
 declare type Visitor = (value: any, key: string, record: Record) => void
 
+declare type MapVisitor = (value: any, key: string, record: Record) => any
```
lutovich: I think it's better to use a generic type here instead of `any`.

Visitor type would be:

```typescript
declare type MapVisitor<T> = (value: any, key: string, record: Record) => T
```

and `map` function would become:

```typescript
map<T>(visitor: MapVisitor<T>): T[];
```

types/record.d.ts
```diff
@@ -31,6 +33,8 @@ declare class Record {
 
   forEach(visitor: Visitor): void
 
+  map(visitor: MapVisitor): Array<any>
```
lutovich: Would be nice to add a test for this new function declaration in `record.test.ts`

lutovich (approved):
LGTM, two small comments

zhenlineo:
Can you also clean your git commit history to remove unneeded commits such as `Merge remote-tracking branch 'origin/4.0' into recordmapfunction`
and merge small commits into bigger one commits, such as commits for `Update TOOLS.md`.

This is needed for anyone else who need to read git history when coding to figure out why you need to make changes at certain lines. 

You can use git commands such as `rebase` to clean up git history.

TOOLS.md
```diff
@@ -1,19 +1,64 @@
 # List of tools used throughout the project
 
-### npm
+## boltkit
```
zhenlineo: I feel most of the changes to this file is not needed.
1. we should rarely include a link to website in docs as 1) it is hard to make sure these links works over time. and 2) it is not that hard to find them with google. So I would suggest to remove all the links added in this doc.

2. There is no need to mention how to install boltkit and what java version are needed to install for testing as it not specific to this driver. They are instead driver test infrastructure. Writing to here will just add an extra place to maintain the correctly of how to use them. If you find such info is missing, you can add a card in inbox describe what you think it is missing and we will discuss the priority of this job.

I would suggest to remove section `boltkit`, `java version` and links that you added in this file.

TOOLS.md
```diff
@@ -1,19 +1,64 @@
...
+> `npm cmd` will check the `package.json` with the key `scripts` and run the command.
 
 `NEOCTRL_ARGS="-e 4.0" npm test`
 
 `NEOCTRL_ARGS="-e 4.0" npm run start-neo4j`
 
 `NEOCTRL_ARGS="-e 4.0" npm run stop-neo4j`
 
-### gulp
+**Note:** setting the variable does not work on Windows.
```
zhenlineo: This sounds more like a bug rather than a thing to document in this file.

TOOLS.md
```diff
@@ -1,19 +1,56 @@
 # List of tools used throughout the project
 
-### npm
+## boltkit
+
+Python code to start a neo4j server.
+
+https://github.com/neo4j-drivers/boltkit/tree/1.3
+
+python -m pip install https://github.com/neo4j-drivers/boltkit@1.3#egg=boltkit
```
zhenlineo: I feel most of the changes to this file is not needed.

we should rarely include a link to website in docs as 1) it is hard to make sure these links works over time. and 2) it is not that hard to find them with google. So I would suggest to remove all the links added in this doc.

There is no need to mention how to install boltkit and what java version are needed to install for testing as it not specific to this driver. They are instead driver test infrastructure. Writing to here will just add an extra place to maintain the correctly of how to use them. If you find such info is missing, you can add a card in inbox describe what you think it is missing and we will discuss the priority of this job.

I would suggest to remove section boltkit, java version and links that you added in this file.

PR #507, opened 2019-12-03
JSDocs updates and fixes.

martin-neotech:
Minor typo fixes in JSDocs. Nothing Major found to be fixed.
- Cypher queries examples uses the dollar sign now.

### DIFF
--- a/src/temporal-types.js
+++ b/src/temporal-types.js
@@ -46,10 +46,10 @@ const DATE_TIME_IDENTIFIER_PROPERTY = '__isDateTime__'
 export class Duration {
   /**
    * @constructor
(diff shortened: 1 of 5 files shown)


src/temporal-types.js
```diff
@@ -142,11 +142,11 @@ export class LocalTime {
   }
 
   /**
-   * Create a local time object from the given standard JavaScript `Date` and optional nanoseconds.
+   * Create a `LocalTime` object from the given standard JavaScript `Date` and optional nanoseconds.
```
zhenlineo: You can use class reference `{@link DateTime}` instead. Same for the others changes.

PR #532, opened 2020-02-29
Add type-safety to Record class

parzhitsky:
### Description

This PR adds ability to create type-safe Record entities by assigning type to Record's value.
The implementation is based on adding type parameters to types, interfaces, and classes. It is implemented however in a backwards-compatible manner, so that opting in to type safety is optional.

***

### Usage

```ts
const record = new Record<Person>(['name', 'age'], ['Alice', 20])
```

##### Before (unsafe):

```ts
record.get('neam') // typo, but no errors

record.toObject().neam[0] // run-time (late) error
```

##### After (safe):

```ts
record.get('neam')
// Error: property 'neam' does not exist

record.toObject().neam[0] // design-time (early) error
```
image

### DIFF
--- a/test/types/record.test.ts
+++ b/test/types/record.test.ts
@@ -19,15 +19,25 @@
 
 import Record from '../../types/record'
 
+interface Person {
+  name: string
+  age: number
+}
+
 const record1 = new Record(['name', 'age'], ['Alice', 20])
-const record2 = new Record(['name', 'age'], ['Bob', 22], { key: 'value' })
+const record2 = new Record(['name', 'age'], ['Bob', 22], { firstName: 0 })
+const record3 = new Record<Person>(['name', 'age'], ['Carl', 24])
+
 const isRecord1: boolean = record1 instanceof Record
 const isRecord2: boolean = record2 instanceof Record
+const isRecord3: boolean = record3 instanceof Record
 
 const record1Keys: string[] = record1.keys
+const record3Keys: Array<keyof Person> = record3.keys
 const record1Length: number = record1.length
 
 const record1Object: object = record1.toObject()
+const record3Object: Person = record3.toObject()
 
 record1.forEach(() => {})
 
@@ -36,6 +46,16 @@ record1.forEach((value: any) => {})
 record1.forEach((value: any, key: string) => {})
 
 record1.forEach((value: any, key: string, record: Record) => {})
+
+record3.forEach(
+  (value: string | number, key: 'name' | 'age', record: Record<Person>) => {}
+)
+
+const record3Mapped: [
+  string | number,
+  'name' | 'age',
+  Record<Person>
+][] = record3.map((...args) => args)
 
 const record1Entries: IterableIterator<[string, any]> = record1.entries()
 const record2Entries: IterableIterator<[string, any]> = record2.entries()
@@ -49,8 +69,17 @@ const record2ToArray: any[] = [...record2]
 const record1Has: boolean = record1.has(42)
 const record2Has: boolean = record1.has('key')
 
-const record1Get1: any = record1.get(42)
-const record2Get1: any = record2.get('key')
+const record1Get1: any = record1.get('name')
+const record2Get1: any = record2.get('age')
 
-const record1Get2: object = record1.get(42)
-const record2Get2: string[] = record2.get('key')
+const record1Get2: object = record1.get('name')
+const record2Get2: string[] = record2.get('age')
+
+const record3Get1: string = record3.get('name')
+const record3Get2: number = record3.get('age')
+
+const record2Get3: string = record2.get('firstName')
+const record2Get4: number = record2.get(1)
+
+// @ts-expect-error
+const record2Get5: any = record2.get('does-not-exist')
--- a/types/record.d.ts
+++ b/types/record.d.ts
@@ -17,23 +17,38 @@
  * limitations under the License.
  */
 
-declare type Visitor = (value: any, key: string, record: Record) => void
+declare type Dict<Key extends PropertyKey = PropertyKey, Value = any> = {
+  [K in Key]: Value
+}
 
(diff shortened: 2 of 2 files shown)


martin-neotech:
Hi @parzhitsky can you please resolve the merge conflicts, thanks for contributing.

Cheers Martin

parzhitsky:
Is there anything else I can do with this PR?

martin-neotech:
@parzhitsky Sorry for the delay, we do not have any one at the moment to review your Typescript improvements.

parzhitsky:
Oh, got it. Please take your time, no worries

leefreemanxyz:
Is there any chance of this getting merged soon? It would be a really nice feature to have.

test/types/record.test.ts
```diff
@@ -19,15 +19,25 @@
 
 import Record from '../../types/record'
 
+interface Person {
+  name: string
+  age: number
+}
+
 const record1 = new Record(['name', 'age'], ['Alice', 20])
 const record2 = new Record(['name', 'age'], ['Bob', 22], { key: 'value' })
```
bigmontz: Could you check if we still could get the value of the key entry? 

`const record2Get1: any = record2.get('key')`

parzhitsky: I believe, this is checked [at lines 72-79](https://github.com/neo4j/neo4j-javascript-driver/pull/532/files#diff-90dd771e8066d69253abb1ec7d0feb0f63c21505960146852393df55e2ab2ce5R72-R79)

bigmontz:
Sorry for the delay, @parzhitsky. Could you please move the target branch to 4.3?

test/types/record.test.ts
bigmontz: It's not, i think you remove the line by accident.

test/types/record.test.ts
```diff
@@ -49,8 +69,11 @@ const record2ToArray: any[] = [...record2]
...
-const record1Get1: any = record1.get(42)
-const record2Get1: any = record2.get('key')
+const record1Get1: any = record1.get('name')
+const record2Get1: any = record2.get('age')
+
+const record1Get2: object = record1.get('name')
+const record2Get2: string[] = record2.get('age')
 
-const record1Get2: object = record1.get(42)
-const record2Get2: string[] = record2.get('key')
```
bigmontz, line 56: This line should be put be putted back on the tests

test/types/record.test.ts
```diff
@@ -49,8 +69,11 @@ const record2ToArray: any[] = [...record2]
 const record1Has: boolean = record1.has(42)
 const record2Has: boolean = record1.has('key')
 
-const record1Get1: any = record1.get(42)
-const record2Get1: any = record2.get('key')
```
bigmontz, line 53: This line should be put be putted back on the tests

test/types/record.test.ts
parzhitsky: Oh, right, I get it now

parzhitsky: I couldn't find any usage example of the third parameter of `Record` constructor (`fieldLookup`); judging from [the definition](https://github.com/neo4j/neo4j-javascript-driver/blob/79d7fd431d83e8ef11a9ad5e0b50a4731780195d/src/record.js#L22-L28), it seems to be an aliasing mechanism:

```ts
const record = new Record([ "name" ], [ "Alice" ], { name: 0, firstName: 0 })

record.get("name")
// "Alice"

record.get("firstName")
// "Alice"
```

@bigmontz, Could you confirm? If so, then I should change the test cases a bit, because `'value'` is not a valid array index

test/types/record.test.ts
parzhitsky: What should be the value type in this case?

```ts
const value: unknown = record.get('key')
```

test/types/record.test.ts
parzhitsky: Okay, nevermind. I've done my research, the type of the `fieldLookup` parameter persists since its [definition in `1.7.7`](https://github.com/neo4j/neo4j-javascript-driver/commit/8a3dc008f5767633bee688b410aafab45527c171#diff-a300f394a24c4adc5c7700248ef506551c724ad206949915a5bbdbf5c571fc12R10). Either the functionality associated with its type has changed over time, or this was an error from the start. In either case, now `fieldLookup` is actually of type  <code>{ [key: string]: ~~string~~ number }</code>, – changing to that.

bigmontz: Hi @parzhitsky, actually, i've check [here](https://github.com/neo4j/neo4j-javascript-driver/blob/79d7fd431d83e8ef11a9ad5e0b50a4731780195d/src/record.js#L56) and the def is wrong on the ts file, it works as alias but it is index based just like in our previous message.

parzhitsky: > now fieldLookup is actually of type `{ [key: string]: string }`, – changing to that

Sorry, meant to say `{ [key: string]: number }`

bigmontz: Exactly

bigmontz (approved):
@parzhitsky Thanks for your contribution, should i merge it? 

I will back port to 4.2 after merged, but I will wait a bit for the release it.

parzhitsky:
@bigmontz Sorry, I didn’t get the point. Could you rephrase it? No offense 🙂

Should be okay to merge the PR.

bigmontz:
> @bigmontz Sorry, I didn’t get the point. Could you rephrase it? No offense 🙂
> 
> Should be okay to merge the PR.

I will release this feature in the 4.2.x in the near future (one week, or something like this), I will wait a bit to see if I manage to join it with other improvements or fixes.

bigmontz:
Hi @parzhitsky, I've released this improvement on the version 4.2.2 of the driver. 

Thanks for your contribution.

parzhitsky:
Happy to help! 🙂

PR #537, opened 2020-03-11
Align TypeScript exports with what is being exported in Javascript

### LINKED ISSUE #536: Exports in TypeScript type definition are out of date.
### General Information

**Driver version**: JS driver 4.0.1  

As mentioned in #535 it seems that the TypeScript typings for version 4 of the driver were not updated to reflect the changes from version 1.7.

For example, the types (DateTime etc) are all still on the top level export which they should not.

2hdddg:
All types were exported from top level in TypeScript, export those the
same way as done in Javascript.
Made sure that index.d.ts exports all types needed to create a driver.
Made sure ordering of exports in typescript were the same (at least in
some files).

### DIFF
--- a/types/driver.d.ts
+++ b/types/driver.d.ts
@@ -107,7 +107,8 @@ export {
   Config,
   EncryptionLevel,
   TrustStrategy,
-  SessionMode
+  SessionMode,
+  LoggingConfig
 }
 
 export default Driver
--- a/types/graph-types.d.ts
+++ b/types/graph-types.d.ts
@@ -85,14 +85,14 @@ declare function isPathSegment(obj: object): boolean
 
 export {
   Node,
-  Relationship,
-  UnboundRelationship,
-  Path,
-  PathSegment,
   isNode,
+  Relationship,
   isRelationship,
+  UnboundRelationship,
   isUnboundRelationship,
+  Path,
   isPath,
+  PathSegment,
   isPathSegment,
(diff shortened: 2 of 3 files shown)


2hdddg:
Not really sure about how we want imports to work in TypeScript. Added some imports on top level that I guess is hande when a driver is created and removed some that were more related to working at the session level. Removed all types from top level that were already brough in through types. I guess this one could potentially break someones code.

zhenlineo:
@huboneo Can you give us some insight on this PR?

martin-neotech:
> Removed all types from top level that were already brough in through types. I guess this one could potentially break someones code.

I think we should just add a comment that they should be removed in the next major version for now.

kimroen:
> > Removed all types from top level that were already brough in through types. I guess this one could potentially break someones code.
> 
> I think we should just add a comment that they should be removed in the next major version for now.

I disagree with this - we broke our system today because we updated to the 4.x version of the driver and trusted that we could import `DateTime` from the top level like the types said. I don't think updating the types in a way that would have avoided that is a breaking change. It's not "slightly more correct", it's a change to reflect the reality of a thing not existing.

If we want to maintain the functionality for people who only import things like `DateTime` for use as types, I believe it might be possible to export just for use as a type?

bigmontz:
This PR is not needed anymore, since the driver surface was almost all migrated to typescript.

PR #539, opened 2020-03-13
Only override neo4j config when it is really needed

2hdddg:
Way too large pagecache is configured, boltkit sets up a cachesize of
30MB. That should be enough.


zhenlineo:
I fixed the failed docs gen.

PR #544, opened 2020-03-19
Count pending connects when considering max pool size

### LINKED ISSUE #501: MaxConnectionPoolSize is not respected under heavy load
The default internal connection pool size for this driver seems to be set at 100. I'm seeing that this can be broken. Basic test to reproduce this behavior:
```
    test('Connection pool', async () => {
        const driver = Neo4J.driver('bolt://localhost:7687', Neo4J.auth.basic('neo4j', 'neo4j'), {
            disableLosslessIntegers: true,
        });

        const p: Promise<any>[] = [];
        for (let i = 0; i < 1500; i++) {
            p.push(new Promise(async (res, rej) => {
                const session = driver.session();
                await session.run('MATCH (n) RETURN n');
                session.close();
                res();
            }));
        }

        await Promise.all(p);
    }, 30000);
```

Watch netstat while you run this test and see that the number of connections goes into the thousands:
```
watch -n .5 'netstat -an | grep "7687" | grep ESTABLISHED | wc -l'
```

I've tested this against Neo4j 3.5.3 and 3.5.12 using driver versions 1.7.5 and 4.0.0-beta01.

2hdddg:
Without this there could be a burst of connect and neogitations with the
server on connections just to be shut down immediately.

### DIFF
--- a/src/internal/pool.js
+++ b/src/internal/pool.js
@@ -56,6 +56,7 @@ class Pool {
     this._maxSize = config.maxSize
     this._acquisitionTimeout = config.acquisitionTimeout
     this._pools = {}
+    this._pendingCreates = {}
     this._acquireRequests = {}
     this._activeResourceCounts = {}
     this._release = this._release.bind(this)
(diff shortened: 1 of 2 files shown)


jharris4:
Hi @2hdddg, your changes look good to me! I assume that the test case you added was failing before you made these changes?

2hdddg:
> Hi @2hdddg, your changes look good to me! I assume that the test case you added was failing before you made these changes?

Yes it was. Also ran the test program mentioned in the linked issue. Verified by logging in the connection code, there were the correct number of connects.

jharris4:
> > Hi @2hdddg, your changes look good to me! I assume that the test case you added was failing before you made these changes?
> 
> Yes it was. Also ran the test program mentioned in the linked issue. Verified by logging in the connection code, there were the correct number of connects.

Nice! From a JS perspective this all looks good to me, I don't have any other feedback.

jharris4 (approved):

zhenlineo (commented):
🎉

PR #594, opened 2020-07-01
Aura example

mneedham:


mneedham:
This is how it'll look on the Aura connect page:

image

I couldn't think of a better way to define the Bolt URL/User/Password so that they can use the Aura placeholders, so I've put them in comments that get removed when the code is pulled into Aura land. 

Happy to change that if there's a cleaner way to do it.

test/examples.test.js
```diff
@@ -405,6 +405,73 @@ describe('#integration examples', () => {
...
+      const writeQuery = `MERGE (p1:Person { name: $person1Name })
+                          MERGE (p2:Person { name: $person2Name })
+                          MERGE (p1)-[:KNOWS]->(p2)
+                          RETURN p1, p2`
+
+      // Write transactions allow the driver to handle retries and transient errors
+      const writeResult = await session.writeTransaction(tx =>
+        tx.run(writeQuery, { person1Name, person2Name })
+      )
+      writeResult.records.forEach(row => {
```
2hdddg: Nitpick, but record is perhaps clearer than row

test/examples.test.js
```diff
@@ -405,6 +405,73 @@ describe('#integration examples', () => {
...
+        console.log(
+          `Created friendship between: ${row.get('p1').properties.name}, ${
```
2hdddg: Also nitpick, could be more clear if we use intermediate node variable
person1Node = row.get('p1')

test/examples.test.js
```diff
@@ -405,6 +405,73 @@ describe('#integration examples', () => {
...
+      console.error('Something went wrong: ', error)
+    } finally {
+      await session.close()
+    }
+
+    // Don't forget to close the driver connection when you're finished with it
+    await driver.close()
+    // end::driver-introduction-example[]
+
+    expect(await consoleLoggedMsg).toEqual(
```
2hdddg, line 470: So this part isn't visible in Aura page?

2hdddg (commented):
Looks nice

test/examples.test.js
mneedham: Yeh, so it'll only include what's between these lines:

```
// tag::driver-introduction-example[]
```

and

```
// end::driver-introduction-example[]
```

This is how we pull in the samples on the Aura side:

```
include::https://raw.githubusercontent.com/mneedham/neo4j-javascript-driver/aura-example/test/examples.test.js[tag="driver-introduction-example-imports", indent=0]
include::https://raw.githubusercontent.com/mneedham/neo4j-javascript-driver/aura-example/test/examples.test.js[tag="driver-introduction-example-variables", indent=0]
include::https://raw.githubusercontent.com/mneedham/neo4j-javascript-driver/aura-example/test/examples.test.js[tag="driver-introduction-example", indent=0]
```

That's pointing at my branch at the moment, but I'll update it to use the official repo after it's merged.

test/examples.test.js
mneedham: Makes sense

test/examples.test.js
mneedham: Have updated row -> record

2hdddg (approved):

PR #605, opened 2020-07-27
update about different behaviour for export in type script

martin-neotech:


martin-neotech:
https://github.com/neo4j/neo4j-javascript-driver/issues/536

bigmontz:
The miss-alignment between the code and the type was already solved in 4.3

PR #609, opened 2020-08-24
Revert "Dependency updates"

2hdddg:
Reverting dependency updates due to performance degradation due
to one or more dependencies as compared to 4.0.


martin-neotech (approved):
nice

jexp:
@2hdddg @martin-neotech what do we need to unblock this? `12 days ago` ?

PR #613, opened 2020-09-02
Fix connection leak due to race condition in counting resources

voutilad:
Since resources were being acquired in one async context, but being counted in another, it was possible that under high load checks of the total number of connections could report zero when connections were actually created.

This commit moves the call to resourceAcquired() to the same async context where it's being created initially or reacquired from the existing pool.


martin-neotech (approved):

martin-neotech:
I will trigger the build some more times.

PR #619, opened 2020-09-14
Implementing commands to allows the retry tests

bigmontz:
The `SessionReadTransaction` command is the responsilble for encapsulate the read function pass to the
transaction and the other ones use the transaction object got by the SessionReadTransaction.


2hdddg (commented):
Nice!

PR #626, opened 2020-09-29
Create bolt protocol 4.2

bigmontz:
It creates all the infrastructure for handle the 4.2 handshake and test
if all the existing behaviours on the 4.1 still working on the new version.

It also removes the version 2 from the handshake and remove older versions test resources which are not used anymore.


gjmwoods (approved):
🚀

PR #631, opened 2020-10-20
testkit: waiting for the current result get completed

bigmontz:
Tests which are treating big size data was failing because the result was not
being completed before the next `session.run.`.

### DIFF
--- a/testkit-backend/main.js
+++ b/testkit-backend/main.js
@@ -124,14 +124,22 @@ class Backend {
               params[key] = cypherToNative(value)
             }
           }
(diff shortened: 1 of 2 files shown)


2hdddg:
@bigmontz merge this one and reconfigure testkit TC config to run on 4.2?

bigmontz:
> @bigmontz merge this one and reconfigure testkit TC config to run on 4.2?

Sure, i will just check why the tests are failing.

bigmontz:
@2hdddg, done. Should I port this change to 4.1 ?

PR #642, opened 2020-11-24
Supporting browser, integration and stress tests on testkit

bigmontz:
Teskit doesn't run Neo4j on the localhost and with the same credentials of boltkit, but it exposes
the host and credentials as env variables.

So all this configuration being loaded to karma on the config files, parsed in `shared-neo4j` with fallback to
the process env (case it's running the usual integration test) and to the default boltkit values.

All the integration tests were changed to use the values from the shareNeo4j object.

Disabled tests:
* IPV6 tests are skipped on the testkit because of the lack of support for this
* TLS tests are disabled since it's already tested on the testkit suite
* Tests which depends on the enterprise edition of Neo4j is running only on this scenario

### DIFF
--- a/test/bolt-v3.test.js
+++ b/test/bolt-v3.test.js
@@ -37,7 +37,10 @@ describe('#integration Bolt V3 API', () => {
   let originalTimeout
 
   beforeEach(async () => {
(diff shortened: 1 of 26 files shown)


oskarhane (approved):
LGTM ☸️

PR #646, opened 2020-12-03
Create 4.3 Protocol and rediscovery the Routing Table using the Route Message

bigmontz:
Starting the development of the version 4.3 with the creation of the protocol class making it available on the handshake. The replacement of the routing procedure by the route message for routing table discovery on this new version pushes the complexity to the server and makes the protocol cleaner and more efficient.

The routing table requisition details was pushed to the protocol implementation classes, the request provides a observer which parses the response and make it available as a RawRoutingTable on completed. 

The rediscovery class uses the method, constructs the RoutingTable and treats the exceptional cases.


src/internal/routing-table-getter/routing-table-factory.js
```diff
@@ -0,0 +1,210 @@
...
+}
+
+/**
+ * Parse the ttls from the raw routing table and return it
+ *
+ * @param {RawRoutingTable} rawRoutingTable the routing table
+ * @param {string} routerAddress the router address
+ * @returns {number} the ttl
+ */
+function parseTtl (rawRoutingTable, routerAddress) {
```
2hdddg: Looks more like it is calculating expire time than parsing ttl

src/internal/routing-table-getter/routing-table-factory.js
```diff
@@ -0,0 +1,210 @@
...
+  try {
+    const now = int(Date.now())
+    const expires = int(rawRoutingTable.ttl)
+      .multiply(1000)
+      .add(now)
+    // if the server uses a really big expire time like Long.MAX_VALUE this may have overflowed
+    if (expires.lessThan(now)) {
```
2hdddg: Looks a bit dangerous, maybe you can do the calculations in seconds instead to get read of the multiplication and potential overflow.

src/internal/routing-table-getter/routing-table-factory.js
```diff
@@ -0,0 +1,210 @@
...
+      `Unable to parse TTL entry from router ${routerAddress} from raw routing table:\n${JSON.stringify(
+        rawRoutingTable
+      )}\nError message: ${error.message}`,
+      PROTOCOL_ERROR
+    )
+  }
+}
+
+/**
+ * Assset if serverAddressesArray is not empty, throws and PROTOCOL_ERROR otherwise
```
2hdddg: Assert

src/internal/routing-table-getter/routing-table-factory.js
```diff
@@ -0,0 +1,210 @@
...
+      } else if (role === 'WRITE') {
+        writers = parseArray(addresses).map(address =>
+          ServerAddress.fromUrl(address)
+        )
+      } else if (role === 'READ') {
+        readers = parseArray(addresses).map(address =>
+          ServerAddress.fromUrl(address)
+        )
+      } else {
+        throw newError('Unknown server role "' + role + '"', PROTOCOL_ERROR)
```
2hdddg: I would prefer to not consider unknowns an error, as long as we get the roles that we want it is probably fine to go on

src/internal/routing-table-getter/routing-table-getter-factory.js
```diff
@@ -43,14 +45,20 @@ export default class RoutingTableGetterFactory {
...
    * @param {string} routerAddress the URL of the router.
-   * @returns {ProcedureRoutingTableGetter} The routing table getter
+   * @returns {ProcedureRoutingTableGetter | RouteMessageRoutingTableGetter} The routing table getter
    */
   create (connection) {
-    const runner =
-      connection.protocol().version < BOLT_PROTOCOL_V4_0
-        ? new SingleDatabaseRoutingProcedureRunner()
-        : new MultiDatabaseRoutingProcedureRunner(this._initialAddress)
+    if (connection.protocol().version < BOLT_PROTOCOL_V4_3) {
```
2hdddg: I know this is old code but isn't the point of having a bolt protocol version inheritance structure to handle things like this?

src/internal/routing-table-getter/routing-table-factory.js
bigmontz: It's a good point, i will skip if it's unknown

src/internal/routing-table-getter/routing-table-getter-factory.js
bigmontz: Are you saying about the factory code or all the get routing table implementation?

2hdddg: Just questioning the need for a factory that checks protocol versions when there already are a protocol inheritance hierarchy that should know type of routing to use for that specific protocol version.

src/internal/routing-table-getter/routing-table-factory.js
bigmontz: I already have test which considers the overflow and it puts the `Integer.MAX_VALUE` as the expiration time. It is the current behaviour of the routing table expiration time nowadays.

src/internal/routing-table-getter/routing-table-getter-factory.js
bigmontz: Yes, the question is really pertinent. This code factory was created by me on the previous pr, the code before was doing this check really deep on the `RoutingUtil` class and it was making the code inflexible (see #645). 

I think almost all the code should go to the protocol (at the least until the part which get the raw routing table). I just need to think about how to handle some circular dependencies which could occurs if the code got moved as it is.

PR #650, opened 2020-12-14
teskit: Fine tunning stress tests to run in testkit

bigmontz:
Fix the way the url is generated during the stress test enabling to use configurations came from the test environment.
Skipping the non-cluster-safe stress test commands to avoid unstable behaviours and flaky tests.

### DIFF
--- a/src/internal/connection-provider-routing.js
+++ b/src/internal/connection-provider-routing.js
@@ -37,7 +37,7 @@ const UNAUTHORIZED_ERROR_CODE = 'Neo.ClientError.Security.Unauthorized'
 const DATABASE_NOT_FOUND_ERROR_CODE =
   'Neo.ClientError.Database.DatabaseNotFound'
 const SYSTEM_DB_NAME = 'system'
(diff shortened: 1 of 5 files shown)


src/internal/connection-provider-routing.js
```diff
@@ -37,7 +37,7 @@ const UNAUTHORIZED_ERROR_CODE = 'Neo.ClientError.Security.Unauthorized'
 const DATABASE_NOT_FOUND_ERROR_CODE =
   'Neo.ClientError.Database.DatabaseNotFound'
 const SYSTEM_DB_NAME = 'system'
-const DEFAULT_DB_NAME = ''
+const DEFAULT_DB_NAME = null
```
2hdddg, line 40: Was this discovered with new default db test? I didn't see any failure

2hdddg (approved):

src/internal/connection-provider-routing.js
bigmontz: I discovered it when I test against clusters using testkit locally.

PR #653, opened 2020-12-15
routing: Fix default database name in the routing connection provider

bigmontz:
A RoutingTableRegistry was extracted from the Connection Provider to help re-use
some behaviours.

### DIFF
--- a/src/internal/connection-provider-routing.js
+++ b/src/internal/connection-provider-routing.js
@@ -63,7 +63,6 @@ export default class RoutingConnectionProvider extends PooledConnectionProvider 
     })
 
     this._seedRouter = address
-    this._routingTables = {}
     this._rediscovery = new Rediscovery(routingContext, address.toString())
     this._loadBalancingStrategy = new LeastConnectedLoadBalancingStrategy(
       this._connectionPool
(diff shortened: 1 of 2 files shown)


src/internal/connection-provider-routing.js
```diff
@@ -206,36 +207,40 @@ export default class RoutingConnectionProvider extends PooledConnectionProvider
...
-      this._routingTables[database].forgetWriter(address)
-    } else {
-      Object.values(this._routingTables).forEach(routingTable =>
-        routingTable.forgetWriter(address)
-      )
-    }
+    this._routingTableRegistry.apply(database, {
+      applyWhenExists: routingTable => routingTable.forgetWriter(address),
+      applyWhenDontExists: () => {
+        throw newError(
```
2hdddg: Are database routing tables never removed? If they are we could end up throwing here. The old code seems to assume that they are never cleaned up so you are probably fine.

src/internal/connection-provider-routing.js
```diff
@@ -501,3 +501,96 @@ export default class RoutingConnectionProvider extends PooledConnectionProvider
...
+    return typeof defaultSupplier === 'function'
+      ? defaultSupplier()
+      : defaultSupplier
+  }
+
+  /**
+   * Remove the routing table which is already expired
+   * @returns {RoutingTableRegistry} this
+   */
+  removeExpired () {
```
2hdddg, line 560: Oh! Here you remove a routing table for a database. Throwing in comment above should probably not be done. Could be a timeout that kicks in after a long time when the routing table has been removed and in that case you would get wrong type of exception.

src/internal/connection-provider-routing.js
bigmontz: Yes, they assumes it always exists when it came this path. I think we could just ignore if doesn't exist since it's not the case something will stop to work if it tries to forget a writer for a non-existing routing table.

src/internal/connection-provider-routing.js
bigmontz: Agree. This behaviour did not change, but throwing the exception in that situation could stop something to work because of a problem which not exists at all.

PR #655, opened 2020-12-16
Make the stress test more reliable

bigmontz:
* [x] Avoid session.run usage
* [x] Treat the session.close() exception in the command. 
* [x] Check the number of commands run

### DIFF
--- a/test/stress.test.js
+++ b/test/stress.test.js
@@ -110,6 +110,7 @@ describe('#integration stress tests', () => {
         }
 
         verifyServers(context)
+          .then(() => verifyCommandsRun(context, TEST_MODE.commandsCount))
           .then(() => verifyNodeCount(context))
           .then(() => done())
           .catch(error => done.fail(error))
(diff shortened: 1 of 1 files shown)


test/stress.test.js
```diff
@@ -387,10 +397,10 @@ describe('#integration stress tests', () => {
     }
   }
 
-  function verifyQueryResult (result) {
+  function verifyQueryResult (result, context) {
     if (!result) {
       return new Error('Received undefined result')
-    } else if (result.records.length === 0) {
+    } else if (result.records.length === 0 && context.writeCommandsRun < 20) {
```
2hdddg: Why 20?

2hdddg (approved):
Perhaps simpler to do something like process.exit (after printing something useful) whenever an error is encountered, no need for running anymore anyway and more clear.

test/stress.test.js
bigmontz: It was totally arbitrary. I tried to put whenever we had run any write commands, but it doesn't work because of the parallelism and we don't know if the write command finished before the read starts. And since we have 24 executors, 20 seems to be a reasonable number to start to check if the length is not zero. 

I will put this as function of the number of executors.

bigmontz:
> Perhaps simpler to do something like process.exit (after printing something useful) whenever an error is encountered, no need for running anymore anyway and more clear.

When the callback is called with an error, the process just stopped and back to the `it("basic",..` and fails calling the `done.fail(error)`. It's not that obvious, depends a bit on the knowledge about how the `parallelLimit` works.

PR #657, opened 2020-12-22
Providing protocol version in Result get by Transaction.run

bigmontz:
The result summary was not filled with the protocol version because the transactions were creating
the Result object with an empty connection holder. Since the Result releases the connection from holder
when it completes, the original holder could not be passed to it.

The ReadOnlyConnectionHolder provides a safe way to pass the holder forward without let the consumer initialize,
close or release the connection by simulating this methods without do the action and delegating part of the requests to the original ConnectionHolder.

These changes also made shared-neo4j cluster friendly.

### DIFF
--- a/src/internal/connection-holder-readonly.js
+++ b/src/internal/connection-holder-readonly.js
@@ -1,0 +1,77 @@
+/**
+ * Copyright (c) 2002-2020 "Neo4j,"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import ConnectionHolder from './connection-holder'
+
+/**
+ * Provides a interaction with a ConnectionHolder without change it state by
+ * releasing or initilizing
+ */
+export default class ReadOnlyConnectionHolder extends ConnectionHolder {
+  /**
+   * Contructor
+   * @param {ConnectionHolder} connectionHolder the connection holder which will treat the requests
+   */
+  constructor (connectionHolder) {
+    super({
+      mode: connectionHolder._mode,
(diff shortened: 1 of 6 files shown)


src/internal/connection-holder-readonly.js
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Provides a interaction with a ConnectionHolder without change it state by
+ * releasing or initilizing
+ */
+export default class ReadOnlyConnectionHolder extends ConnectionHolder {
+  /**
+   * Contructor
+   * @param {ConnectionHolder} connectionHolder the connection holder which will treat the requests
+   */
+  constructor (connectionHolder) {
```
2hdddg, line 30: Both inherits and aggregates connection holder. Is that necessary?

src/result.js
```diff
@@ -180,7 +180,9 @@ class Result {
       connectionHolder.getConnection().then(
         // onFulfilled:
         connection => {
-          release(connection ? connection.protocol().version : undefined)
+          release(
```
2hdddg: Why should connection be undefined here?
In onRejected at 188 I can see that there is no connection but that one still calls release. Why?

src/transaction.js
```diff
@@ -257,7 +258,12 @@ const _states = {
         })
         .catch(error => new FailedObserver({ error, onError }))
 
-      return newCompletedResult(observerPromise, query, parameters)
+      return newCompletedResult(
+        observerPromise,
+        query,
+        parameters,
+        connectionHolder
```
2hdddg, line 265: Wouldn't it be easier to initialize completed result with protocol version instead of connection holder? I mean the transaction should have used a valid connection at some point (like _begin should have been called succesfully) and wherever that happens the protocol version could be stashed away and reused here.

src/result.js
bigmontz: The release is completing the result and release the connection in some cases. It's doing two things. I think it could be splited.

About checking the connection, the get connection should not resolve it a null connection, so this check should could be removed.

src/transaction.js
bigmontz: Yes, it could be simpler. I will check if i have it unwrapped outside a promise some where on the session.run and on the transaction.run

src/internal/connection-holder-readonly.js
bigmontz: I'd like to re-use part of the methods and behaviours.

src/transaction.js
bigmontz: In the moment the result is created, we only have a promise of connection, not the connection.

2hdddg: Haven't _begin executed? I guess it must execute before run somehow..

bigmontz: Yes, but in the Session.run, we don't have it unwrapped. We could inform a promise of protocol version.

PR #659, opened 2020-12-23
Improve connection/protocol layers

bigmontz:
The implementation of the connection channel, the protocol and handshake are dependent in the other in cross mutual dependency situation. This cross dependency situation could leave to unexpected behaviours when something change on this layers.

The solution for this issue was separate the bolt protocol with it handshake process, creation, observers and so on in a different package to hide the details and make easier to spot the unwanted dependencies. 

The following changes were done to achieve these goals:
* [x] Extract a factory for the connection removing the create method
* [x] Extract the handshake process and separate in two, the handshake message exchange and the protocol initialisation
* [x] Remove the response parsing and response observers management from the connection and put this in the bolt package (`response-handler.js`)
* [x] Move the message write responsibility to the protocol object
* [x] Remove connection dependency from the stream-observers
* [x] Fix the connection integration tests
* [x] Improve the unit tests
* [x] Do not touch or break test kit tests and stress tests


src/driver.js
```diff
@@ -369,12 +369,11 @@ function sanitizeIntValue (rawValue, defaultWhenAbsent) {
  */
 function validateFetchSizeValue (rawValue, defaultWhenAbsent) {
   const fetchSize = parseInt(rawValue, 10)
-  if (fetchSize > 0 || fetchSize === ALL) {
+  if (fetchSize > 0 || fetchSize === REQUEST_ALL) {
```
2hdddg: Since you are renaming ALL, maybe it should be FETCH_ALL since fetch is already used all over the place.

src/internal/bolt/bolt-protocol-util.js
```diff
@@ -41,17 +41,17 @@ function assertTxConfigIsEmpty (txConfig, connection, observer) {
 /**
  * Asserts that the passed-in database name is empty.
  * @param {string} database
- * @param {Connection} connection
+ * @param {fuction(err: String)} onProtocolError Called when it doesn't have database set
```
2hdddg, line 44: I guess there are more conditions for protocol error than not having a database (at least in the long run)

src/internal/bolt/bolt-protocol-v1.js
```diff
@@ -91,16 +109,13 @@ export default class BoltProtocol {
    */
   initialize ({ userAgent, authToken, onError, onComplete } = {}) {
     const observer = new LoginObserver({
-      connection: this._connection,
+      beforeError: err => this._onProtocolError(err.message),
```
2hdddg: beforeError sounds a bit magic, how would we know ;-)
Clearer to just call it what it is, onProtocolError
The existing afterError is also strangely named, runtimeError?

src/internal/bolt/bolt-protocol-v1.js
```diff
@@ -302,4 +313,68 @@ export default class BoltProtocol {
...
   }
+
+  /**
+   * Write a message to the network channel.
+   * @param {RequestMessage} message the message to write.
+   * @param {ResultStreamObserver} observer the response observer.
+   * @param {boolean} flush `true` if flush should happen after the message is written to the buffer.
+   */
+  write (message, observer, flush) {
+    const queued = this._queueObserver(observer)
```
2hdddg: Maybe the write function would be clearer if you expand the _queueObserver function in this function. Handling of the _fatalError is a bit obscured by something that looks optional. Or at least rename _queueObserver...

src/internal/bolt/bolt-protocol-v1.js
```diff
@@ -262,16 +277,12 @@ export default class BoltProtocol {
...
+    assertDatabaseIsEmpty(database, this._onProtocolError, observer)
 
-    this._connection.write(
-      RequestMessage.run(query, parameters),
-      observer,
-      false
-    )
-    this._connection.write(RequestMessage.pullAll(), observer, flush)
+    this.write(RequestMessage.run(query, parameters), observer, false)
+    this.write(RequestMessage.pullAll(), observer, flush)
```
2hdddg, line 289: I guess this is the same as the old code but the observer will be completed trice, once for run and once for pullAll. Would it be simpler in the impl of observer to receive only one completed (but errors from both)? Then you can create a temp observer for run message.

src/internal/bolt/bolt-protocol-v3.js
```diff
@@ -56,22 +56,19 @@ export default class BoltProtocol extends BoltProtocolV2 {
...
       afterError: onError,
       afterComplete: onComplete
     })
 
-    this._connection.write(
-      RequestMessage.hello(userAgent, authToken),
-      observer,
-      true
-    )
+    this.write(RequestMessage.hello(userAgent, authToken), observer, true)
```
2hdddg, line 63: These calls look a lot nicer!

src/internal/bolt/handshake.js
```diff
@@ -0,0 +1,145 @@
...
+
+const createHandshakeMessageMemoized = memoize(createHandshakeMessage)
+
+/**
+ * @return {BaseBuffer}
+ * @private
+ */
+function newHandshakeBuffer () {
+  return createHandshakeMessageMemoized([
+    [BOLT_PROTOCOL_V4_3, BOLT_PROTOCOL_V4_2],
```
2hdddg: Maybe simpler to not import the BOLT_xxx things and just write the major/minor versions here, we kind of assume the values of them anyway here, no need for verionify with Math.floor or memoize...

2hdddg (commented):
Looks like good improvements! Got halfway through the changes, need coffee!!

src/internal/bolt/bolt-protocol-util.js
bigmontz: Yes, but in the case of this function (`assertDatabaseIsEmpty`), it only will call the callback in this condition

src/internal/bolt/bolt-protocol-v1.js
bigmontz: Or we could go with `onError`. In the original LoginObserver, it did something like:

```
if(this.beforeError)
   this.beforeError(error)
// change some state on the connection
if (this.afterError)
  this.afterError(error)
```

Since we removed the part on the middle, i think the methods should join together as onError and then the initialize method should handle the call to the this._onProtocolError and redirect the on error call.

src/internal/bolt/bolt-protocol-v1.js
bigmontz: I will make this more evident.

src/internal/bolt/bolt-protocol-v1.js
bigmontz: It receive both, the `ResultStreamObserver` is stateful and move the internal state machine in accordance with the events received. 

The client which called run, you only receive one event in the end because the `ResultStreamObserver` will handle it.

2hdddg: Just meant that it is a bit "leaky" to the observer that pipelining is in use or that more than one command is needed on bolt level at all. But maybe ResultStreamObserver kicks off some other event when RUN is acked

src/internal/bolt/handshake.js
bigmontz: Did you mean like a tuple, for instance, 4.3 being [4, 3]?

src/internal/bolt/bolt-protocol-v1.js
bigmontz: Yes, now i've got it. 

The chain of observers looks like something that should be done in the protocol class (the one sent the message). 

I think It's a good improvement, but i will let this for future work since this part is quite isolated and it's not impacting much in the other parts of the software outside the `bolt-protocol` folder.

PR #662, opened 2021-01-13
Remove circular dependencies

bigmontz:
Remove circular dependencies to decouple modules.

Circular dependencies overview: 

circular

### DIFF
--- a/src/internal/node/node-utf8.js
+++ b/src/internal/node/node-utf8.js
@@ -17,7 +17,6 @@
  * limitations under the License.
  */
 
-import CombinedBuffer from '../buf/combined-buf'
 import NodeBuffer from './node-buf'
 import { newError } from '../../error'
 import node from 'buffer'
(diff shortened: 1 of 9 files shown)


src/internal/node/node-utf8.js
```diff
@@ -30,9 +29,9 @@ function encode (str) {
 }
 
 function decode (buffer, length) {
-  if (buffer instanceof NodeBuffer) {
+  if (Object.prototype.hasOwnProperty.call(buffer, '_buffer')) {
```
2hdddg, line 32: Looks like there should be a decode function on NodeBuffer and CombinedBuffer

2hdddg (approved):
Looks good! I assume temporal functions has been more or less copied ?

src/internal/node/node-utf8.js
bigmontz: Yes, It should be a method present in all the buffers versions, the issue is way it's done vary a bit between buffer types, I could not figure out a way to do it properly.

PR #667, opened 2021-01-21
Run stress tests out of Jasmine

bigmontz:
The goal of this change is improve the stability of the stress test avoiding unexpected timeouts.

### DIFF
--- a/test/stress-test.js
+++ b/test/stress-test.js
@@ -1,0 +1,743 @@
+/**
+ * Copyright (c) 2002-2020 "Neo4j,"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import neo4j from '../src'
+import { READ, WRITE } from '../src/driver'
+import parallelLimit from 'async/parallelLimit'
+import _ from 'lodash'
+import sharedNeo4j from './internal/shared-neo4j'
+import { resolve } from 'uri-js'
+
+const TEST_MODES = {
+  fastest: {
+    commandsCount: 10000,
+    parallelism: 24,
+    maxRunTimeMs: 3 * 60000 // 3 minutes
+  },
+  fast: {
+    commandsCount: 5000,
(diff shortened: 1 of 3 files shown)


test/stress-test.js
```diff
@@ -0,0 +1,743 @@
...
+
+function addressesWithRole (records, role) {
+  return _.uniq(
+    records
+      .filter(record => record.get('role') === role)
+      .map(record => record.get('addresses')[0].replace('bolt://', ''))
+  )
+}
+
+function assertAllAddressesServedReadQueries (addresses, readQueriesByServer) {
```
2hdddg: I'm don't believe that is safe to do this check if the cluster is reforming while the tests are running we might end up with new members of the cluster right before this check.
Load balancing algorithm is better tested with other means..

test/stress-test.js
```diff
@@ -0,0 +1,743 @@
...
+  addresses.forEach(address => {
+    const queries = readQueriesByServer[address]
+    if (queries <= 0) {
+      throw Error(`Expect queries to be greater then 0, but it is ${queries}`)
+    }
+  })
+}
+
+function assertAllAddressesServedSimilarAmountOfReadQueries (
```
2hdddg: See above

test/stress-test.js
```diff
@@ -0,0 +1,743 @@
...
+ */
+
+import neo4j from '../src'
+import { READ, WRITE } from '../src/driver'
+import parallelLimit from 'async/parallelLimit'
+import _ from 'lodash'
+import sharedNeo4j from './internal/shared-neo4j'
+import { resolve } from 'uri-js'
+
+const TEST_MODES = {
```
2hdddg, line 26: Would be nice to be able to configure this from testkit invocation.
I'm not fully convinced that specifying the number of commands is the best way to run this in stress tests but  it will work in drop pipeline.

From testkit it would be easier to specify the number of seconds that the suite should be running. The scenario is: 
* Start the cluster
* Start stress test suite
* Wait x seconds
* Add another member to cluster
* Wait y seconds
* Remove a member from the cluster
* Wait z seconds

In this case we need to know that the test runs for a bit longer than x+y+z seconds otherwise testkit should fail the test since it cannot verify that the driver could handle rolling cluster members.

test/stress-test.js
```diff
@@ -0,0 +1,743 @@
...
+  fastest: {
+    commandsCount: 10000,
+    parallelism: 24,
+    maxRunTimeMs: 3 * 60000 // 3 minutes
```
2hdddg: Is maxRunTimeMs used?

test/stress-test.js
bigmontz: Sure, I will remove all the `verifyCausalClusterMembers` function, since it's could cause problem.

test/stress-test.js
bigmontz: Not anymore, it should be removed.

test/stress-test.js
bigmontz: Yes, I will do this one today in a second iteration, this PR was more about porting the existing code.

bigmontz: Solved in this PR

PR #670, opened 2021-01-27
[porting] stress-test: Improve Node Count verification for work better in clusters (#669)

bigmontz:


test/stress-test.js
```diff
@@ -449,7 +449,7 @@ function verifyNodeCount (context) {
 
   const session = context.driver.session()
   return session
-    .readTransaction(tx => tx.run('MATCH (n) RETURN count(n)'))
+    .writeTransaction(tx => tx.run('MATCH (n) RETURN count(n)'))
```
fbiville, line 452: that's... counter-intuitive :)
out of curiosity, is the change done to force the query to be routed to the cluster leader?

PR #680, opened 2021-02-11
testkit: Update NodeJS version in the Dockerfile

bigmontz:
The support for the `NodeJS 8.x` was dropped and the `npm update` was failing because of that.

The NodeJS upgrade is being done to `10.x` since it's the oldest version in maintenance.

This update was done by updating the Ubuntu version since the other of updates was broken the TLS configuration. Along with this, the TLS 1.0 was enable in the OpenSSL config since it is disabled by default in this ubuntu's version.

### DIFF
--- a/testkit/Dockerfile
+++ b/testkit/Dockerfile
@@ -1,18 +1,30 @@
-FROM ubuntu:18.04
+FROM ubuntu:20.04
+
+ENV DEBIAN_FRONTEND noninteractive
 
 RUN apt-get update && \
     apt-get install -y \
         git \
         curl \
         python3 \
-        firefox \
         nodejs \
-        npm \
+        npm \ 
+        firefox \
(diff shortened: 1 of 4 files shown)


testkit/Dockerfile
```diff
@@ -10,6 +10,8 @@ RUN apt-get update && \
         npm \
     && rm -rf /var/lib/apt/lists/*
 
+RUN npm install -g n
+RUN n 10
```
fbiville: is this a typo?

bigmontz: Not this time. 

https://www.npmjs.com/package/n

fbiville: what a great package name #not 😅
interesting, I've only used `nvm` for this.
TIL!

bigmontz: yeah, It remembers the var names I used to use in algorithms classes.

PR #696, opened 2021-03-29
Add support for BigInt

bigmontz:
Recent V8 release adds a new type - BigInt. It is thus available in latest NodeJS (v10+) and Chrome (v67+). 

Driver could use this type instead of the custom Integer to represent lossy integer numbers. This is configurable because BigInt and Number are not fully compatible, so driver can't always return BitInts. For example `BigInt(1) / Number(1)` results in a TypeError, as well as `JSON.stringify({v: BigInt(1)})`.

The configuration property is `useBigInt` and it's `false` by default, in manner to break drivers user on update. 

This configuration take precedence over `disableLosslessIntegers`

Compatibility matriz: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt


fbiville:
Are we OK with not supporting IE if big int usage is enabled?

bigmontz:
> Are we OK with not supporting IE if big int usage is enabled?

I will check what's happen when it is not available and the flag is disabled. It should not fail.

bigmontz:
@fbiville, I've check in environment which doesn't support bigint, and if the `useBigInt` is not enabled, the BigInt constructor is not called and nothing undesired happen. If you enable, it will explode.

testkit-backend/src/cypher-native-binders.js
```diff
@@ -58,7 +59,7 @@ export function cypherToNative (c) {
     case 'CypherString':
       return value
     case 'CypherInt':
-      return int(value)
+      return BigInt(value)
```
oskarhane, line 62: There's no conditional here, which would imply that BigInt always is used.
I'm not very familiar with this part of the code so I might be mistaken.

oskarhane (commented):
Just one comment where I think we're missing a `useBigInt` check.

In general, I like the move towards this over the customer integer type we currently have.  

Just a side note:
In most of the applications we build we use web workers to get the result and transfer the data via `postMessage` with the result serialized as JSON.  

Surprisingly BigInt's can be serialized at the moment so this kinds of usages of the driver need to add a custom `.toJSON()` to the big int results. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#use_within_json

testkit-backend/src/cypher-native-binders.js
bigmontz: It's not an issue in this, because this file is part of the testkit-backend, a test server.

bigmontz: About the JSON serialization, I could add some notes on the documentation to users be aware of this when enable BigInt.

oskarhane: Ah, great, that's what I figured.
Yes, I think a note on serialization in the docs would be good. Thanks!

test/driver.test.js
```diff
@@ -26,6 +26,7 @@ import {
 } from '../bolt-connection/lib/pool/pool-config'
 import { ServerVersion, VERSION_4_0_0 } from '../src/internal/server-version'
 import testUtils from './internal/test-utils'
+import { json as JSON } from 'neo4j-driver-core'
```
OskarDamkjaer: It's a detail, any specific reason some imports shadow JSON here and change all uses to `json` instead?

core/src/json.ts
```diff
@@ -17,17 +17,14 @@
...
-
-    expect(protocol.version).toBe(2)
-  })
-})
+/**
+ * Provides a interface to serialize values which doesn't support serialization by JSON.stringify
+ * @private
+ * @param val The value to be serialized as JSON
+ * @returns the json
+ */
```
OskarDamkjaer, line 25: Perhaps this doc comment could be updated to something like:

Custom version on JSON.stringify that can handle values that normally don't support serialization, such as BigInt.

@param val — A JavaScript value, usually an object or array, to be converted.
@returns A JSON string representing the given value.

core/src/json.ts
```diff
@@ -17,17 +17,14 @@
...
+export function stringify (val: any) {
+  return JSON.stringify(val, (_, value) =>
+    typeof value !== 'bigint' ? value : `${value}n`
```
OskarDamkjaer: Personally I find it clearer to read ternaries as ```typeof value === 'bigint' ? `${value}n` : value```

core/src/internal/temporal-util.ts
```diff
@@ -575,3 +580,12 @@ function formatNumber (
   }
   return isNegative ? '-' + numString : numString
 }
+
+function add (x: NumberOrInteger, y: number): NumberOrInteger {
+  if (x instanceof Integer) {
```
OskarDamkjaer, line 585: any reason functions have a space before listing arguments in this file?

bolt-connection/test/bolt/bolt-protocol-v4x3.test.js
```diff
@@ -17,21 +17,22 @@
...
-import RequestMessage from '../../../bolt-connection/lib/bolt/request-message'
+import BoltProtocolV4x3 from '../../src/bolt/bolt-protocol-v4x3'
+import RequestMessage from '../../src/bolt/request-message'
 import utils from '../test-utils'
-import { WRITE } from '../../../src/driver'
-import { RouteObserver } from '../../../bolt-connection/lib/bolt/stream-observers'
+import { RouteObserver } from '../../src/bolt/stream-observers'
 import { internal } from 'neo4j-driver-core'
 
+const WRITE = 'WRITE'
```
OskarDamkjaer, line 26: Don't see why this variable is needed

bolt-connection/test/bolt/bolt-protocol-v4x0.test.js
```diff
@@ -17,25 +17,26 @@
...
-import { WRITE } from '../../../src/driver'
 import {
   ProcedureRouteObserver,
   ResultStreamObserver
-} from '../../../bolt-connection/lib/bolt/stream-observers'
+} from '../../src/bolt/stream-observers'
 
 import { internal } from 'neo4j-driver-core'
 
+const WRITE = 'WRITE'
```
OskarDamkjaer, line 30: Not sure why this variable is needed, the string seems to be inlined in other changes

OskarDamkjaer:
I don't have a full view of how the driver/BigInt works, but looking through the code I didn't find anything that seemed off. I left some small code comments, feel free to disregard if you don't find them useful

core/src/internal/temporal-util.ts
bigmontz: It is in the repository codestyle and it's done automatically. But it's a bit useful when you're searching for methods use or declarations.

test/driver.test.js
bigmontz: Sorry, It's a leftover.

core/src/json.ts
bigmontz: It makes sense, It was my `C/C++` accent.

bolt-connection/test/bolt/bolt-protocol-v4x0.test.js
bigmontz: Actually I prefer this `const` solution, it's cleaner and easy to change if the value change.

PR #702, opened 2021-04-06
Fix HELLO message RoutingContext and send bookmarks in the ROUTE message

bigmontz:
The routing context was missing information about the initial routing server. This information is important for the server side routing.

The bookmarks in the ROUTE message are important in the case of a creation of a new database to make sure the database is running in all the clusters nodes returned.


gjmwoods (approved):
💪

PR #719, opened 2021-04-19
Adjusting build scripts and add ts-node to neo4j-driver dev-deps

bigmontz:


buildDependencies.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 
-npm --prefix core ci
-npm --prefix core run build
-npm --prefix bolt-connection ci
-npm --prefix bolt-connection run build
+cd core 
+npm ci
+npm run build 
+cd ../bolt-connection
```
fbiville, line 6: what about `npm --prefix core ...` and `npm --prefix ../bolt-connection`?

fbiville (approved):

buildDependencies.sh
bigmontz: The `--prefix` was generating packages not compatible with npm 6.x which is legacy but it still broadly used.

PR #740, opened 2021-05-20
Fix `Driver.supportsMultiDb` return type

bigmontz:
Also add a explicity return type for `supportsTransactionConfig`


jharris4 (approved):
Thank you for fixing this!

PR #761, opened 2021-07-22
Connection Hint: supporting connection.recv_timeout_seconds

bigmontz:
This value defines the time the client should wait for a response on a connection before it times out.

Co-authored-by: Rouven Bauer <rouven.bauer@neotechnology.com>

### DIFF
--- a/bolt-connection/src/channel/browser/browser-channel.js
+++ b/bolt-connection/src/channel/browser/browser-channel.js
@@ -172,7 +172,12 @@ export default class WebSocketChannel {
   }
 
   /**
-   * @todo document it
+   * Setup the receive timeout for the channel.
+   *
+   * Not supported for the browser channel.
+   *
+   * @param {number} receiveTimeout The amount of time the channel will keep without receive any data before timeout (ms)
+   * @returns {void}
    */
   setupReceiveTimeout (receiveTimeout) {}
 
--- a/bolt-connection/src/channel/browser/browser-channel.js
+++ b/bolt-connection/src/channel/browser/browser-channel.js
@@ -172,6 +172,11 @@ export default class WebSocketChannel {
   }
 
   /**
+   * @todo document it
+   */
+  setupReceiveTimeout (receiveTimeout) {}
+
+  /**
    * Set connection timeout on the given WebSocket, if configured.
    * @return {number} the timeout id or null.
    * @private
--- a/bolt-connection/src/channel/node/node-channel.js
+++ b/bolt-connection/src/channel/node/node-channel.js
@@ -148,7 +148,7 @@ const TrustStrategy = {
  * @param {function} onFailure - callback to execute on connection failure.
  * @return {*} socket connection.
  */
-function connect (config, onSuccess, onFailure = () => null) {
+function _connect (config, onSuccess, onFailure = () => null) {
   const trustStrategy = trustStrategyName(config)
   if (!isEncrypted(config)) {
     const socket = net.connect(
@@ -230,7 +230,7 @@ export default class NodeChannel {
    * Create new instance
    * @param {ChannelConfig} config - configuration for this channel.
    */
-  constructor (config) {
+  constructor (config, connect = _connect) {
     const self = this
 
     this.id = _CONNECTION_IDGEN++
--- a/bolt-connection/src/channel/node/node-channel.js
+++ b/bolt-connection/src/channel/node/node-channel.js
@@ -305,12 +305,12 @@ export default class NodeChannel {
   _setupConnectionTimeout (config, socket) {
     const timeout = config.connectionTimeout
     if (timeout) {
-      socket.on('connect', () => {
+      const connectListener = () => {
         // connected - clear connection timeout
         socket.setTimeout(0)
-      })
+      }
 
-      socket.on('timeout', () => {
+      const timeoutListener = () => {
         // timeout fired - not connected within configured time. cancel timeout and destroy socket
         socket.setTimeout(0)
         socket.destroy(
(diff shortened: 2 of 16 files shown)


bolt-connection/src/channel/node/node-channel.js
```diff
@@ -319,12 +319,43 @@ export default class NodeChannel {
...
+      }
 
       socket.setTimeout(timeout)
     }
   }
 
+  /**
+   * Setup the receive timeout for the channel.
+   *
+   * @param {number} receiveTimeout The amount of time the channel will keep without receive any data before timeout (ms)
```
robsdedude:
```suggestion
   * @param {number} receiveTimeout How long the channel will wait for receiving data before timing out (ms)
```

bolt-connection/src/channel/browser/browser-channel.js
```diff
@@ -171,6 +171,16 @@ export default class WebSocketChannel {
...
+  /**
+   * Setup the receive timeout for the channel.
+   *
+   * Not supported for the browser channel.
+   *
+   * @param {number} receiveTimeout The amount of time the channel will keep without receive any data before timeout (ms)
+   * @returns {void}
+   */
+  setupReceiveTimeout (receiveTimeout) {}
+
```
robsdedude, line 183: Most certainly out of scope of this PR, but it should be possible to implement a timeout mechanism for websockets with a wrapper that uses `window.setTimeout`. I wouldn't be surprised if someone has published a library or code-snippets for exactly this.

bolt-connection/src/connection/connection-channel.js
```diff
@@ -198,6 +198,25 @@ export default class ChannelConnection extends Connection {
...
+                receiveTimeoutRaw !== null &&
+                receiveTimeoutRaw !== undefined
+              ) {
+                const receiveTimeoutInSeconds = toNumber(receiveTimeoutRaw)
+                if (receiveTimeoutInSeconds > 0) {
+                  this._ch.setupReceiveTimeout(receiveTimeoutInSeconds * 1000)
+                } else {
+                  this._log.info(
+                    `Server located at ${this._address} supplied an invalid connection receive timeout value (${receiveTimeoutInSeconds}). ` +
+                      'Please, verify the server configuration and status because this can a symptom of a bigger issue.'
```
robsdedude:
```suggestion
                    `Server located at ${this._address} supplied an invalid connection receive timeout value (${receiveTimeoutInSeconds}). ` +
                      'Please, verify the server configuration and status because this can be the symptom of a bigger issue.'
```

bolt-connection/test/channel/node/node-channel.test.js
```diff
@@ -0,0 +1,138 @@
...
+        [
+          newError(
+            "Connection lost. Server didn't respond in 42ms",
+            SERVICE_UNAVAILABLE
+          )
+        ]
+      ])
+    })
+
+    it('should not unsubscribe to the any on connect or on timeout if connectionTimeout is not set', () => {
```
robsdedude:
```suggestion
    it('should not unsubscribe from on connect nor from on timeout if connectionTimeout is not set', () => {
```

bolt-connection/src/connection/connection-channel.js
```diff
@@ -198,6 +198,25 @@ export default class ChannelConnection extends Connection {
...
+
+            if (metadata.hints) {
+              const receiveTimeoutRaw =
+                metadata.hints['connection.recv_timeout_seconds']
+              if (
+                receiveTimeoutRaw !== null &&
+                receiveTimeoutRaw !== undefined
+              ) {
+                const receiveTimeoutInSeconds = toNumber(receiveTimeoutRaw)
+                if (receiveTimeoutInSeconds > 0) {
```
robsdedude: Not sure if even feasible in JS's limited number type, but we also don't expect floats from the server.

robsdedude (changes requested):

bolt-connection/src/channel/browser/browser-channel.js
bigmontz: It could be done something like keep a var the timestamp with the last received message and check every `receiveTimeout` if the (current timestamp - last received message timestamp) > receiveTimeout.

bolt-connection/src/connection/connection-channel.js
bigmontz: I will create tests with float number to limit this option. One question, should `3.0` be considered an invalid value?

bolt-connection/src/channel/browser/browser-channel.js
robsdedude: Yes. Or have only one timeout that gets canceled and reset every time something is sent over the socket. And when it ever triggers the calback that's equivalent of a socket timeout.

robsdedude: But as said, seems very much out of scope of this PR.

bolt-connection/src/connection/connection-channel.js
robsdedude: I don't know how the JS drivers handles numbers internally. But the server is expected to sent an Integer (strongly typed through packstream). It might well be, however, that at this point in the code, the type information is already lost. In that case, I'd say 3.0 is also ok.

bigmontz: At that point, I already lost this information. Since 3 is the same as 3.0 for JS. So, I'm verifying if it's a Integer using the JS Number definition of integer.

robsdedude (approved):
:computer:

PR #765, opened 2021-08-11
Fix Path unpacking when `useBigInt` is on

bigmontz:
The code was trying to do math between Number and BigInt during the Path unpacking. The usage of `toNumber` convert any BigInt|Number|Integer to number in optimal way.


injectives (approved):
LGMT

PR #769, opened 2021-09-03
Fix issues which are blocking the build pipeline

bigmontz:
The wrong typecasting the `record.ts` was giving a `Type 'any' cannot be used to index type 'Entries'.` during the build. And it was fixed just by adjusting the type casting.

The test case `Result Promise should call finally on error` was not handling correctly the error before call finally and it was giving `Unhandled promise rejection` error during the tests.


injectives (approved):
LGTM

PR #772, opened 2021-09-16
testkit: Speedup the test by enable skip the browser tests

bigmontz:
Browser tests are consuming a huge amount of time and making the testkit PR take a long time just to change unrelated issue.

The env var `TEST_DRIVER_SKIP_BROWSER` could be used to skip the browser tests.

### DIFF
--- a/testkit/integration.py
+++ b/testkit/integration.py
@@ -8,7 +8,8 @@ def run(args):
 
 
 def test_driver():
-    if os.environ.get("TEST_DRIVER_SKIP_BROWSER", False):
+    if (os.environ.get("TEST_DRIVER_SKIP_BROWSER", "false").lower()
+            in ("y", "yes", "t", "true", "1", "on")):
         run(["gulp", "test-browser"])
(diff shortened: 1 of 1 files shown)


testkit/integration.py
```diff
@@ -8,7 +8,8 @@ def run(args):
 
 
 def test_driver():
-    run(["gulp", "test-browser"])
+    if os.environ.get("TEST_DRIVER_SKIP_BROWSER", False):
```
robsdedude: This is equivalent to `if "TEST_DRIVER_SKIP_BROWSER" in os.environ`. Did you maybe mean something along the lines of 
```suggestion
    if (os.environ.get("TEST_DRIVER_SKIP_BROWSER", "false").lower()
            in ("y", "yes", "t", "true", "1", "on")):
```

robsdedude (commented):
:rocket:

testkit/integration.py
bigmontz: Thanks, It does make sense.

robsdedude (approved):

PR #773, opened 2021-09-16
Enable auth tests in TestKit backend

robsdedude:
Kerberos auth should use the `ticket` field instead of the `credentials` field
to transmit the auth token.

Furthermore, extend the TestKit backend to utilize the different auth helper
functions and enable feature flags to test them.

Waiting for https://github.com/neo4j-drivers/testkit/pull/237 to be merged.

### DIFF
--- a/testkit-backend/src/request-handlers.js
+++ b/testkit-backend/src/request-handlers.js
@@ -10,6 +10,27 @@ export function NewDriver (context, data, { writeResponse }) {
     userAgent,
     resolverRegistered
   } = data
+  let parsedAuthToken = authToken
+  switch (authToken.scheme) {
+    case 'basic':
+      parsedAuthToken = neo4j.auth.basic(
+        authToken.principal,
+        authToken.credentials,
+        authToken.realm
+      )
+      break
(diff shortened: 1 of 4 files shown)


robsdedude:
Turns out this change is not needed, and that `credentials` is actually the correct field to have the kerberos token in. Still, this PR has value in that it enables the according TestKit tests.

testkit-backend/src/request-handlers.js
```diff
@@ -10,14 +10,35 @@ export function NewDriver (context, data, { writeResponse }) {
...
+  switch (authToken.scheme) {
+    case 'basic':
+      parsedAuthToken = neo4j.auth.basic(
+        authToken.principal,
+        authToken.credentials,
+        authToken.realm
+      )
+      break
+    case 'kerberos':
+      parsedAuthToken = neo4j.auth.kerberos(authToken.ticket)
```
robsdedude:
```suggestion
      parsedAuthToken = neo4j.auth.kerberos(authToken.credentials)
```

bigmontz (approved):

PR #781, opened 2021-10-05
Add type assertion to is functions

thebestnom:
Allow for easier use in typescript of is function

### DIFF
--- a/core/src/graph-types.ts
+++ b/core/src/graph-types.ts
@@ -105,7 +105,7 @@ Object.defineProperty(
  * @param {Object} obj the object to test.
  * @return {boolean} `true` if given object is a {@link Node}, `false` otherwise.
  */
-function isNode(obj: object): boolean {
+function isNode(obj: object): obj is Node {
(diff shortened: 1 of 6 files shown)


bigmontz (commented):
Thanks for the contribution. 

It doesn't see to have harm in the changes you did, but do you mind to add unit some tests? 

It could be something like:

```
const potentialNode :object = getSomeNode()
if (isNode(potentialNode)) {
   potentialNode.identity
}
```

Apart of this, it looks really nice.

thebestnom:
@bigmontz added the "test" 😅
didn't find anywhere to put the other tests (no graph types) is but it shouldn't be any different and I don't think any futther tests will make any different

bigmontz (approved):
🌵 

Thanks for your contribution. 

Post here whenever it gets released.

bigmontz:
Type improvement release in version 4.3.4. 

Release notes: https://github.com/neo4j/neo4j-javascript-driver/wiki/4.3-changelog#434

cc: @thebestnom

PR #785, opened 2021-10-14
Deprecate Integer

bigmontz:
The custom implementation of Integer will be fully replaced by for BigInt in the future version of the driver.


AndyHeap-NeoTech (approved):

robsdedude (approved):
:wastebasket:

PR #796, opened 2021-10-22
testkit: Implements ForcedRoutingTableUpdate

bigmontz:
It also fix the GetRoutingTable request handler adding a default value where there is no routing table registered.

### DIFF
--- a/packages/testkit-backend/src/request-handlers.js
+++ b/packages/testkit-backend/src/request-handlers.js
@@ -307,17 +307,42 @@ export function GetRoutingTable (context, { driverId, database }, wire) {
     driver &&
     driver._getOrCreateConnectionProvider() &&
     driver._getOrCreateConnectionProvider()._routingTableRegistry &&
-    driver._getOrCreateConnectionProvider()._routingTableRegistry.get(database)
+    driver._getOrCreateConnectionProvider()._routingTableRegistry.get(database, () => {
+      return {
+        database,
+        ttl: 0,
+        readers: [],
+        writers: [],
+        routers: []
+      }
+    })
 
   if (routingTable) {
     wire.writeResponse('RoutingTable', {
       database: routingTable.database,
-      ttl: routingTable.ttl,
+      ttl: Number(routingTable.ttl),
       readers: routingTable.readers.map(serverAddressToString),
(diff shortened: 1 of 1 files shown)


packages/testkit-backend/src/request-handlers.js
```diff
@@ -307,17 +307,42 @@ export function GetRoutingTable (context, { driverId, database }, wire) {
...
-    wire.writeError('Could not find routing table')
+    wire.writeError('Driver does not support routing')
+  }
+}
+
+export function ForcedRoutingTableUpdate (context, { driverId, database, bookmarks }, wire) {
+  const driver = context.getDriver(driverId)
+
+  if (driver._getOrCreateConnectionProvider()._freshRoutingTable) {
+    driver._getOrCreateConnectionProvider()._freshRoutingTable ({
```
robsdedude: Is `_freshRoutingTable` the right method here? This TestKit message should cause a new RT to be fetched regardless if the current one is stale or not. `_refreshRoutingTable` sounds more like the right candidate to me :thinking:

bigmontz: The usage of the other method was given strange errors while iterate over servers. I could remove routing table for the given database and the _freshRoutingTable.

packages/testkit-backend/src/request-handlers.js
```diff
@@ -307,17 +307,45 @@ export function GetRoutingTable (context, { driverId, database }, wire) {
...
+    wire.writeError('Driver does not support routing')
+  }
+}
+
+export function ForcedRoutingTableUpdate (context, { driverId, database, bookmarks }, wire) {
+  const driver = context.getDriver(driverId)
+  const provider = driver._getOrCreateConnectionProvider()
+
+  if (provider._freshRoutingTable) {
+    // Removing database from the routing table registry
```
robsdedude, line 338: I'll take the PR as is. But I'll suggest to extend that comment with a few words explaining why you decided to delete the RT and go through `_freshRoutingTable` instead of just calling `_refreshRoutingTable` which might seem like the better solution (but isn't as you commented).

robsdedude (approved):
:electric_plug:

PR #798, opened 2021-10-26
testkit-backend: Add support for testing driver in browser

bigmontz:
The initial implementation of the `testkit-backend` is NodeJS only which is not an issue for the time being. However, It starts to be problematic  with new features being implemented test driven using `testkit` as the source of tests. First, the time for run all the suite increased a lot with the number of tests added and the need of keeping the some integration tests to not let browser uncovered. Second, the driver running in browser was not being tested with the same rigour as in Nodejs.

The following changes were done with the intent to support testkit tests in Browser.

### Refactoring 

The logic of reading and writing in the socket, call the handlers and parse the messages were mixed in the Backend class. This makes difficult to add different protocols, channel implementation or controllers to the backend.

The goal of this change is separate this three concerns in `Channel`, `Controller` and `Protocol` with the `Backend` class as the glue between these concepts. New implementations of the `Controller` could enable this backend to test the driver running in the browser without have to change the logic of parsing the messages between testkit and testkit-backend, for instance.

### Supporting Browser

The first step for supporting testkit in browser was creating a `web socket` server to redirect the requests to a client. `RemoteController` is the class responsible for it, it acts like a proxy redirecting the requests and responses. In the other side, `WebSocketChannel` is responsible for managing the messages exchanges between `browser-backend` and `nodejs-backend` in the browser side. 

The `main` function glued everything together using env variables for selecting the correct channel and controller. 

A bit more of structure had to be taking in place for bundle testkit for running in browser. The bundle is supported by `rollup` and defined in `rollup.config.js`, this runs during the build and the outcome is stored in `public/index.js`. The following env variables are defined during the browser build to enforce the correct configuration:
* `TEST_ENVIRONMENT` = "LOCAL"
* `CHANNEL_TYPE` = "WEBSOCKET"
*  `BACKEND_PORT` = `WEB_SERVER_PORT` (or `8000` if nothing is set in `WEB_SERVER_PORT`)

The last thing done during the browser build process is done by some replacements done in the `package.json`, the `./src/controller/remote.js` and `./src/channel/socket.js` are replaced by their interfaces since the some modules used in this files are really tricky to support in browser and they are not need at browser. 

After we have the browser bundle done, we need to server it. It is done by the `RemoteController` as well.

At this point, you can run `TEST_ENVIRONMENT=REMOTE npm run start-testkit-backend` and open your favorite browser in `localhost:8000` for running testkit tests in browser.  

### Glue everything in Testkit

Last but not least, the testkit scripts were adapt to use the environment `TEST_DRIVER_BROWSER` to decide if the tests should be run in browser and in nodejs. Integration tests were segregated for running only the ones needed for the environment under tests. 

`backend.py` is responsible for configuring correctly the backend using the env-variables and also for starting a headless browser with the browser-backend url.  

### IMPORTANTE NOTE

**This version doesn't support TLS_TESTS and STUB_TESTS since the mock servers doesn't support websocket yet**


packages/testkit-backend/package.json
```diff
@@ -24,10 +30,17 @@
...
-    "neo4j-driver": "4.4.0-dev"
+    "node-static": "^0.7.11",
+    "ws": "^8.2.3"
   },
   "devDependencies": {
-    "esm": "^3.2.25"
+    "@rollup/plugin-commonjs": "^21.0.1",
+    "@rollup/plugin-node-resolve": "^13.0.6",
+    "esm": "^3.2.25",
+    "rollup": "^2.59.0",
```
robsdedude, line 42: Why is `rollup` a dev dependency? Isn't is strictly necessary to run the backend in browser mode since `public/index.js` is not shipped with the package but compiled before start-up?

packages/testkit-backend/src/channel/socket.js
```diff
@@ -0,0 +1,83 @@
...
+import Protocol from './testkit-protocol'
+
+function generateRandomId () {
+  return randomBytes(16).toString('hex')
+}
+
+/**
+ * This is communication channel handles the direct communcation with testkit using it own protocol.
+ *
+ * This implementatio is meant to be runned in the NodeJS, it doesn't support Browser.
```
robsdedude:
```suggestion
 * This implementation is meant to be run in NodeJS, it doesn't support Browser.
```

packages/testkit-backend/src/channel/socket.js
```diff
@@ -0,0 +1,83 @@
...
+import net from 'net'
+import { randomBytes } from 'crypto'
+import Protocol from './testkit-protocol'
+
+function generateRandomId () {
+  return randomBytes(16).toString('hex')
+}
+
+/**
+ * This is communication channel handles the direct communcation with testkit using it own protocol.
```
robsdedude:
```suggestion
 * This is communication channel handles the direct communication with TestKit using its protocol.
```

packages/testkit-backend/src/channel/socket.js
```diff
@@ -0,0 +1,83 @@
+import Channel from './interface'
+import net from 'net'
+import { randomBytes } from 'crypto'
+import Protocol from './testkit-protocol'
+
+function generateRandomId () {
+  return randomBytes(16).toString('hex')
+}
```
robsdedude, line 8: Why a random ID instead of incrementing a counter? It could well be that I didn't understand the structure of this change. I thought there is exactly one channel per backend and TestKit should only ever start one backend.

packages/testkit-backend/src/channel/testkit-protocol.js
```diff
@@ -0,0 +1,80 @@
...
+  _emitRequest () {
+    const request = JSON.parse(this._request)
+    const { name, data } = request
+    console.log('> Got request ' + name, data)
+    this.emit('request', { name, data })
+  }
+  
+  _stringify (val) {
+    return JSON.stringify(val, (_, value) =>
+      typeof value === 'bigint' ? `${value}n` : value
```
robsdedude, line 77: Technically, BigInt is not part of the TestKit protocol and this could, if there is a bug in the request handler implementation, lead to TestKit receiving invalid JSON, couldn't it? I'm not saying this is necessarily a problem, as TestKit would then crash. Just want to invite you to think about. Maybe there is another solution.

packages/testkit-backend/src/channel/websocket.js
```diff
@@ -0,0 +1,62 @@
+import Channel from "./interface"
+
+/**
+ * This communication channels is meant to connect to other instances of the `testkit-backend` for receiving its events.
```
robsdedude:
```suggestion
 * This communication channel is meant to connect to other instances of the `testkit-backend` for receiving its events.
```

packages/testkit-backend/src/channel/websocket.js
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * This channel is only supported in browsers since it depends on WebSocket client be avaiable globaly.
```
robsdedude:
```suggestion
 * This channel is only supported in browsers since it depends on WebSocket client to be available globally.
```

packages/testkit-backend/src/controller/index.js
```diff
@@ -0,0 +1,15 @@
+import Controller from './interface'
+import LocalController from './local'
+import RemoteController from './remote'
+
+/**
+ * Controllers are piece of code responsabile for redicting the requests to the correct handler.
```
robsdedude:
```suggestion
 * Controllers are pieces of code responsible for redirecting requests to the correct handler.
```

packages/testkit-backend/src/channel/index.js
```diff
@@ -0,0 +1,17 @@
+import Channel from "./interface"
+import SocketChannel from "./socket"
+import WebSocketChannel from "./websocket"
+/**
+ * Channels are the piece of code responsible for communicate with tesktit.
```
robsdedude:
```suggestion
 * Channels are the pieces of code responsible for communicating with testkit.
```

packages/testkit-backend/src/channel/interface.js
```diff
@@ -0,0 +1,30 @@
+import { EventEmitter } from "events"
+
+/**
+ * Defines the interface used for receiving commands form the teskit.
```
robsdedude:
```suggestion
 * Defines the interface used for receiving commands form teskit.
```

packages/testkit-backend/src/channel/interface.js
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * This is a thin layer only responsilbe for receive messages and send messages to testkit.
```
robsdedude:
```suggestion
 * This is a thin layer only responsible for receiving and sending messages from and to testkit.
```

packages/testkit-backend/src/channel/interface.js
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @event contextOpen This event is triggered when a new testkit client start its work.
```
robsdedude:
```suggestion
 * @event contextOpen This event is triggered when a new testkit client starts its work.
```

packages/testkit-backend/src/channel/interface.js
```diff
@@ -0,0 +1,30 @@
...
+ * @event contextClose This event is triggered when an existing client finish it work
```
robsdedude:
```suggestion
 * @event contextClose This event is triggered when an existing client finishes it work
```

packages/testkit-backend/src/channel/interface.js
```diff
@@ -0,0 +1,30 @@
...
+ * @event request Thiis event is triggered when the channel receives a request
```
robsdedude:
```suggestion
 * @event request This event is triggered when the channel receives a request
```

packages/testkit-backend/src/channel/websocket.js
```diff
@@ -0,0 +1,62 @@
...
+  stop () {
+    if(this._ws) {
+      this._ws.close()
+      this._ws = null
+    }
+  }
+
+  writeResponse (contextId, response) {
+    if (this._ws) {
+      console.debug('[WebSocketChannel] Wring response', { contextId, response })
```
robsdedude:
```suggestion
      console.debug('[WebSocketChannel] Writing response', { contextId, response })
```

packages/testkit-backend/src/controller/index.js
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * {@link LocalController} delegates the requests to be handled by local handlers.
+ * {@link RemoteController} delegates the request to be handled by remote clients by forwarding the requests over websockets.
```
robsdedude:
```suggestion
 * {@link RemoteController} delegates the requests to be handled by remote clients by forwarding the requests over websockets.
```

packages/testkit-backend/src/controller/interface.js
```diff
@@ -0,0 +1,30 @@
+import { EventEmitter } from 'events'
+
+/**
+ * Controller is the unit responsabile for redirecting the requests to the correct handler and managing the
```
robsdedude:
```suggestion
 * Controller is the unit responsible for redirecting the requests to the correct handler and managing the
```

packages/testkit-backend/src/controller/local.js
```diff
@@ -0,0 +1,70 @@
+import Context from '../context'
+import Controller from './interface'
+
+
+/**
+ * Local controller handles locally the requests by redirecting to the correct request handler/service.
```
robsdedude:
```suggestion
 * Local controller handles the requests locally by redirecting them to the correct request handler/service.
```

packages/testkit-backend/src/controller/local.js
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * This controller when testing browser and locally.
```
robsdedude:
```suggestion
 * This controller is used when testing browser and locally.
```

packages/testkit-backend/src/controller/remote.js
```diff
@@ -0,0 +1,138 @@
...
+import { createServer } from "http"
+import { Server } from "node-static"
+
+/**
+ * Remote controller handles the requests by send it to a remote client.
+ *
+ * This controller only support only on client connected at the same time and it will
+ * give an error response whenever any request came and it doesn't have any client connected.
+ *
+ * This controller could only be used in Node since it depends on {@link createServer}, {@link WebSocketServer} and {@link Server}
```
robsdedude:
```suggestion
 * RemoteController handles the requests by sending them a remote client.
 *
 * This controller requires a client to be connected. Otherwise, it will reply with a BackendError
 * to every incoming request.
 *
 * This controller can only be used in Node since it depends on {@link createServer}, {@link WebSocketServer} and {@link Server}
```

packages/testkit-backend/src/index.js
```diff
@@ -0,0 +1,46 @@
+import Backend from './backend'
+import { SocketChannel, WebSocketChannel } from './channel'
+import { LocalController, RemoteController } from './controller'
+import * as REQUEST_HANDLERS from './request-handlers'
+
+/**
+ * Responsabile for configure and run the backend server.
```
robsdedude:
```suggestion
 * Responsible for configure and run the backend server.
```

testkit/backend.py
```diff
@@ -3,8 +3,25 @@
 Assumes driver and backend has been built.
 Responsible for starting the test backend.
 """
-from common import run_in_driver_repo
+from common import open_proccess_in_driver_repo, is_browser
```
robsdedude:
```suggestion
from common import (
    open_proccess_in_driver_repo,
    is_browser,
)
```

makes git diffs cleaner ;)

testkit/backend.py
```diff
@@ -3,8 +3,25 @@
...
+        os.environ["TEST_ENVIRONMENT"] = "REMOTE"
+        print("openning firefox")
+
+    print("npm run start-testkit-backend")
+    with open_proccess_in_driver_repo([
+            "npm", "run", "start-testkit-backend"
+    ], env=os.environ) as backend:
+        if (is_browser()):
+            time.sleep(5)
+            with open_proccess_in_driver_repo([
```
robsdedude:
```suggestion

    print("npm run start-testkit-backend")
    with open_proccess_in_driver_repo([
        "npm", "run", "start-testkit-backend"
    ], env=os.environ) as backend:
        if (is_browser()):
            time.sleep(5)
            print("openning firefox")
            with open_proccess_in_driver_repo([
```

Main change is: print "openning firefox" before it actually happens.

testkit/integration.py
```diff
@@ -1,13 +1,6 @@
...
 import os
-from common import run_in_driver_repo, is_lite
-
-
-def should_test_browser():
-    return os.environ.get("TEST_DRIVER_SKIP_BROWSER", "false").lower() not in (
-        "y", "yes", "t", "true", "1", "on"
-    )
-
+from common import run_in_driver_repo, is_lite, is_browser
```
robsdedude:
```suggestion
from common import (
    is_browser,
    is_lite,
    run_in_driver_repo,
)
```

testkit/stress.py
```diff
@@ -1,15 +1,16 @@
 import os
-from common import run_in_driver_repo, is_lite
+from common import run_in_driver_repo, is_lite, is_browser
```
robsdedude:
```suggestion
from common import (
    is_browser,
    is_lite,
    run_in_driver_repo,
)
```

robsdedude (approved):
Quite a few comments. None of them are a deal-breaker. Mainly typos and question.

Great PR!!!

:rocket:

packages/testkit-backend/src/channel/testkit-protocol.js
bigmontz: It's an issue the original code already has it. It's related to the summary information or similar parts, I will keep this way for this PR but myself of the feature will have to solve it properly soon.

packages/testkit-backend/src/channel/socket.js
bigmontz: The random id is to not have any global state in responsible for generate id. It's a bit of preciosity from my side.

About the support for more than one Testkit (client) being connect, It was already something we supported in the original code. 

For each client connect, the backend started it own protocol parser and request handler context. However, now this behaviour is really clear because it's not hidden in a lot of code doing different things. 

I'd like to keep this support for multiple clients at same time, It makes things well segregated and enables the backend run tests in parallel.

packages/testkit-backend/package.json
bigmontz: The `public/index.js` is generated using `rollup` during the `npm run build`. `Rollup` is not need anymore when we run the server, since everything is already built. 

It's pretty much what's happening with the driver, but the driver uses `browserify`, `babel` and `gulp` for building the `lib` and the bundled version which could be imported in browsers. This libraries are used as building tools in the development process, but they are not need in runtime.

PR #802, opened 2021-10-28
Connection errors should be thrown with the root cause

bigmontz:
Suppressing the original failure could make the retry logic doesn't work as expected and the wrong report could masquerade the issue making it difficult to react on errors.


robsdedude (approved):
:bomb:

PR #803, opened 2021-11-01
Handle AuthorisationExpired during rediscovery

bigmontz:
The connections to a server should not be sent back to the pool if an AuthenticationExpired failure happens.

Sending the connection back to the pool will make the next work using the connection fails with the same error.

Testing: https://github.com/neo4j-drivers/testkit/pull/288

### DIFF
--- a/packages/neo4j-driver/test/internal/shared-neo4j.js
+++ b/packages/neo4j-driver/test/internal/shared-neo4j.js
@@ -308,16 +308,23 @@ function restart (configOverride) {
   startNeo4j()
 }
 
-async function cleanupAndGetProtocolVersion (driver) {
+async function cleanupAndGetProtocolVersionAndBookmark (driver) {
   const session = driver.session({ defaultAccessMode: neo4j.session.WRITE })
   try {
     const result = await session.writeTransaction(tx =>
       tx.run('MATCH (n) DETACH DELETE n')
     )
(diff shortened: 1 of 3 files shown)


packages/neo4j-driver/test/internal/shared-neo4j.js
```diff
@@ -314,12 +314,28 @@ async function cleanupAndGetProtocolVersion (driver) {
     const result = await session.writeTransaction(tx =>
       tx.run('MATCH (n) DETACH DELETE n')
     )
+    let dbIsEmpty = false
+    while (!dbIsEmpty) {
+      dbIsEmpty = await checkIfDbIsEmpty(driver)
+    }
```
robsdedude: What this does is to make sure that the clean-up propagated to at least one reader. I'm not sure this is what you want. I assume you want all readers to have received the update.  
But should you indeed want only that, I'd recommend sending a single query but with the bookmark from the `DELETE` attached so that the reader waits before running the query and returning the results until it has caught to the bookmark.

bigmontz: Fair enough. I was waiting for the changes be propagated. In the best scenario, I should go and do it for each reader. But It could be an overkill. 

I've changed the implementation for checking for the bookmark.

robsdedude (approved):
:bookmark:

PR #805, opened 2021-11-03
Fix minimum supported TLS feature

bigmontz:
Testkit only supports 1.1 or later


robsdedude (approved):
:closed_lock_with_key:

PR #809, opened 2021-11-08
Improve performance for packstream in browser.

### LINKED ISSUE #764: Replace text-encoding-utf-8 with a new library that has license
Can we change the dependencies text-encoding-utf-8 for neo4j-driver-bolt-connection to a new npm package? text-encoding-utf-8 license is null. In my org, we cannot use an npm library that uses null icense

bigmontz:
The buffer and utf-8 encoding implementation in browser was using the heap and `text-encoding-utf-8` library respectively. This process was really slower than in nodejs. For example, a string with 85000 length was taking 180 seconds to serialize in the browser, the same string takes milliseconds to be serialised in nodejs.

This performance issue was resolved by using the libraries `buffer` and `string_decoder`, libraries which is already available in nodejs, also in browser. This substitution bring the serialisation of the 85000 length string to take only 350 ms when it's done in browser.

The `text-encoding-utf-8` was removed from the project dependency in the process.


packages/bolt-connection/test/channel/encode-decode.test.js
```diff
@@ -0,0 +1,48 @@
...
+    '±!@#$%^&*()_+~`\'|][{}=-+±§<>,."',
+    'àáâäæãåā',
+    'èéêëēėę',
+    'îïíīīįì',
+    'ôöòóœøōõ',
+    'ûüùúū'
+  ])('should decode encoded string', str => {
+    const encoded = encode(str)
+    
+    const encodedThenDecoded = decode(encoded, encoded.length)
```
robsdedude, line 40: This only checks that encode and decode are inverse functions. It does not assert anything else. I thinks it's worth checking the content of `encoded` at least for some examples.

robsdedude (approved):
:zap: :running_woman:

packages/bolt-connection/test/channel/encode-decode.test.js
bigmontz: I will create some snapshot tests for cover this case.

bigmontz: Done, I've created snapshot tests to verify the encode case. 

I've checked the values in the snapshot using https://mothereff.in/utf-8

robsdedude (approved):
:tada:

PR #810, opened 2021-11-08
testkit: Add Bolt versions to feature list

bigmontz:


robsdedude (approved):
Why did you move it down? It might be my (not real) OCD but... What about lexicographically ordering the entries...

bigmontz:
> Why did you move it down? It might be my (not real) OCD but... What about lexicographically ordering the entries...

I will order at some point in the future, this list should not be there in the long run.

PR #813, opened 2021-11-10
teskit-backend: Enabling tests for Result.then

bigmontz:
These changes enable the test relate to consume all records as list in one function call.

The Result implementation in javascript doesn't allow resolve and subscribe at the same result without having strange behaviours. Then, the moment `ResultObserver` subscribes to the `Result` events was moved to the moment someone interacts with some `ResultObserver` method. This also implies in text which call next before list being skipped.


bigmontz:
Depends on: https://github.com/neo4j-drivers/testkit/pull/303

injectives (approved):

PR #819, opened 2021-12-02
testkit-backend: Enabling skip different tests for browser and node

bigmontz:


packages/testkit-backend/src/skipped-tests/index.js
```diff
@@ -0,0 +1,19 @@
...
+import browserSkippedTests from './browser'
+
+const skippedTestsByContext = new Map([
+  ['browser', browserSkippedTests]
+])
+
+export function getShouldRunTest (contexts) {
+  const skippedTests = contexts
+    .filter(context => skippedTestsByContext.has(context))
+    .map(context => skippedTestsByContext.get(context))
```
fbiville, line 11: nit: you could merge filter and map into a flatMap with `!skippedTestsByContext.has(context) ? [] : [skippedTestsByContext.get(context)]` (less readable though)

fbiville (approved):

packages/testkit-backend/src/skipped-tests/index.js
bigmontz: For sure, It was my first implementation. But I preferred to be a bit more explicit in this case and go with the two steps.

PR #823, opened 2021-12-17
Fix connections not being destroyed when released to purged pool

bigmontz:
This scenario happens when the pull for a given address is purged while a connection stills in use and then another pool for the same address is created. In this case, the connection was being wrongly added to the existing pool.

The correct behaviour is destroy this orphan connections.

### DIFF
--- a/packages/bolt-connection/test/pool/pool.test.js
+++ b/packages/bolt-connection/test/pool/pool.test.js
@@ -237,7 +237,7 @@ describe('#unit Pool', () => {
     expect(r0.destroyed).toBeTruthy()
   })
 
-  it('destroys resource when key was purged event when key is acquired again', async () => {
+  it('destroys resource when pool is purged even if a new pool is created for the same address', async () => {
     let counter = 0
     const address = ServerAddress.fromUrl('bolt://localhost:7687')
     const pool = new Pool({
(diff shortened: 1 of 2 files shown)


packages/bolt-connection/test/pool/pool.test.js
```diff
@@ -237,6 +237,45 @@ describe('#unit Pool', async () => {
     expect(r0.destroyed).toBeTruthy()
   })
 
+  it('destroys resource when key was purged event when key is acquired again', async () => {
```
robsdedude: I don't understand what "event" means in this context.

packages/bolt-connection/test/pool/pool.test.js
```diff
@@ -237,6 +237,45 @@ describe('#unit Pool', async () => {
...
+    expect(pool.has(address)).toBeTruthy()
+    expect(r0.id).toEqual(0)
+
+    // Purging the key
+    await pool.purge(address)
+    expect(pool.has(address)).toBeFalsy()
+    expect(pool.activeResourceCount(address)).toBe(0)
+    expect(r0.destroyed).toBeFalsy()
+
+    // Acquiring second resolve should recreate the pool
```
robsdedude:
```suggestion
    // Acquiring second resource should recreate the pool
```

robsdedude (approved):
:star2:

packages/bolt-connection/test/pool/pool.test.js
bigmontz: Nothing, It was my dyslexia talking.

PR #824, opened 2021-12-17
Fix connections not being destroyed when released to purged pool (#823)

bigmontz:
This scenario happens when the pull for a given address is purged while a connection stills in use and then another pool for the same address is created. In this case, the connection was being wrongly added to the existing pool.

The correct behaviour is destroy this orphan connections.


bigmontz:
Cherry-picks: https://github.com/neo4j/neo4j-javascript-driver/pull/823

robsdedude (approved):
:cherries: :pick:

PR #827, opened 2021-12-23
Use isolatedModules for easier compatibility with Deno (and others)

bradenmacdonald:
This PR changes the driver's TypeScript code to use TypeScript's stricter [`isolatedModules` flag](https://www.typescriptlang.org/tsconfig#isolatedModules).

`isolatedModules` is required to support some transpilers that operate on a single file at a time, so they can tell which imports/exports contain only type information and are safe to discard. This change is more "strict" and does not change the API, so is completely backwards compatible.

Rationale: The reason for this PR is to allow a common codebase to work for both Node.js and Deno. While TypeScript/Node don't care if isolatedModules is enabled or not, Deno requires it. So this is a prerequisite to supporting Deno. See [the relevant forum discussion](https://community.neo4j.com/t/driver-for-deno-land/40567/5). I will soon be opening a PR that shows how to generate a Deno driver, which depends on this change.

Without this change, Deno will report this TypeScript error when trying to transpile the driver code:

```
Re-exporting a type when the '--isolatedModules' flag is provided requires using 'export type'.
```


bigmontz (approved):
🥇 

Thanks for your contribution. 

Just waiting the pipeline go green to

PR #828, opened 2021-12-23
Fix some broken imports

bradenmacdonald:
While I was working on [a Deno driver for Neo4j that uses this codebase](https://deno.land/x/neo4j_lite_client@4.4.1-preview), I found that Deno was detecting some invalid imports in the code.

In particular, a few places in the code try to import a constant called `ALL` from [`bolt/request-message.js`](https://github.com/neo4j/neo4j-javascript-driver/blob/4.4/packages/bolt-connection/src/bolt/request-message.js), but that constant [was renamed a year ago](https://github.com/neo4j/neo4j-javascript-driver/pull/659) to `FETCH_ALL`.

I'm not sure why the existing tooling didn't catch this import error or why Node/TypeScript didn't complain about it.

For the other changes, I'll add an inline comment in this PR to explain them.

### DIFF
--- a/packages/bolt-connection/src/bolt/bolt-protocol-v1.js
+++ b/packages/bolt-connection/src/bolt/bolt-protocol-v1.js
@@ -21,7 +21,7 @@ import {
   assertTxConfigIsEmpty,
   assertImpersonatedUserIsEmpty
 } from './bolt-protocol-util'
-import { Chunker } from '../channel/chunking'
+import { Chunker } from '../channel'
 import { v1 } from '../packstream'
 import RequestMessage from './request-message'
 import {
--- a/packages/bolt-connection/src/bolt/bolt-protocol-v1.js
+++ b/packages/bolt-connection/src/bolt/bolt-protocol-v1.js
@@ -21,7 +21,7 @@ import {
   assertTxConfigIsEmpty,
   assertImpersonatedUserIsEmpty
 } from './bolt-protocol-util'
-import { Chunker } from '../chunking'
+import { Chunker } from '../channel/chunking'
 import { v1 } from '../packstream'
 import RequestMessage from './request-message'
 import {
--- a/packages/bolt-connection/src/bolt/bolt-protocol-v4x0.js
+++ b/packages/bolt-connection/src/bolt/bolt-protocol-v4x0.js
@@ -17,7 +17,7 @@
  * limitations under the License.
  */
 import BoltProtocolV3 from './bolt-protocol-v3'
-import RequestMessage, { ALL } from './request-message'
+import RequestMessage from './request-message'
 import { assertImpersonatedUserIsEmpty } from './bolt-protocol-util'
 import {
   ResultStreamObserver,
@@ -28,7 +28,7 @@ import { internal } from 'neo4j-driver-core'
 
 const {
   bookmark: { Bookmark },
-  constants: { BOLT_PROTOCOL_V4_0 },
+  constants: { BOLT_PROTOCOL_V4_0, FETCH_ALL },
   txConfig: { TxConfig }
 } = internal
 
@@ -90,7 +90,7 @@ export default class BoltProtocol extends BoltProtocolV3 {
       afterComplete,
       flush = true,
       reactive = false,
-      fetchSize = ALL
+      fetchSize = FETCH_ALL
     } = {}
   ) {
     const observer = new ResultStreamObserver({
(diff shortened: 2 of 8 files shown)


packages/bolt-connection/src/rediscovery/rediscovery.js
```diff
@@ -18,7 +18,7 @@
  */
 import RoutingTable from './routing-table'
 import { RawRoutingTable } from '../bolt'
-import { newError, error, Session, ServerAddress } from 'neo4j-driver-core'
+import { newError, error, Session } from 'neo4j-driver-core'
```
bradenmacdonald, line 21: This was giving:
`SyntaxError: The requested module '../../core/index.ts' does not provide an export named 'ServerAddress'`

It seems that it's only used in a JSDoc comment here, and many other files reference it in a JSDoc without importing it so I assume this is fine.

packages/bolt-connection/src/connection/connection.js
```diff
@@ -17,7 +17,7 @@
  * limitations under the License.
  */
 
-import { ResultStreamObserver, BoltProtocol } from '..bolt'
+import { ResultStreamObserver, BoltProtocol } from '../bolt'
```
bradenmacdonald, line 20: Deno doesn't like this `..folder` format. (Was this intentional, or just an oversight? Just wanted to flag in case it's a convention or deliberate usage of some feature I don't know about.)

packages/core/src/internal/observers.ts
```diff
@@ -17,7 +17,6 @@
  * limitations under the License.
  */
 
-import { observer } from '.'
```
bradenmacdonald, line 20: I'm not sure if this import had a purpose, since it doesn't seem to be used in this file. It seems like a circular import issue, since `./index.ts` has `import * as observer from './observers'`

packages/bolt-connection/src/bolt/bolt-protocol-v1.js
```diff
@@ -21,7 +21,7 @@ import {
   assertTxConfigIsEmpty,
   assertImpersonatedUserIsEmpty
 } from './bolt-protocol-util'
-import { Chunker } from '../chunking'
+import { Chunker } from '../channel/chunking'
```
bradenmacdonald: No such file `../chunking` - I think this is the right import.

bigmontz: This was leftover of big refactoring, thanks for spotting.

Actually, it should be import from `../channel` directly, because we'd like to keep the modules only depending on the things exported by other module indexes. This way it's possible to reshape the internal of the modules without breaking the other ones.

packages/bolt-connection/src/bolt/bolt-protocol-v4x0.js
```diff
@@ -90,7 +90,7 @@ export default class BoltProtocol extends BoltProtocolV3 {
       afterComplete,
       flush = true,
       reactive = false,
-      fetchSize = ALL
+      fetchSize = FETCH_ALL
```
bigmontz, line 93: Thanks for catching it, I will make sure I will have tests covering this default value.

packages/bolt-connection/src/connection/connection.js
bigmontz: It was probably some typo or auto-complete issue. It was not deliberated.

packages/bolt-connection/src/rediscovery/rediscovery.js
bigmontz: It's fine, the api documentation of the internals is not exported.

packages/core/src/internal/observers.ts
bigmontz: It's not used at all, probably some issue with the auto-complete. Thanks for catching.

bigmontz (approved):
⭐ 

The PR looks good, just a minor issue with `Chunker` importing. 

Thanks for your contribution.

packages/bolt-connection/src/bolt/bolt-protocol-v4x0.js
bigmontz: (No action needed here)

packages/bolt-connection/src/bolt/bolt-protocol-v1.js
bradenmacdonald: Makes sense. I've changed it to `../channel` now.

PR #829, opened 2021-12-23
Fix some TypeScript errors found with TypeScript 4.5 / Deno

bradenmacdonald:
While I was working on [a Deno driver for Neo4j that uses this codebase](https://deno.land/x/neo4j_lite_client@4.4.1-preview), I ran into some Typescript errors which this PR fixes.

These could be only relevant with the newer version of TypeScript (4.5) or with Deno's stricter `tsconfig` defaults, but I tried to fix them in a way that will work with Node and other versions/configurations of TypeScript without issues.

I'll put explanations of each change inline.

### DIFF
--- a/packages/core/src/connection.ts
+++ b/packages/core/src/connection.ts
@@ -22,8 +22,8 @@
  * @private
  */
 class Connection {
-  id: string
-  databaseId: string
+  id: string = ""
+  databaseId: string = ""
   server: any
 
   /**
--- a/packages/core/src/internal/connection-holder.ts
+++ b/packages/core/src/internal/connection-holder.ts
@@ -194,7 +194,7 @@ class ConnectionHolder implements ConnectionHolderInterface {
    */
   private _releaseConnection(): Promise<Connection | void> {
     this._connectionPromise = this._connectionPromise
-      .then((connection?: Connection) => {
+      .then((connection?: Connection|void) => {
         if (connection) {
           if (connection.isOpen()) {
             return connection
--- a/packages/core/src/internal/observers.ts
+++ b/packages/core/src/internal/observers.ts
@@ -174,6 +174,6 @@ export class FailedObserver implements ResultStreamObserver {
 
 function apply<T>(thisArg: any, func?: (param: T) => void, param?: T): void {
   if (func) {
-    func.bind(thisArg)(param)
+    func.bind(thisArg)(param as any)
   }
 }
--- a/packages/core/src/internal/transaction-executor.ts
+++ b/packages/core/src/internal/transaction-executor.ts
@@ -30,13 +30,14 @@ type TransactionCreator = () => Transaction
 type TransactionWork<T> = (tx: Transaction) => T | Promise<T>
 type Resolve<T> = (value: T | PromiseLike<T>) => void
 type Reject = (value: any) => void
+type Timeout = ReturnType<typeof setTimeout>
 
 export class TransactionExecutor {
   private _maxRetryTimeMs: number
   private _initialRetryDelayMs: number
   private _multiplier: number
   private _jitterFactor: number
-  private _inFlightTimeoutIds: NodeJS.Timeout[]
+  private _inFlightTimeoutIds: Timeout[]
 
   constructor(
     maxRetryTimeMs?: number | null,
--- a/packages/neo4j-driver-lite/src/index.ts
+++ b/packages/neo4j-driver-lite/src/index.ts
@@ -17,6 +17,7 @@
  * limitations under the License.
  */
 import VERSION from './version'
+import { logging } from './logging'
 
 import {
   Neo4jError,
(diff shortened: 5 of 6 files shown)


packages/core/src/connection.ts
```diff
@@ -22,8 +22,8 @@
  * @private
  */
 class Connection {
-  id: string
-  databaseId: string
+  id: string = ""
+  databaseId: string = ""
```
bradenmacdonald, line 26: This fixes:
`"Property 'id' has no initializer and is not definitely assigned in the constructor."`

packages/core/src/internal/connection-holder.ts
```diff
@@ -194,7 +194,7 @@ class ConnectionHolder implements ConnectionHolderInterface {
    */
   private _releaseConnection(): Promise<Connection | void> {
     this._connectionPromise = this._connectionPromise
-      .then((connection?: Connection) => {
+      .then((connection?: Connection|void) => {
```
bradenmacdonald, line 197: This fixes:

```
error: Uncaught (in promise) TypeError: TS2345 [ERROR]: Argument of type '(connection?: Connection | undefined) => Promise<void>' is not assignable to parameter of type '(value: void | Connection) => void | PromiseLike<void>'.
  Types of parameters 'connection' and 'value' are incompatible.
    Type 'void | Connection' is not assignable to type 'Connection | undefined'.
      Type 'void' is not assignable to type 'Connection | undefined'.
      .then((connection?: Connection) => {
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

packages/core/src/internal/observers.ts
```diff
@@ -174,6 +174,6 @@ export class FailedObserver implements ResultStreamObserver {
 
 function apply<T>(thisArg: any, func?: (param: T) => void, param?: T): void {
   if (func) {
-    func.bind(thisArg)(param)
+    func.bind(thisArg)(param as any)
```
bradenmacdonald, line 177: This fixes:

```
error: Uncaught (in promise) TypeError: TS2345 [ERROR]: Argument of type 'T | undefined' is not assignable to parameter of type 'T'.
  'T' could be instantiated with an arbitrary type which could be unrelated to 'T | undefined'.
    func.bind(thisArg)(param)
```

packages/core/src/internal/transaction-executor.ts
```diff
@@ -30,13 +30,14 @@ type TransactionCreator = () => Transaction
...
 type Reject = (value: any) => void
+type Timeout = ReturnType<typeof setTimeout>
 
 export class TransactionExecutor {
   private _maxRetryTimeMs: number
   private _initialRetryDelayMs: number
   private _multiplier: number
   private _jitterFactor: number
-  private _inFlightTimeoutIds: NodeJS.Timeout[]
+  private _inFlightTimeoutIds: Timeout[]
```
bradenmacdonald, line 40: The type `NodeJS.Timeout` is (obviously) specific to Node, and will cause an error in Deno.

Using `ReturnType<typeof setTimeout>` gives the same effect in Node but is also compatible with Deno (or any other runtime).

packages/neo4j-driver-lite/src/index.ts
```diff
@@ -337,7 +337,7 @@ const logging = {
     return {
       level: level,
       logger: (level: LogLevel, message: string) =>
-        console.log(`${global.Date.now()} ${level.toUpperCase()} ${message}`)
+        console.log(`${globalThis.Date.now()} ${level.toUpperCase()} ${message}`)
```
bradenmacdonald: `global` isn't defined in Deno and will cause an error.

[`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) works in Node (v12+) and Deno.

bigmontz: We'd like to keep compatibility with Node v10, since it stills the one installed by `apt get` in Ubuntu LTS.  If you move this object declaration to it own file, it won't have any name conflict and then we will could use `Date.now()` without need any access to the global object.

bigmontz (changes requested):
Thanks for your contribution.

The PR looks good, but the compatibility break with Node 10 should be resolved before the PR get merged. Moving the `logging` to a file call `logging.ts` should solve the naming conflict.

packages/neo4j-driver-lite/src/index.ts
bradenmacdonald: Sure thing. I've pushed a commit to move it to `logging.ts` as you suggested.

bradenmacdonald:
Hmm, I see there is now a CI error but I'm not sure what it is.

bigmontz (approved):
🎹 

Thanks for your contribution.

PR #831, opened 2021-12-28
AsyncIterator API for Result consumption

### LINKED ISSUE #542: [Feature Request] Pull through streaming API to prevent back-pressure
I have written a Neo4j client library that uses this driver under the hood. (https://drivine.org/). 

One of the use-cases is streaming without **back-pressure**. Back-pressure is where the source produces data faster than the sync can process, for example writing too quickly into a file-stream, say if we were producing a report that presents COVID-19 cases for a region. 

The library includes an API as follows: 

```openCursor<T>(spec: QuerySpecification<T>): Promise<Cursor<T>>;```

. . whereby you can open a `Cursor<T>` representing a set of results. This cursor is `AsyncIterable` (supports `for await (const item of cursor)` and more importantly can turn itself into a Node.JS `Readable` stream. 

When these streams are piped together, it is the sink stream that will pull through from the source, at the rate the sink can handle. So no back-pressure problems, excessive memory usage or going over a high-water mark. Good for many use-case. We can easily create an Observable from a stream too. 

*How it is currently Implemented:* 

The way this is implemented is to fetch data from Neo4j in batches using SKIP and LIMIT. 

I wanted to see if I can replace this using the driver's scheming capabilities, however from what I understand RxJS has no capability to handling back-pressure. It is a push-through library. Right? 

For pull-through we should use the companion _IxJS_ (https://github.com/ReactiveX/IxJS) instead. Pulling through will push the onus of handling back-pressure onto the source, which would need to hold the entire result set and emit items when requested. This should not be a problem, as it needs to hold the results in memory for a period of time in any case. 

So how about supporting pull-through streaming at the driver level? Either as `Readable` streams or with `IxJS`. (Or apparently there is a way to do it in RxJS: https://itnext.io/lossless-backpressure-in-rxjs-b6de30a1b6d4 <--- I'm still digesting this article).

### PR DESCRIPTION
**Pull request by bigmontz**
The current driver implementation has only two options for record consumption, the default async/promise api and the reactive api. 

In the async side, there is no simple way to interact over the results before consuming all the data stream and back-pressure the consumption. Meanwhile in the reactive, the back-pressure is implemented  but this api usage depends on the knowledge of reactive programming and `RxJS`.  Another issue with our current reactive implementation is not having the way to configure the back-pressure or pause the stream consumption. 

The introduction of the Async Iterator is part of a group of improvements to make the driver more idiomatic and memory efficient.

### CODE REVIEW

File: packages/neo4j-driver/test/examples.test.js [hunk 1]
```diff
@@ -1511,6 +1511,24 @@ describe('#integration examples', () => {
...
+      try {
+        const result = session.run('UNWIND RANGE(0, 10) AS x RETURN x')
+        const xs = []
+        for await (const record of result) {
+          xs.push(record.get('x').toInt())
+        }
+        expect(xs).toEqual([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
+      } finally {
+        await session.close()
+      }
```
bigmontz, line 1529: Usage example

PR #838, opened 2022-01-10
Fix `neo4j-driver-lite` test suite in newer node versions

bigmontz:
UncatchedException errors were being triggered when non-consumed promises are rejected.


bigmontz:
Cherry picks: https://github.com/neo4j/neo4j-javascript-driver/pull/836

PR #840, opened 2022-01-10
Enables set different Node versions in Testkit Dockerfile

bigmontz:
This change enables testing the driver using different NodeJS versions.

### DIFF
--- a/testkit/Dockerfile
+++ b/testkit/Dockerfile
@@ -1,20 +1,31 @@
 FROM ubuntu:20.04
 
+ARG NODE_VERSION=10
+
 ENV DEBIAN_FRONTEND noninteractive
-ENV NODE_OPTIONS --max_old_space_size=4096
+ENV NODE_OPTIONS --max_old_space_size=4096 --use-openssl-ca
+
+# Configuring NodeJS version
+RUN apt-get clean 
 
 RUN apt-get update && \
+    apt-get upgrade -y && \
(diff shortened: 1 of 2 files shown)


testkit/Dockerfile
```diff
@@ -41,4 +39,12 @@ USER driver
 WORKDIR /home/driver
 CMD /bin/bash
 RUN mkdir /home/driver/.npm_global
+
+# Configuring NodeJS version
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash
```
fbiville: you could parameterize NVM version as well, or?

testkit/Dockerfile
```diff
@@ -41,4 +39,12 @@ USER driver
...
+RUN nvm install ${NODE_VERSION:=10}
```
fbiville: probably for another PR, but Node 10 has reached EOL as you know

fbiville (approved):

testkit/Dockerfile
bigmontz: It's end-of-life, but it is the one which you could install with apt in debian/ubuntu lts.

fbiville: (not important for this PR but:) for Go, I manually install a recent version, I am not installing the default, EOL one.

fbiville (approved):

PR #841, opened 2022-01-12
Allow tx timeout to be 0 and send it

robsdedude:
Altering TestKit back end to treat exceptions on session.run and tx.run
as `DriverError`s

### DIFF
--- a/packages/bolt-connection/src/bolt/request-message.js
+++ b/packages/bolt-connection/src/bolt/request-message.js
@@ -286,7 +286,7 @@ function buildTxMetadata (bookmark, txConfig, database, mode, impersonatedUser) 
   if (!bookmark.isEmpty()) {
     metadata.bookmarks = bookmark.values()
   }
-  if (txConfig.timeout != null) {
+  if (txConfig.timeout !== null) {
     metadata.tx_timeout = txConfig.timeout
   }
   if (txConfig.metadata) {
--- a/packages/bolt-connection/src/bolt/request-message.js
+++ b/packages/bolt-connection/src/bolt/request-message.js
@@ -256,7 +256,7 @@ export default class RequestMessage {
     if ( databaseContext.databaseName ) {
       dbContext.db = databaseContext.databaseName
     }
-    
+
     if ( databaseContext.impersonatedUser ) {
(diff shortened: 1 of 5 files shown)


packages/testkit-backend/src/request-handlers.js
```diff
@@ -153,7 +153,14 @@ export function SessionRun (context, data, wire) {
...
     .then(_ => {
-      const result = session.run(cypher, params, { metadata, timeout })
+      let result
+      try {
+        result = session.run(cypher, params, { metadata, timeout })
+      } catch (e) {
+        console.log('got some err: ' + JSON.stringify(e))
+        wire.writeError(e)
+        return
+      }
```
bigmontz, line 163: Is it a leftover? The backend should be able to capture this kind of errors in the upstream. If it is not, there is a bug in the backend. 

Same for the other block like this.

packages/bolt-connection/src/bolt/request-message.js
```diff
@@ -286,7 +286,7 @@ function buildTxMetadata (bookmark, txConfig, database, mode, impersonatedUser)
   if (!bookmark.isEmpty()) {
     metadata.bookmarks = bookmark.values()
   }
-  if (txConfig.timeout) {
+  if (txConfig.timeout != null) {
```
bigmontz: It's a good practice use '===' and '!==' instead of '==' and '!=' for comparisons in Javascript. The regular comparison methods try to  do type conversion during the comparison and it could lead to strange behaviours.

bigmontz (changes requested):
Looks good apart of the usage of `!=` instead of `!==`.

packages/testkit-backend/src/request-handlers.js
bigmontz: Actually It is correct, my bad 🛏️

packages/bolt-connection/src/bolt/request-message.js
bigmontz: Since we'd like the same behaviour for undefined, issue solved.

bigmontz (approved):
⏲️ 💥

PR #842, opened 2022-01-14
Add method for convert driver temporal types to Javascript Date

bigmontz:
The conversion between date types in the javascript driver can be a bit convoluted and have to do conversions like `new Date(driverDate.toString())` doesn't help.

This change introduces the method `toStandardDate` to `Date`, `DateTime` and `LocalDateTime` which could avoid the repeating conversion boilerplate in the developer code.


AndyHeap-NeoTech (approved):
LGTM

PR #844, opened 2022-01-18
Make newly created session return no bookmarks

robsdedude:
Depends on neo4j-drivers/testkit#364

### DIFF
--- a/packages/core/src/session.ts
+++ b/packages/core/src/session.ts
@@ -55,7 +55,8 @@ class Session {
   private _writeConnectionHolder: ConnectionHolder
   private _open: boolean
   private _hasTx: boolean
-  private _lastBookmark: Bookmark
+  private _bookmarkIn: Bookmark
+  private _bookmarkOut?: string
   private _transactionExecutor: TransactionExecutor
   private _impersonatedUser?: string
   private _onComplete: (meta: any) => void
@@ -117,7 +118,7 @@ class Session {
     this._open = true
     this._hasTx = false
     this._impersonatedUser = impersonatedUser
-    this._lastBookmark = bookmark || Bookmark.empty()
+    this._bookmarkIn = bookmark || Bookmark.empty()
     this._transactionExecutor = _createTransactionExecutor(config)
     this._onComplete = this._onCompleteCallback.bind(this)
     this._databaseNameResolved = this._database !== ''
@@ -150,7 +151,7 @@ class Session {
     return this._run(validatedQuery, params, connection => {
       this._assertSessionIsOpen()
       return (connection as Connection).protocol().run(validatedQuery, params, {
-        bookmark: this._lastBookmark,
+        bookmark: this._bookmarkIn,
         txConfig: autoCommitTxConfig,
         mode: this._mode,
         database: this._database,
@@ -271,7 +272,7 @@ class Session {
       reactive: this._reactive,
       fetchSize: this._fetchSize
     })
-    tx._begin(this._lastBookmark, txConfig)
+    tx._begin(this._bookmarkIn, txConfig)
     return tx
   }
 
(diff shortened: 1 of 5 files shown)


packages/core/src/transaction.ts
```diff
@@ -226,8 +226,8 @@ class Transaction {
    * @param {object} meta The meta with bookmark
    * @returns {void}
    */
-  _onCompleteCallback(meta: { bookmark?: string | string[] }): void {
-    this._onBookmark(new Bookmark(meta.bookmark))
+  _onCompleteCallback(meta: { bookmark?: string }): void {
+    this._onBookmark(meta.bookmark)
```
bigmontz, line 230: Why change `onBookmark` signature?

packages/core/src/session.ts
```diff
@@ -376,12 +377,13 @@ class Session {
...
+   * @param {string | undefined} newBookmark - The new bookmark.
    * @returns {void}
    */
-  _updateBookmark(newBookmark?: Bookmark): void {
-    if (newBookmark && !newBookmark.isEmpty()) {
-      this._lastBookmark = newBookmark
+  _updateBookmark(newBookmark?: string): void {
+    if (newBookmark !== undefined) {
+      this._bookmarkIn = new Bookmark(newBookmark)
+      this._bookmarkOut = newBookmark
```
bigmontz, line 386: Actually, the previous `newBookmark` will be always defined. So, it could be changed for `if(!newBookmark.isEmpty())`

packages/core/src/session.ts
```diff
@@ -55,7 +55,8 @@ class Session {
   private _writeConnectionHolder: ConnectionHolder
   private _open: boolean
   private _hasTx: boolean
-  private _lastBookmark: Bookmark
+  private _bookmarkIn: Bookmark
+  private _bookmarkOut?: string
```
bigmontz, line 59: Maybe it could be called `currentBookmark` and `lastReceivedBookmark`, or something like this. Because the `_bookmarkIn` is being changed when a new bookmark arrives.

packages/core/src/session.ts
```diff
@@ -296,10 +297,10 @@ class Session {
...
    * Return the bookmark received following the last completed {@link Transaction}.
    *
-   * @return {string[]} A reference to a previous transaction.
+   * @return {string | undefined} A reference to a previous transaction if present.
    */
-  lastBookmark(): string[] {
-    return this._lastBookmark.values()
+  lastBookmark(): string | undefined {
+    return this._bookmarkOut
   }
```
bigmontz, line 304: Why did the interface changed?

bigmontz (commented):
I've made some comments for understand some changes in the code and specially in the interface. 

Apart of the other question, why does the received bookmark could not be considered last seem until we receive a new one?

packages/core/src/transaction.ts
robsdedude: Because I want to fail fast should the server send multiple bookmarks. According to the protocol spec, there should only ever be 0 or 1 bookmark in a `SUCCESS` message.

robsdedude: Also, thinking further down the call stack, I want `_bookmarkOut` to be a string. How do you turn a `Bookmark` (which can contain a string array) into a single string in a type safe way?

packages/core/src/session.ts
robsdedude: But what was the `!newBookmark.isEmpty()` there fore before if not to check if the `SUCCESS` message that triggered this update actually contains a bookmark?

packages/core/src/session.ts
robsdedude: I'm fine with that :ok_hand: Will do.

packages/core/src/session.ts
robsdedude: I saw a recent question from support regarding the Java driver: they wanted to know if they can safely assume that `.lastBookmark` would only ever return one bookmark. With this signature, there is no doubt about it.

robsdedude:
Superseded by https://github.com/neo4j/neo4j-javascript-driver/pull/848

PR #845, opened 2022-01-20
testkit: Fix `Unknown request` error path

bigmontz:
There was two different errors durint the unknown request error treatment.

First, the stringfy function was not define in the LocalController scope. It get solve by extract this function to it own file and import it in the local.js.

Second, the backend was capturing the error in the `request` event and calling `writeBackendError` with the error object instead of the message.


robsdedude (approved):
:thread:

PR #846, opened 2022-01-25
Introduce `.peek()` to the Async Iterator API

bigmontz:
This feature allows the user peek the element on the head of the cursor without moving the cursor forward.

The implementation of this feature has as side effect a new implementation `.return` method in the iterator.
This new version of the method does most the same thing as the previous one, but it also cancels the stream
since this method moves the cursor to the end.

The `next` method also got tweaked.

### DIFF
--- a/packages/core/test/result.test.ts
+++ b/packages/core/test/result.test.ts
@@ -19,7 +19,6 @@
 import { observer, connectionHolder } from '../src/internal'
 import {
   Connection,
-  internal,
   newError,
   Record,
   ResultObserver,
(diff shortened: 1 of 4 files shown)


bigmontz:
Depends on: https://github.com/neo4j-drivers/testkit/pull/380

packages/core/test/result.test.ts
```diff
@@ -757,6 +797,256 @@ describe('Result', () => {
...
+          const rawRecord2 = [3, 4]
+
+          streamObserverMock.onKeys(keys)
+          streamObserverMock.onNext(rawRecord1)
+          streamObserverMock.onNext(rawRecord2)
+
+          streamObserverMock.onCompleted({})
+
+          const it = result[Symbol.asyncIterator]()
+          const { value: record } = await it.peek()
```
bigmontz, line 974: Usage example

robsdedude (approved):
:rocket:

PR #847, opened 2022-01-26
Introduce Transaction.close method

bigmontz:
This method closes the transaction and rolled it back if it stills open.

### DIFF
--- a/packages/core/src/transaction.ts
+++ b/packages/core/src/transaction.ts
@@ -220,6 +220,19 @@ class Transaction {
     return this._state === _states.ACTIVE
   }
 
+  /**
+   * Closes the transaction
+   *
+   * This method will rollback the transaction if it is not committed or rolled back.
+   *
+   * @returns {Promise<void>} An empty promise if closed successfully or error if any error happened during
+   */
+  async close(): Promise<void> {
+    if(this.isOpen()) {
+      await this.rollback()
+    }
+  }
+
   _onErrorCallback(err: any): Promise<Connection | void> {
     // error will be "acknowledged" by sending a RESET message
     // database will then forget about this transaction and cleanup all corresponding resources
--- a/packages/core/src/transaction.ts
+++ b/packages/core/src/transaction.ts
@@ -223,12 +223,12 @@ class Transaction {
   /**
    * Closes the transaction
    *
-   * This method will rollback the transaction if it is not committed or rolled back.
+   * This method will roll back the transaction if it is not already committed or rolled back.
    *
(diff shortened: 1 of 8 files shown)


packages/core/src/transaction.ts
```diff
@@ -220,6 +220,19 @@ class Transaction {
     return this._state === _states.ACTIVE
   }
 
+  /**
+   * Closes the transaction
+   *
+   * This method will rollback the transaction if it is not committed or rolled back.
```
fbiville:
```suggestion
   * This method will roll back the transaction if it is not already committed or rolled back.
```

packages/core/src/transaction.ts
```diff
@@ -220,6 +220,19 @@ class Transaction {
...
+   *
+   * @returns {Promise<void>} An empty promise if closed successfully or error if any error happened during
+   */
+  async close(): Promise<void> {
+    if(this.isOpen()) {
```
fbiville:
```suggestion
    if (this.isOpen()) {
```

packages/core/test/transaction.test.ts
```diff
@@ -60,6 +60,59 @@ describe('Transaction', () => {
 
   })
 
+  describe('.close()', () => {
+    describe('when transaction is open', () => {
+      it('should rollback the transaction', async () => {
```
fbiville:
```suggestion
      it('should roll back the transaction', async () => {
```

packages/core/test/transaction.test.ts
```diff
@@ -60,6 +60,59 @@ describe('Transaction', () => {
...
+      const error = async (tx: Transaction, conn: FakeConnection) => {
+        conn.withRollbackError(new Error('rollback error'))
+        return tx.rollback().catch(() => { })
+      }
+
+      it.each([
+        ['commmited', commit],
+        ['rolled back', rollback],
+        ['with error', error]
+      ])('should not rollback the connection', async (_, operation) => {
```
fbiville:
```suggestion
      ])('should not roll back the connection', async (_, operation) => {
```

packages/neo4j-driver/src/transaction-rx.js
```diff
@@ -90,4 +90,22 @@ export default class RxTransaction {
         .catch(err => observer.error(err))
     })
   }
+
+  /**
+   * Closes the transaction
+   *
+   * This method will rollback the transaction if it is not committed or rolled back.
```
fbiville:
```suggestion
   * This method will roll back the transaction if it is not already committed or rolled back.
```

fbiville (approved):

PR #848, opened 2022-01-26
Rename `Session.lastBookmark` to `Session.lastBookmarks`

robsdedude:
Made a clear internal distinction between what single bookmark is (a string that
is exchanged with the server) and what Bookmarks are (an array of such strings).


bigmontz (approved):
🔖 🔖

PR #849, opened 2022-01-26
Session.close should DISCARD and ROLLBACK active TXs

bigmontz:


bigmontz:
Sends `RESET` during the `Session.close` is the correct behaviour since it unlocks queries and tx blocked in the server.

PR #853, opened 2022-01-31
Add back `(rx)session.lastBookmark` methods

bigmontz:
These method were added back deprecated for keeping the driver backward compatible.

### DIFF
--- a/packages/core/src/session.ts
+++ b/packages/core/src/session.ts
@@ -307,6 +307,17 @@ class Session {
   /**
    * Return the bookmarks received following the last completed {@link Transaction}.
    *
+   * @deprecated This method will be removed in 6.0 version. Please, use {@link Session#lastBookmarks} instead.
+   * 
+   * @return {string[]} A reference to a previous transaction.
+   */
+   lastBookmark(): string[] {
+    return this.lastBookmarks()
+  }
+
+  /**
+   * Return the bookmarks received following the last completed {@link Transaction}.
+   *
    * @return {string[]} A reference to a previous transaction.
    */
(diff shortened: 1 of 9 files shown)


packages/core/src/session.ts
```diff
@@ -304,6 +304,17 @@ class Session {
     this._hasTx = false
   }
 
+  /**
+   * Return the bookmarks received following the last completed {@link Transaction}.
+   *
+   * @deprecated This method will be removed in 6.0 version. Please, use {@link Session#lastBookmarks} instead.
```
robsdedude:
```suggestion
   * @deprecated This method will be removed in version 6.0. Please, use {@link Session#lastBookmarks} instead.
```

packages/neo4j-driver/src/driver.js
```diff
@@ -67,7 +67,7 @@ class Driver extends CoreDriver {
     return new RxSession({
       session: this._newSession({
         defaultAccessMode,
-        bookmarks,
+        bookmarkOrBookmarks: bookmarks,
```
robsdedude, line 70: Interesting decision to rename `bookmarks` internally (admittedly `bookmarkOrBookmarks` is the more precise name). :shrug:
Definitely out of scope of this PR.

robsdedude (approved):

packages/neo4j-driver/src/driver.js
bigmontz: Yes, and it's actually a bug fix I am porting to 4.4

https://github.com/neo4j/neo4j-javascript-driver/pull/854

PR #854, opened 2022-01-31
Fix `driver.rxSession` initial bookmarks configuration and `RxSession.lastBookmark` return mapping

bigmontz:
The initial bookmark configuration for the RxSession were not working since the wrong parameter were
being passed to the Session. The typescript mapping for `RxSession.lastBookmark` was also wrong, `lastBookmark`
always returns an string array.

### DIFF
--- a/packages/neo4j-driver/test/driver.test.js
+++ b/packages/neo4j-driver/test/driver.test.js
@@ -26,7 +26,9 @@ import {
 } from '../../bolt-connection/lib/pool/pool-config'
 import { ServerVersion, VERSION_4_0_0 } from '../src/internal/server-version'
 import testUtils from './internal/test-utils'
-import { json } from 'neo4j-driver-core'
+import { json, internal } from 'neo4j-driver-core'
+
+const { bookmark } = internal
 
 // As long as driver creation doesn't touch the network it's fine to run
 // this as a unit test.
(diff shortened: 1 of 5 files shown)


packages/neo4j-driver/test/driver.test.js
```diff
@@ -26,7 +26,9 @@ import {
 } from '../../bolt-connection/lib/pool/pool-config'
 import { ServerVersion, VERSION_4_0_0 } from '../src/internal/server-version'
 import testUtils from './internal/test-utils'
-import { json } from 'neo4j-driver-core'
+import { json, internal } from 'neo4j-driver-core'
+
+const { bookmark } = internal
```
fbiville, line 31: can you write `import { json, { bookmark } } from 'neo4j-driver-core'`? :P

packages/neo4j-driver/test/types/driver.test.ts
```diff
@@ -135,6 +135,15 @@ const rxSession9: RxSession = driver.rxSession({
   bookmarks: ['bookmark1', 'bookmark2']
 })
 
+const rxSession10: RxSession = driver.rxSession({
```
fbiville, line 138: where are those used?

fbiville (approved):

packages/neo4j-driver/test/types/driver.test.ts
bigmontz: The type tests are a bit strange, we just call a method with the expected params and receiving the expect results. Then we check if it get perfect checked by the TS compiler.

fbiville: or is it just to exercise the typescript compiler?

packages/neo4j-driver/test/driver.test.js
bigmontz: Unfortunately, no. 

I've put everything from core which is not supposed to be externalised by the driver in the `internal` package for making easier to realize it should not be externalised. 

It will be good with. could do `import { internal: { bookmark } } from 'neo4j-driver-core`, but unfortunately It can't be done. 😢

packages/neo4j-driver/test/types/driver.test.ts
bigmontz: Yes, to check if it get typed correctly

PR #864, opened 2022-02-04
Improve back-pressure control in `Result[Symbol.asyncIterator]`

bigmontz:
The previous implemenation only controls the flow while is iterating over the records, but if the interation
is done in a really low pace, the control flow method won't be called in time for pausing the flow.

Changing the flow control for observing the queue size instead of check on each time it iterates solve this issue.


robsdedude (approved):
:ocean:

PR #869, opened 2022-02-08
Validate query type in the ResultSummary

bigmontz:
Unexpected query types are protocol violations and it should be treaten like it.


thelonelyvulpes (approved):
Looks good.

PR #870, opened 2022-02-08
Introduce the ability of awaiting for eager beginning a Transaction

bigmontz:
The current implementation of `session.beginTransaction` does not wait the success message. This kind of behaviour can cause future calls to `run` and `commit` because of non-reported errors during the begin. 

For solving this issue, the transaction object return by the `session.beginTransaction` was changed to `TransactionPromise`. This new type holds all the behaviours from a Transaction and also holds the promise for the beginned transaction. This way, the caller can eagerly eager check the transaction begin and at the same type keep the API existing in the current codebase.

**Eager Mode:**
```javascript
const tx = await session.beginTransaction() // waits for begin result
const result = await tx.run('RETURN 1') // only throws RUN related errors, not reachable when begin fails
```

**Lazy Mode:**
```javascript
const tx = session.beginTransaction() // no wait, any begin errors will be thrown
const result = await tx.run('RETURN 1') // throws BEGIN and RUN related errors
```


bigmontz:
Depends on: https://github.com/neo4j-drivers/testkit/pull/397

packages/core/src/transaction-promise.ts
```diff
@@ -0,0 +1,193 @@
...
+  private _beginPromise?: Promise<Transaction>;
+  private _reject?: (error: Error) => void;
+  private _resolve?: (value?: Transaction | PromiseLike<Transaction> | undefined) => void;
+
+  /**
+   * @constructor
+   * @param {ConnectionHolder} connectionHolder - the connection holder to get connection from.
+   * @param {function()} onClose - Function to be called when transaction is committed or rolled back.
+   * @param {function(bookmarks: Bookmarks)} onBookmarks callback invoked when new bookmark is produced.
+   * * @param {function()} onConnection - Function to be called when a connection is obtained to ensure the conneciton
```
fbiville:
```suggestion
   * @param {function()} onConnection - Function to be called when a connection is obtained to ensure the connection
```

fbiville: the docs also miss the watermarks

packages/core/src/transaction-promise.ts
```diff
@@ -0,0 +1,193 @@
...
+ * Represents a {@link Promise<Transaction>} object and a {@link Transaction} object.
+ *
+ * Resolving this object promise verifies the result of the transaction begin and returns the {@link Transaction} object in case of success.
+ *
+ * The object can still also used as {@link Transaction} for convenience. The result of begin will be checked 
+ * during the next API calls in the object as it is in the transaction.
+ *
+ * @access public
+ */
+class TransactionPromise extends Transaction implements Promise<Transaction>{
```
fbiville, line 38: being pedantic here (feel free to ignore): why use inheritance instead of composition here?

packages/core/src/transaction-promise.ts
```diff
@@ -0,0 +1,193 @@
...
+      commit: super.commit.bind(this),
+      rollback: super.rollback.bind(this),
+      close: super.close.bind(this),
+      isOpen: super.isOpen.bind(this),
+      _begin: this._begin.bind(this),
+    }
+  }
+
+  /**
+   * @access private
```
fbiville, line 162: wrong documented access or missing keyword?

packages/core/src/transaction-promise.ts
```diff
@@ -0,0 +1,193 @@
...
+  finally(onfinally?: (() => void) | null): Promise<Transaction> {
+    return this._getOrCreateBeginPromise().finally(onfinally);
+  }
+
+  private _getOrCreateBeginPromise(): Promise<Transaction> {
+    if (!this._beginPromise) {
+      this._beginPromise = new Promise((resolve, reject) => {
+        this._resolve = resolve;
+        this._reject = reject;
+        if (this._beginError) {
```
fbiville, line 134: how can this condition be truthy when the promise has not been initialized yet? I have trouble following the flow here

packages/core/src/transaction-promise.ts
bigmontz: It is private for outside work, we still using it inside the `Session.beginTransaction`.

packages/core/src/transaction-promise.ts
bigmontz: I will add docs about the bookmarks, this docs came from the original Transaction class.

packages/core/src/transaction-promise.ts
bigmontz: That's fun. As shown bellow, the begin method already subscribe to the begin observer and keep track of the errors and any promises are created. So, if the user took sufficient time before await in the transaction, the `beginError` or `beginMetadata` could be already created. So the promise return by `_getOrCreateBeginPromise` will be already solved.

```
_begin(bookmarks: string | Bookmarks | string[], txConfig: TxConfig): void {
    return super._begin(bookmarks, txConfig, {
      onError: this._onBeginError.bind(this),
      onComplete: this._onBeginMetadata.bind(this)
    });
  }
```

Why do I not create the promise as soon as I create the TxPromise object? Simple, if the promise get reject without anyone catching it,  It will have `UncaughtPromiseException` in newer versions of Nodejs. So, I have to postpone the internal promise creation until the moment the user tries to resolve `TransactionPromise` and keep the begin events in memory for when i need it.

packages/core/src/transaction-promise.ts
bigmontz: I've tried to thing in way of doing it using composition. But then I will need to have a `Transaction` interface for documenting the Transaction behaviour and use the TransactionPromise as a kind of mixin. The problem is the docs, the documentation tool doest not work pretty well with typescript interfaces, I will have a second thinking round about it, just to see if I could came with a solution for this problem.

bigmontz: Using composition will also avoid the method `toTransaction`, this method is pure madness.

packages/core/src/transaction-promise.ts
fbiville: > if the promise get reject without anyone catching it, It will have UncaughtPromiseException in newer versions of Nodejs.

isn't that a misuse of the promise if the user does not catch anything? Shouldn't you let the default behavior happen (ie `UncaughtPromiseException` if rejections ain't handled)?

bigmontz: Yes, it is a misuse. But the TransactionPromise should behave as the Transaction, so the user could choose for not trying to resolve it without any issue. 

But if the user call `then` or `finally` without have a catch in the chain, they will still have the `UncaughtPromiseException` as expected.

fbiville (approved):

PR #882, opened 2022-02-22
Improve back-pressure mechanism for RxResult

### LINKED ISSUE #542: [Feature Request] Pull through streaming API to prevent back-pressure

bigmontz:
The lack of consumer oriented back-pressure mechanism in the RxJS apis was causing memory issues and other problems in the client code. 

For solving this issue, the records observer returned by `RxResult.records()` was changed for using the async iterator as foundation. New methods for enabling the client for fine controlling the stream were added to the `RxResult` api. These methods are:

- `pause()`: Pause the record streaming. No new record will be pushed to the stream util `push` or `resume` get called. 
- `resume()`: Resumes the records streaming. 
- `push()`: Push the next record. If the stream is not paused, this method will pause it for giving the push control to the client.

### DIFF
--- a/packages/neo4j-driver/test/examples.test.js
+++ b/packages/neo4j-driver/test/examples.test.js
@@ -17,10 +17,16 @@
  * limitations under the License.
  */
 
-import neo4j from '../src'
+import neo4j, { session } from '../src'
 import sharedNeo4j from './internal/shared-neo4j'
 import { ServerVersion, VERSION_4_0_0 } from '../src/internal/server-version'
-import { map, materialize, toArray } from 'rxjs/operators'
+import {
+  bufferCount,
+  map,
+  materialize,
+  mergeMap,
+  toArray
+} from 'rxjs/operators'
 import { Notification } from 'rxjs'
 
 /**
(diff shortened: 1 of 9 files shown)


packages/neo4j-driver/test/examples.test.js
```diff
@@ -1529,6 +1535,66 @@ describe('#integration examples', () => {
...
+            // it pauses the automatic pushing
+            await result.push()
+          },
+          complete: async () => {
+            expect(xs).toEqual([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
+            await session.close().toPromise()
+            done()
+          },
+          error: done.fail.bind(done)
+        })
```
bigmontz, line 1563: Usage for result.push api

packages/neo4j-driver/test/examples.test.js
```diff
@@ -1529,6 +1535,66 @@ describe('#integration examples', () => {
...
+            return theXs
+          }),
+          toArray()
+        )
+        .toPromise()
+
+      expect(xs).toEqual([[0, 1, 2, 3, 4], [5, 6, 7, 8, 9], [10]])
+    } finally {
+      await session.close().toPromise()
+    }
```
bigmontz, line 1596: Examples for pause and resume

PR #884, opened 2022-03-01
Enabling BoltProtocol v5 and the new elementId in the graph types

bigmontz:
The format of the identifiers used by the kernel is changing from the current long that represents a file offset to a more flexible format needed by the new Freki storage engine. In 5.0 the new storage engine will be ‘off’ by default, at some point in the future it will be enabled by default. The drivers will need to be able to support the old format alongside the new one in 5.0. To do this the existing long will remain and a new parallel identifier will be added that is of type string.

The new string version will contain the long value for servers using the older non Freki storage engine e.g. versions < 5.0 and versions 5.x with Freki disabled.

The following changes in the graph types was done for accommodating the elementId:

- in `Node`:
  - Add `elementId: string` for identifying the element;
  - Deprecate `identity: NumberOrInteger`
- in `Relationship`:
  - Add `elementId: string` for identifying the element;
  - Deprecate `identity: NumberOrInteger`
  - Add `startNodeElementId: string` for identifying the start node;
  - Deprecate `start: NumberOrInteger`
  - Add `endNodeElementId: string` for identifying the end node;
  - Deprecate `end: NumberOrInteger`
- in `UnboundRelationship`:
  - Add `elementId: string` for identifying the element;
  - Deprecate `identity: NumberOrInteger`


thelonelyvulpes (approved):
💯

PR #889, opened 2022-03-08
Discard active auto-commit results on Session.close()

bigmontz:
Discarding the elements before reseting the Session allows non-consumed results to be committed before reset the result.

This PR also improves the `keys()` and `summary()` methods by allowing get errors cached in the result.


robsdedude (approved):
:chains:

PR #893, opened 2022-03-08
Un-deprecate Integer (#891)

bigmontz:
The usage of BigInt is not as simple in some scenarios, so Integer stills a useful object to used in some scenarios.


bigmontz:
Cherry picked #891

PR #901, opened 2022-03-16
Add `.retriable` to the `Neo4jError`

bigmontz:
This property should helps users to implement custom retries logics.

Along with the property, the static function `Neo4jError.isRetriable(error)` and
the function `isRetriableError(error)` were added for saving boilerplates and avoid
the error type checking to be spread across the client and driver code base.

The internal function `canRetryOn(error)` and the `retry-strategy` file were removed
since the logic implemented there was moved to the `error` module replaced by
the `isRetriableError(error)` usage.


packages/core/src/error.ts
```diff
@@ -90,8 +109,62 @@ function newError (message: string, code?: Neo4jErrorCode): Neo4jError {
...
+    _isAuthorizationExpired(code) ||
+    _isTransientError(code)
+}
+
+/**
+ * @private
+ * @param {string} code the error to check
+ * @return {boolean} true if the error is a transient error
+ */
+function _isTransientError (code?: Neo4jErrorCode): boolean {
```
injectives: Please note that we are currently looking at whether we might be able to change these 2 errors to a different category.

injectives (approved):

packages/core/src/error.ts
bigmontz: Sure, this function is internal, but I will change the name to something like `_isRetriableTransientError()`.

PR #903, opened 2022-03-23
Fix `receiveTimeout` behaviour

bigmontz:
The previous implementation of the `timeout` was not considering if requests were being treated by the server or not.
This behaviour causes connections being wrongly close by inactivity.

For fixing this issue, the `timeout` configuration changed to only be applied whenever there are ongoing requests.

### DIFF
--- a/packages/bolt-connection/src/channel/node/node-channel.js
+++ b/packages/bolt-connection/src/channel/node/node-channel.js
@@ -242,6 +242,8 @@ export default class NodeChannel {
       this
     )
     this._connectionErrorCode = config.connectionErrorCode
+    this._receiveTimeout = null
+    this._receiveTimeoutStarted = false
 
     this._conn = connect(
       config,
@@ -353,7 +355,27 @@ export default class NodeChannel {
       )
     })
 
-    this._conn.setTimeout(receiveTimeout)
+    this._receiveTimeout = receiveTimeout
+  }
+
+  /**
+   * Stops the receive timeout for the channel.
+   */
+  stopReceiveTimeout() {
+    if (this._receiveTimeout !== null && this._receiveTimeoutStarted) {
+      this._receiveTimeoutStarted = false
+      this._conn.setTimeout(0)
+    }
+  }
+
+  /**
+   * Start the receive timeout for the channel.
+   */
+  startReceiveTimeout () {
+    if (this._receiveTimeout !== null && !this._receiveTimeoutStarted) {
+      this._receiveTimeoutStarted = true
+      this._conn.setTimeout(this._receiveTimeout)
+    }
   }
 
   /**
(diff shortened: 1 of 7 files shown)


packages/bolt-connection/test/channel/node/node-channel.test.js
```diff
@@ -59,13 +59,13 @@ describe('NodeChannel', () => {
   })
 
   describe('.setupReceiveTimeout()', () => {
-    it('should call socket.setTimeout(receiveTimeout)', () => {
+    it('should call not call socket.setTimeout(receiveTimeout)', () => {
```
fbiville: typo?

packages/bolt-connection/src/channel/node/node-channel.js
```diff
@@ -353,7 +355,27 @@ export default class NodeChannel {
...
+    this._receiveTimeout = receiveTimeout
+  }
+
+  /**
+   * Stops the receive timeout for the channel.
+   */
+  stopReceiveTimeout() {
+    if (this._receiveTimeout !== null && this._receiveTimeoutStarted) {
+      this._receiveTimeoutStarted = false
+      this._conn.setTimeout(0)
```
fbiville, line 367: do these `setTimeout` calls bode well with connection pooling/reuse?
could you end with a connection that has received a custom timeout for a previous op and remains with such a timeout for a new op (that would not normally time out)?

packages/bolt-connection/test/channel/node/node-channel.test.js
bigmontz: Yes

packages/bolt-connection/src/channel/node/node-channel.js
bigmontz: Yes, It actuality solves this issue. The first implementation was setting the timeout when the HELLO SUCCESS was received and it just timeouts any time in the feature because of inactivity.

Now I am deferring the timeout set for whenever a request is sent and it is unset just after there aren't ongoing requests.

The implementation is not the most obvious since the `ChannelConnection` doesn't control the send/receive messages, it is done by the protocol (which is weird and I have to change it in the future).

packages/bolt-connection/test/channel/node/node-channel.test.js
```diff
@@ -59,13 +59,13 @@ describe('NodeChannel', () => {
   })
 
   describe('.setupReceiveTimeout()', () => {
-    it('should call socket.setTimeout(receiveTimeout)', () => {
+    it('should not call socket.setTimeout(receiveTimeout)', () => {
```
fbiville, line 62: nit: given the test name, I'd expect to see an assertion about the fact socket.setTimeout being called 0 times in the test, but I don't see any

fbiville (approved):

packages/bolt-connection/test/channel/node/node-channel.test.js
bigmontz: It always called when the connection is established

PR #904, opened 2022-03-24
Fix `receiveTimeout` behaviour (#903)

bigmontz:
The previous implementation of the `timeout` was not considering if requests were being treated by the server or not.
This behaviour causes connections being wrongly close by inactivity.

For fixing this issue, the `timeout` configuration changed to only be applied whenever there are ongoing requests.


bigmontz:
Cherry picks #903

PR #911, opened 2022-04-04
Deprecate `Session.(read|write)Transaction` in favor of `execute(Read|Write)` methods

bigmontz:
The new methods provides a `ManagedTransaction` objects to the transaction functions. These
transaction objects don't have `commit`, `rollback` and `close` capabilities exposed in the API.

### DIFF
--- a/packages/core/src/transaction-managed.ts
+++ b/packages/core/src/transaction-managed.ts
@@ -24,10 +24,6 @@ interface Run {
   (query: Query, parameters?: any): Result
 }
 
-interface IsOpen {
-  (): boolean
-}
-
 /**
  * Represents a transaction that is managed by the transaction executor.
  * 
@@ -35,17 +31,12 @@ interface IsOpen {
  */
 class ManagedTransaction {
   private _run: Run
-  private _isOpen: IsOpen
 
-  constructor({ run, isOpen }: { run: Run, isOpen: IsOpen }) {
+  constructor({ run }: { run: Run }) {
     /**
      * @private
      */
     this._run = run
-    /**
-     * @private
-     */
-    this._isOpen = isOpen
   }
 
   /**
@@ -58,14 +49,6 @@ class ManagedTransaction {
    */
   run(query: Query, parameters?: any): Result {
     return this._run(query, parameters)
-  }
-
-  /**
-  * Check if this transaction is active, which means commit and rollback did not happen.
-  * @return {boolean} `true` when not committed and not rolled back, `false` otherwise.
-  */
-  isOpen(): boolean {
-    return this._isOpen()
   }
 }
 
--- a/packages/core/src/transaction-managed.ts
+++ b/packages/core/src/transaction-managed.ts
@@ -1,0 +1,72 @@
+/**
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Result from './result'
+import { Query } from './types'
+
+interface Run {
+  (query: Query, parameters?: any): Result
+}
+
+interface IsOpen {
+  (): boolean
+}
+
+/**
+ * Represents a transaction that is managed by the transaction executor.
+ * 
+ * @public
+ */
+class ManagedTransaction {
+  private _run: Run
+  private _isOpen: IsOpen
+
+  constructor({ run, isOpen }: { run: Run, isOpen: IsOpen }) {
+    /**
+     * @private
+     */
+    this._run = run
+    /**
+     * @private
+     */
+    this._isOpen = isOpen
+  }
+
+  /**
+   * Run Cypher query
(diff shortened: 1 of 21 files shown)


packages/core/src/transaction-managed.ts
```diff
@@ -0,0 +1,72 @@
...
+   */
+  run(query: Query, parameters?: any): Result {
+    return this._run(query, parameters)
+  }
+
+  /**
+  * Check if this transaction is active, which means commit and rollback did not happen.
+  * @return {boolean} `true` when not committed and not rolled back, `false` otherwise.
+  */
+  isOpen(): boolean {
```
fbiville: why would you need this on a managed transaction?

packages/core/test/session.test.ts
```diff
@@ -279,6 +280,59 @@ describe('session', () => {
...
+        expect(typeof tx).toEqual('object')
+        expect(tx).toBeInstanceOf(ManagedTransaction)
+
+        status.functionCalled = true
+      })
+
+      expect(status.functionCalled).toEqual(true)
+    })
+
+    it('should proxy run to the begined transaction', async () => {
```
fbiville:
```suggestion
    it('should proxy run to the begun transaction', async () => {
```

packages/core/test/session.test.ts
```diff
@@ -279,6 +280,59 @@ describe('session', () => {
...
+      await execute(session)(async (tx: ManagedTransaction) => {
+        status.functionCalled = true
+        await tx.run(query, params)
+      })
+
+      expect(status.functionCalled).toEqual(true)
+      expect(run).toHaveBeenCalledWith(query, params)
+    })
+
+    it('should proxy isOpen to the begined transaction', async () => {
```
fbiville:
```suggestion
    it('should proxy isOpen to the begun transaction', async () => {
```

packages/neo4j-driver/src/session-rx.js
```diff
@@ -104,6 +107,45 @@ export default class RxSession {
...
+
+  /**
+   * Executes the provided unit of work in a {@link READ} reactive transaction which is created with the provided
+   * transaction configuration.
+   * @public
+   * @param {function(txc: RxManagedTransaction): Observable} work - A unit of work to be executed.
+   * @param {TransactionConfig} transactionConfig - Configuration for the enclosing transaction created by the driver.
+   * @returns {Observable} - A reactive stream returned by the unit of work.
+   */
+   executeRead (work, transactionConfig) {
```
fbiville, line 119:
```suggestion
   executeRead(work, transactionConfig) {
```
nit: is the space put on purpose here? (same question below)

packages/neo4j-driver/src/session-rx.js
```diff
@@ -104,6 +107,45 @@ export default class RxSession {
...
+  }
+
+  /**
+   * @private
+   * @param {function(txc: RxManagedTransaction): Observable} work
+   * @param {TransactionConfig} transactionConfig
+   * @returns {Observable}
+   */
+  _executeInTransaction (accessMode, work, transactionConfig) {
+    const wrapper = txc => new RxManagedTransaction({
```
fbiville, line 142: wouldn't it be easier if `RxManagedTransaction` constructor accepted `txc` as its argument instead of an arbitrary object?

packages/neo4j-driver/src/transaction-managed-rx.js
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * Creates a reactive result that will execute the query in this transaction, with the provided parameters.
+   *
+   * @public
+   * @param {string} query - Query to be executed.
+   * @param {Object} parameters - Parameter values to use in query execution.
+   * @returns {RxResult} - A reactive result
+   */
+   run (query, parameters) {
```
fbiville, line 54:
```suggestion
   run(query, parameters) {
```

bigmontz: JS Standard formatting requires a space after the method name. It is ensured by linter/autoformatter.

packages/neo4j-driver/src/session-rx.js
bigmontz: JS Standard formatting requires a space after the method name. It is ensured by linter/autoformatter.

packages/core/src/transaction-managed.ts
bigmontz: True. I will remove it.

packages/neo4j-driver/src/session-rx.js
bigmontz: The bind will only change places at the end. I don't want to keep reference from the transaction inside the ManagedTransaction to avoid the client code by pass the encapsulation and commit/rollback/close it anyway.

fbiville: what about:
```typescript
class RxManagedTransaction {

constructor(txc) {
    this._run = txc.run.bind(txc)
    this._isOpen = tcx.isOpen.bind(txc)
  }
```
so you don't have to repeat the destructuring every time you call the constructor

packages/neo4j-driver/test/internal/transaction-executor.test.js
```diff
@@ -17,7 +17,7 @@
  * limitations under the License.
  */
 
-import { newError, error as err, internal } from 'neo4j-driver-core'
+import { newError, error as err, internal, int } from 'neo4j-driver-core'
```
fbiville: what's `int`?

fbiville (approved):

packages/neo4j-driver/src/session-rx.js
bigmontz: I've moved to the binding to a static factory method called `fromTransaction`. It makes easier to pass the function as parameter.

packages/neo4j-driver/test/internal/transaction-executor.test.js
bigmontz: It is a function which creates the custom Integer object from a `number`, `string`, `bigint` or `{ low: 1, high: 1 }`. But i will remove the import because it is not being used in the file.

PR #920, opened 2022-04-19
Replace legacy TestKit test IDs

robsdedude:


packages/testkit-backend/src/skipped-tests/common.js
```diff
@@ -134,7 +134,7 @@ const skippedTests = [
     ifEquals(
       'stub.iteration.test_result_list.TestResultList.test_tx_func_result_list_pulls_all_records_at_once_next_before_list'
     ),
-    ifEquals(
+    ifEquals(test_disconnect_on_commit
       'stub.iteration.test_result_list.TestResultList.test_session_run_result_list_pulls_all_records_at_once_next_before_list'
```
bigmontz: It didn't seem to be a valid javascript construction and also we won't have any test with full name equal to `test_disconnect_on_commit`

bigmontz (changes requested):
It might have a bug here

packages/testkit-backend/src/skipped-tests/common.js
robsdedude: shame on me

bigmontz (approved):
🐞

PR #937, opened 2022-05-04
Fix lack of ProtocolVersion in the `Result.summary()`

bigmontz:
When there are more than one observer subscribed to the ResultStreamObserver, the first notified observed got the summary with the protocol info because the connection holder stills holding the connection. The next observers notified won't have the protocol version since the connection holder will not have the connection in hands at the moment.

This situation is occurring during some tests where result iterator is not fully consumed and the summary is called in the meantime.

Checking for existance of the summary in the result and notifying the observer with the already created summary should solve issue since the first summary created will have the protocol version set.


robsdedude (approved):
:trophy:

PR #940, opened 2022-06-01
Remove null handling for Node/Relationship Ids

thelonelyvulpes:
Remove code for handling null ids inline with updated specification regarding node/ relationship ids

### DIFF
--- a/packages/bolt-connection/test/packstream/packstream-v5.test.js
+++ b/packages/bolt-connection/test/packstream/packstream-v5.test.js
@@ -238,42 +238,6 @@ describe('#unit PackStreamV5', () => {
       return [nodeStruct, expectedNode, { disableLosslessIntegers: true, useBigInt: false }]
     }
 
-    function validWithoutOldIdentifiersLossy () {
-      const identity = null
-      const labels = ['a', 'b']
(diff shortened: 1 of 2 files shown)


packages/bolt-connection/test/packstream/packstream-v5.test.js
```diff
@@ -238,42 +238,6 @@ describe('#unit PackStreamV5', () => {
...
-      const properties = { a: 1, b: 2 }
-      const elementId = 'element_id_1'
-      const expectedNode = new Node(BigInt(-1), labels, properties, elementId)
-      const nodeStruct = new Structure(0x4e, [
-        identity, labels, properties, elementId
-      ])
-      return [nodeStruct, expectedNode, { disableLosslessIntegers: false, useBigInt: true }]
-    }
-
     function validWithInt () {
```
bigmontz, line 241: Instead of removing the test scenario, update it to reflect the new conditions. The same for the other tests.

bigmontz (changes requested):
Please, update the tests instead of remove it. 

Node, Relationship and UnboundRelationship should be updated to reflect the new contract.

bigmontz (approved):
👍🏼

PR #941, opened 2022-06-07
Fix exported Typescript types

nglgzz:
Use `typeof` to export the correct Typescript types for the graph types exported by the driver.

PR #946, opened 2022-06-09
Fix exported types for result, resultsummary, and record

nglgzz:
Follow up from https://github.com/neo4j/neo4j-javascript-driver/pull/941, I just realized Result, ResultSummary, and Record are also classes and need to use the `typeof` operator to export the correct type.


bigmontz (approved):
🌮

PR #953, opened 2022-06-23
Implementing new structures for DateTime

bigmontz:
The structures with signature `0x46` and `0x66` are being replaced by `0x49` and `0x69`. This new structures changes the meaning of seconds and nano seconds from `adjusted Unix epoch`  to `UTC`.

This changes have with goal of avoiding un-existing or ambiguous ZonedDateTime to be received or sent over Bolt.

Bolt v4.3 and v4.4 were patched to support this feature if the server supports the patch.

This is a back-port of https://github.com/neo4j/neo4j-javascript-driver/pull/948


bigmontz:
Missing `packstream-v2` tests backport

PR #957, opened 2022-07-01
Treat errors on handling income bolt messages

bigmontz:
Invalid income messages which could not be parsed or handled correctly where throwing exceptions in non-safe context. This leads the driver to crash hard since the error is bubble up to the socket thread.


robsdedude (approved):
:star:

PR #958, opened 2022-07-01
Treat errors on handling income bolt messages

bigmontz:
Invalid income messages which could not be parsed or handled correcty where throwing exceptions in non-safe context. This leads the driver to crash hard since the error is bubble up to the socket thread.


bigmontz:
cherry-picks https://github.com/neo4j/neo4j-javascript-driver/pull/957

robsdedude (approved):
:star:

PR #959, opened 2022-07-01
Validate the ZoneId of the DateTime with ZoneId

bigmontz:
The validation of the DateTime was only being done in the new patched protocol while unpacking the struct. This changes force any new DateTime with ZoneID to have a valid ZoneId.

This changes also treats struct unpacking errors and defers the occurred to the moment the object is manipulate. For instance, a DateTime with invalid ZoneId returned in a Record will not break the records consumption until any code try to interacts with the broken DateTime.

### DIFF
--- a/packages/core/src/internal/util.ts
+++ b/packages/core/src/internal/util.ts
@@ -224,31 +224,31 @@ function isString(str: any): str is string {
 }
 
 /**
- * Creates a object which all method call will thrown the given error
+ * Creates a object which all method call will throw the given error
  * 
  * @param {Error} error The error
  * @param {any} object The object. Default: {}
  * @returns {any} A broken object
  */
 function createBrokenObject<T extends object> (error: Error, object: any = {}): T {
-  const thrown = () => {
+  const fail = () => {
     throw error
   }
 
   return new Proxy(object, {
-    get: thrown,
-    set: thrown,
-    apply: thrown,
-    construct: thrown,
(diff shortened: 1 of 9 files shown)


packages/core/src/internal/util.ts
```diff
@@ -223,6 +223,35 @@ function isString(str: any): str is string {
...
+    getOwnPropertyDescriptor: thrown,
+    getPrototypeOf: thrown,
+    has: thrown,
+    isExtensible: thrown,
+    ownKeys: thrown,
+    preventExtensions: thrown,
+    setPrototypeOf: thrown,
+  })
+}
+
```
robsdedude:
```suggestion
 * Creates a object which all method call will throw the given error
 * 
 * @param {Error} error The error
 * @param {any} object The object. Default: {}
 * @returns {any} A broken object
 */
function createBrokenObject<T extends object> (error: Error, object: any = {}): T {
  const fail = () => {
    throw error
  }

  return new Proxy(object, {
    get: fail,
    set: fail,
    apply: fail,
    construct: fail,
    defineProperty: fail,
    deleteProperty: fail,
    getOwnPropertyDescriptor: fail,
    getPrototypeOf: fail,
    has: fail,
    isExtensible: fail,
    ownKeys: fail,
    preventExtensions: fail,
    setPrototypeOf: fail,
  })
}

```

robsdedude (approved):

PR #960, opened 2022-07-04
Add support for error cause in Neo4jError and Routing errors

bigmontz:
Root causes of routing issue were suppressed by the routing error message.

This change enables the `Neo4jError` object use the `cause` property for giving more context about the error. This
also improves the routing error by making use of the error `cause`.

`Error.cause` is a new feature in JS, see which runtime already implemented it:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error

### DIFF
--- a/packages/bolt-connection/src/connection-provider/connection-provider-routing.js
+++ b/packages/bolt-connection/src/connection-provider/connection-provider-routing.js
@@ -121,7 +121,8 @@ export default class RoutingConnectionProvider extends PooledConnectionProvider 
     this.forgetWriter(address, database || DEFAULT_DB_NAME)
     return newError(
       'No longer possible to write to server at ' + address,
-      SESSION_EXPIRED
+      SESSION_EXPIRED,
+      error
     )
   }
 
@@ -338,7 +339,7 @@ export default class RoutingConnectionProvider extends PooledConnectionProvider 
   ) {
     // we start with seed router, no routers were probed before
     const seenRouters = []
-    let newRoutingTable = await this._fetchRoutingTableUsingSeedRouter(
+    let [newRoutingTable, error] = await this._fetchRoutingTableUsingSeedRouter(
       seenRouters,
       this._seedRouter,
       currentRoutingTable,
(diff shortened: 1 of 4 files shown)


packages/bolt-connection/src/connection-provider/connection-provider-routing.js
```diff
@@ -473,19 +478,19 @@ export default class RoutingConnectionProvider extends PooledConnectionProvider
         }
 
         // try next router
-        const session = await this._createSessionForRediscovery(
+        const [session, error] = await this._createSessionForRediscovery(
           currentRouter,
           bookmarks,
           impersonatedUser
         )
         if (session) {
```
robsdedude, line 486: Generally (not only here), the readability might benefit from short-circuiting along the lines of

```typescript
const [foo, error] = await foobar();
if (error) { // alternatively: `if (foo === null) {` or similar
  return [null, error]
}
// ...
```

packages/core/src/error.ts
```diff
@@ -70,8 +70,9 @@ class Neo4jError extends Error {
    * @param {string} message - the error message
    * @param {string} code - Optional error code. Will be populated when error originates in the database.
    */
-  constructor (message: string, code: Neo4jErrorCode) {
-    super(message)
+  constructor (message: string, code: Neo4jErrorCode, cause?: Error) {
+    // @ts-expect-error
+    super(message, cause != null ? { cause } : undefined)
```
robsdedude, line 75: I am surprised this doesn't require a poly-fill. Why is this, do you have a link to docs or similar where I can educate myself?

packages/core/test/error.test.ts
```diff
@@ -43,6 +51,31 @@ describe('newError', () => {
...
+    if (supportsCause) {
+      // @ts-expect-error
+      expect(error.cause).toBe(cause)
+    } else {
+      // @ts-expect-error
+      expect(error.cause).toBeUndefined()
+    }
+  })
+
+  test.each([null, undefined])('should create Neo4jErro without cause (%s)', (cause) => {
```
robsdedude:
```suggestion
  test.each([null, undefined])('should create Neo4jError without cause (%s)', (cause) => {
```

robsdedude (approved):

packages/bolt-connection/src/connection-provider/connection-provider-routing.js
bigmontz: Yes, but I'd like to change that much the code for this fix.

packages/core/src/error.ts
bigmontz: The Error cause was added as an extra parameter to the Error constructor and Javascript doesn't enforce the number of parameters for calling a function. So, if i declare `function fn(a) { ... }` and you call as `fn('a', 'b')` without any issue. So, calling `Error('message', { cause: 'a' })` will not break any old implementation. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error

robsdedude: JS for you :upside_down_face: Thank's for the explanation.

PR #964, opened 2022-07-08
Temporarily enforce DateTime integration tests check for failures in 5.0

bigmontz:
The Neo4j 5.0 server didn't implemented yet the new date time messages and meanwhile the tests should be failing for 5.0.

Changing temporarily the expectation to be failure avoids to have red pipelines and at the same time alert us when the feature gets delivered.


bigmontz:
This PR should be reverted when the data time gets implemented in 5.0 server.

bigmontz (commented):
Fix neo4jServer5xDateTypesImplmentationGuard method

PR #965, opened 2022-07-11
Make Result[Symbol.asyncIterator].return wait for Summary

bigmontz:
The lack of result wait for result summary was causing confusion with API usage. When `await it.return()` is called, the caller expect the result summary to be return by the method call.

### DIFF
--- a/packages/core/test/result.test.ts
+++ b/packages/core/test/result.test.ts
@@ -939,11 +939,11 @@ describe('Result', () => {
           const it = result[Symbol.asyncIterator]()
           await it.next()
           // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-          const sumamryPromise = it.return!(summary)
+          const summaryPromise = it.return!(summary)
 
           streamObserverMock.onCompleted({})
 
-          const { value, done } = await sumamryPromise
+          const { value, done } = await summaryPromise
           expect(value).toBe(summary)
           expect(done).toEqual(true)
         })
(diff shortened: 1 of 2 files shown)


packages/core/test/result.test.ts
```diff
@@ -939,8 +939,33 @@ describe('Result', () => {
           const it = result[Symbol.asyncIterator]()
           await it.next()
           // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-          const { value, done } = await it.return!(summary)
+          const sumamryPromise = it.return!(summary)
```
OskarDamkjaer: Small misspelling here sumamryPromise => summaryPromise

packages/core/test/result.test.ts
```diff
@@ -939,8 +939,33 @@ describe('Result', () => {
...
+          const summary = new ResultSummary('query', {}, {})
+
+          streamObserverMock.onKeys(keys)
+          streamObserverMock.onNext(rawRecord1)
+          streamObserverMock.onNext(rawRecord2)
+
+          const it = result[Symbol.asyncIterator]()
+          await it.next()
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          const sumamryPromise = it.return!()
```
OskarDamkjaer: here as well :+1:

OskarDamkjaer (approved):
Looking good! ⭐

PR #966, opened 2022-07-11
Introduce `sessionConnectionTimeout` and `updateRoutingTableTimeout`

bigmontz:
`sessionConnnectionTimeout` is the maximum amount of time the session will wait when trying to establish a usable read/write connection to the remote host. This encompasses *everything* that needs to happen for this, including, if necessary, updating the routing table<sup>1</sup>, fetching a connection from the pool, and, if necessary performing a BOLT and Authentication handshake with the reader/writer.

`updateRoutingTableTimeout` is the maximum amount of time the driver will attempt to fetch a new routing table. This encompasses *everything* that needs to happen for this, including fetching connections from the pool, performing handshakes, and requesting and receiving a fresh routing table.

<sup>1</sup>Updating the routing table in turn might also require fetching a connection for the pool and performing the handshakes, plus potentially trying other routers.


robsdedude (approved):
:stopwatch: :runner:

PR #974, opened 2022-08-01
Introduce Bookmark Manager

bigmontz:
Neo4j clusters are causally consistent, meaning that by default there is no guarantee a query will be able to read changes made by a previous query. For cases where such a guarantee is necessary, the server provides bookmarks to the client. A bookmark is an abstract token that represents some state of the database. By passing one or multiple bookmarks along with a query, the server will make sure that the query will not get executed before the represented state(s) (or a later state) have been established.

The bookmark manager is an interface used by the driver for keeping track of the bookmarks and this way keeping sessions automatically consistent.

### Usage in the session

Enabling it is done by supplying an BookmarkManager implementation instance to this param.
A default implementation could be acquired by calling the factory function `neo4j.bookmarkManager()`.

**Warning**: Share the same BookmarkManager instance across all session can have a negative impact
on performance since all the queries will wait for the latest changes being propagated across the cluster.
For keeping consistency between a group of queries, use `Session` for grouping them.
For keeping consistency between a group of sessions, use `BookmarkManager` instance for grouping them.

```javascript
const bookmarkManager = neo4j.bookmarkManager()
const linkedSession1 = driver.session({ database:'neo4j', bookmarkManager })
const linkedSession2 = driver.session({ database:'neo4j', bookmarkManager })
const unlinkedSession = driver.session({ database:'neo4j' })

// Creating Driver User
const createUserQueryResult = await linkedSession1.run('CREATE (p:Person {name: $name})', { name: 'Driver User'})

// Reading Driver User will *NOT* wait of the changes being propagated to the server before RUN the query
// So the 'Driver User' person might not exist in the Result
const unlinkedReadResult = await unlinkedSession.run('CREATE (p:Person {name: $name}) RETURN p', { name: 'Driver User'})

// Reading Driver User will wait of the changes being propagated to the server before RUN the query
// So the 'Driver User' person should exist in the Result, unless deleted.
const linkedSession2 = await linkedSession2.run('CREATE (p:Person {name: $name}) RETURN p', { name: 'Driver User'})

await linkedSession1.close()
await linkedSession2.close()
await unlinkedSession.close()
```

### DIFF
--- a/packages/core/src/bookmark-manager.ts
+++ b/packages/core/src/bookmark-manager.ts
@@ -17,47 +17,56 @@
  * limitations under the License.
  */
 
-export default interface BookmarkManager {
-  /* Method called when the bookmarks get updated when a transaction finished.
-  *
-  * This method will be called during when auto-commit queries finished and explicit transactions
-  * get commited.
-  * @param database The database which the bookmarks belongs to
-  * @param previousBookmarks The bookmarks used during the transaction creation
-  * @param newBookmarks The new bookmarks resolved at the end of the transaction.
-  * @returns {void}
+/**
+ * Interface for the piece of software responsible for keeping track of current active bookmarks accross the driver.
+ * @interface
+ */
+export default class BookmarkManager {
+  /**
+   * Method called when the bookmarks get updated when a transaction finished.
+   *
+   * This method will be called during when auto-commit queries finished and explicit transactions
+   * get commited.
+   * @param {string} database The database which the bookmarks belongs to
+   * @param {string[]} previousBookmarks The bookmarks used during the transaction creation
+   * @param {string[]} newBookmarks The new bookmarks resolved at the end of the transaction.
+   * @returns {void}
   */
-  updateBookmarks: (database: string, previousBookmarks: string[], newBookmarks: string[]) => void
+  updateBookmarks (database: string, previousBookmarks: string[], newBookmarks: string[]): void {
+    throw new Error('Not implemented')
+  }
 
   /**
    * Method called by the driver to get the bookmark for one specific database
    *
-   * @param database The database which the bookmarks belongs to
+   * @param {string} database The database which the bookmarks belongs to
    * @returns {string[]} The set of bookmarks
    */
-  getBookmarks: (database: string) => string[]
+  getBookmarks (database: string): string[] {
+    throw new Error('Not implemented')
+  }
 
   /**
    * Method called by the driver for getting all the bookmarks.
    *
    * The return of this method should be all the bookmarks present in the BookmarkManager for all databases.
-   * The databases informed in the method call will be used for enriching the bookmark set by enforcing the bookmark
-   * manager calls `bookmarkSupplier` for these database names even though this database are not present in the bookmark
-   * manager map yet.
    *
-   * @param mustIncludedDatabases The database which must be included in the result even if they don't have be initialized yet.
    * @returns {string[]} The set of bookmarks
    */
-  getAllBookmarks: () => string[]
+  getAllBookmarks (): string[] {
+    throw new Error('Not implemented')
+  }
 
   /**
    * Forget the databases and its bookmarks
    *
    * This method is not called by the driver. Forgetting unused databases is the user's responsibility.
    *
-   * @param databases The databases which the bookmarks will be removed for.
+   * @param {string[]} databases The databases which the bookmarks will be removed for.
    */
-  forget: (databases: string[]) => void
+  forget (databases: string[]): void {
+    throw new Error('Not implemented')
+  }
 }
 
 export interface BookmarkManagerConfig {
@@ -66,6 +75,20 @@ export interface BookmarkManagerConfig {
   bookmarksConsumer?: (database: string, bookmarks: string[]) => void
 }
 
+/**
+ * @typedef {Object} BookmarkManagerConfig
+ * @property {Map<string,string[]>} [initialBookmarks] Defines the initial set of bookmarks. The key is the database name and the values are the bookmarks.
+ * @property {function([database]: string):string[]} [bookmarksSupplier] Called for supplying extra bookmarks to the BookmarkManager
+ * 1. supplying bookmarks from the given database when the default BookmarkManager's `.getBookmarks(database)` gets called.
+ * 2. supplying all the bookmarks when the default BookmarkManager's  `.getAllBookmarks()` gets called
+ * @property {function(database: string, bookmarks: string[]): void} [bookmarksConsumer] Called when the set of bookmarks for database get updated
+ */
+/**
+ * Provides an configured {@link BookmarkManager} instance.
+ *
+ * @param {BookmarkManagerConfig} [config={}]
+ * @returns {BookmarkManager}
+ */
 export function bookmarkManager (config: BookmarkManagerConfig = {}): BookmarkManager {
   const initialBookmarks = new Map<string, Set<string>>()
 
--- a/packages/core/src/bookmark-manager.ts
+++ b/packages/core/src/bookmark-manager.ts
@@ -32,8 +32,8 @@ export default class BookmarkManager {
    * This method will be called during when auto-commit queries finished and explicit transactions
    * get commited.
    * @param {string} database The database which the bookmarks belongs to
-   * @param {string[]} previousBookmarks The bookmarks used during the transaction creation
-   * @param {string[]} newBookmarks The new bookmarks resolved at the end of the transaction.
+   * @param {string[]} previousBookmarks The bookmarks used when starting the transaction
+   * @param {string[]} newBookmarks The new bookmarks received at the end of the transaction.
    * @returns {void}
   */
   updateBookmarks (database: string, previousBookmarks: string[], newBookmarks: string[]): void {
@@ -41,9 +41,9 @@ export default class BookmarkManager {
   }
 
   /**
-   * Method called by the driver to get the bookmark for one specific database
+   * Method called by the driver to get the bookmarks for one specific database
    *
-   * @param {string} database The database which the bookmarks belongs to
+   * @param {string} database The database which the bookmarks belong to
    * @returns {string[]} The set of bookmarks
    */
   getBookmarks (database: string): string[] {
@@ -53,7 +53,7 @@ export default class BookmarkManager {
   /**
    * Method called by the driver for getting all the bookmarks.
    *
-   * The return of this method should be all the bookmarks present in the BookmarkManager for all databases.
+   * This method should return all bookmarks for all databases present in the BookmarkManager.
    *
    * @returns {string[]} The set of bookmarks
    */
--- a/packages/core/src/bookmark-manager.ts
+++ b/packages/core/src/bookmark-manager.ts
@@ -84,7 +84,7 @@ export interface BookmarkManagerConfig {
  * @property {Map<string,string[]>} [initialBookmarks] Defines the initial set of bookmarks. The key is the database name and the values are the bookmarks.
  * @property {function([database]: string):string[]} [bookmarksSupplier] Called for supplying extra bookmarks to the BookmarkManager
  * 1. supplying bookmarks from the given database when the default BookmarkManager's `.getBookmarks(database)` gets called.
- * 2. supplying all the bookmarks when the default BookmarkManager's  `.getAllBookmarks()` gets called
+ * 2. supplying all the bookmarks when the default BookmarkManager's `.getAllBookmarks()` gets called
  * @property {function(database: string, bookmarks: string[]): void} [bookmarksConsumer] Called when the set of bookmarks for database get updated
  */
 /**
--- a/packages/core/src/bookmark-manager.ts
+++ b/packages/core/src/bookmark-manager.ts
@@ -18,12 +18,13 @@
  */
 
 export default interface BookmarkManager {
-  /**
-  * Method called when the bookmarks get update because of a given event
+  /* Method called when the bookmarks get updated when a transaction finished.
   *
+  * This method will be called during when auto-commit queries finished and explicit transactions
+  * get commited.
   * @param database The database which the bookmarks belongs to
-  * @param previousBookmarks The bookmarks used during the session creation
-  * @param newBookmarks The new bookmarks resolved at the end of the session.
+  * @param previousBookmarks The bookmarks used during the transaction creation
+  * @param newBookmarks The new bookmarks resolved at the end of the transaction.
   * @returns {void}
   */
   updateBookmarks: (database: string, previousBookmarks: string[], newBookmarks: string[]) => void
@@ -37,7 +38,12 @@ export default interface BookmarkManager {
   getBookmarks: (database: string) => string[]
 
   /**
-   * Method called by the driver for getting all the bookmarks
+   * Method called by the driver for getting all the bookmarks.
+   *
+   * The return of this method should be all the bookmarks present in the BookmarkManager for all databases.
+   * The databases informed in the method call will be used for enriching the bookmark set by enforcing the bookmark
+   * manager calls `bookmarkSupplier` for these database names even though this database are not present in the bookmark
+   * manager map yet.
    *
    * @param mustIncludedDatabases The database which must be included in the result even if they don't have be initialized yet.
    * @returns {string[]} The set of bookmarks
--- a/packages/core/src/bookmark-manager.ts
+++ b/packages/core/src/bookmark-manager.ts
@@ -36,7 +36,7 @@ export default class BookmarkManager {
    * @param {Iterable<string>} newBookmarks The new bookmarks received at the end of the transaction.
    * @returns {void}
   */
-  updateBookmarks (database: string, previousBookmarks: Iterable<string>, newBookmarks: Iterable<string>): void {
+  async updateBookmarks (database: string, previousBookmarks: Iterable<string>, newBookmarks: Iterable<string>): Promise<void> {
     throw new Error('Not implemented')
   }
 
@@ -46,7 +46,7 @@ export default class BookmarkManager {
    * @param {string} database The database which the bookmarks belong to
    * @returns {Iterable<string>} The set of bookmarks
    */
-  getBookmarks (database: string): Iterable<string> {
+  async getBookmarks (database: string): Promise<Iterable<string>> {
     throw new Error('Not implemented')
   }
 
@@ -57,7 +57,7 @@ export default class BookmarkManager {
    *
    * @returns {Iterable<string>} The set of bookmarks
    */
-  getAllBookmarks (): Iterable<string> {
+  async getAllBookmarks (): Promise<Iterable<string>> {
     throw new Error('Not implemented')
   }
 
@@ -68,24 +68,24 @@ export default class BookmarkManager {
    *
    * @param {Iterable<string>} databases The databases which the bookmarks will be removed for.
    */
-  forget (databases: Iterable<string>): void {
+  async forget (databases: Iterable<string>): Promise<void> {
     throw new Error('Not implemented')
   }
 }
 
 export interface BookmarkManagerConfig {
   initialBookmarks?: Map<string, Iterable<string>>
-  bookmarksSupplier?: (database?: string) => Iterable<string>
-  bookmarksConsumer?: (database: string, bookmarks: Iterable<string>) => void
+  bookmarksSupplier?: (database?: string) => Promise<Iterable<string>>
+  bookmarksConsumer?: (database: string, bookmarks: Iterable<string>) => Promise<void>
 }
 
 /**
  * @typedef {Object} BookmarkManagerConfig
  * @property {Map<string,Iterable<string>>} [initialBookmarks] Defines the initial set of bookmarks. The key is the database name and the values are the bookmarks.
- * @property {function([database]: string):Iterable<string>} [bookmarksSupplier] Called for supplying extra bookmarks to the BookmarkManager
+ * @property {function([database]: string):Promise<Iterable<string>>} [bookmarksSupplier] Called for supplying extra bookmarks to the BookmarkManager
  * 1. supplying bookmarks from the given database when the default BookmarkManager's `.getBookmarks(database)` gets called.
  * 2. supplying all the bookmarks when the default BookmarkManager's `.getAllBookmarks()` gets called
- * @property {function(database: string, bookmarks: Iterable<string>): void} [bookmarksConsumer] Called when the set of bookmarks for database get updated
+ * @property {function(database: string, bookmarks: Iterable<string>): Promise<void>} [bookmarksConsumer] Called when the set of bookmarks for database get updated
  */
 /**
  * Provides an configured {@link BookmarkManager} instance.
@@ -108,13 +108,13 @@ export function bookmarkManager (config: BookmarkManagerConfig = {}): BookmarkMa
 class Neo4jBookmarkManager implements BookmarkManager {
   constructor (
     private readonly _bookmarksPerDb: Map<string, Set<string>>,
-    private readonly _bookmarksSupplier?: (database?: string) => Iterable<string>,
-    private readonly _bookmarksConsumer?: (database: string, bookmark: Iterable<string>) => void
+    private readonly _bookmarksSupplier?: (database?: string) => Promise<Iterable<string>>,
+    private readonly _bookmarksConsumer?: (database: string, bookmark: Iterable<string>) => Promise<void>
   ) {
 
   }
 
-  updateBookmarks (database: string, previousBookmarks: Iterable<string>, newBookmarks: Iterable<string>): void {
+  async updateBookmarks (database: string, previousBookmarks: Iterable<string>, newBookmarks: Iterable<string>): Promise<void> {
     const bookmarks = this._getOrInitializeBookmarks(database)
     for (const bm of previousBookmarks) {
       bookmarks.delete(bm)
@@ -123,7 +123,7 @@ class Neo4jBookmarkManager implements BookmarkManager {
       bookmarks.add(bm)
     }
     if (typeof this._bookmarksConsumer === 'function') {
-      this._bookmarksConsumer(database, [...bookmarks])
+      await this._bookmarksConsumer(database, [...bookmarks])
     }
   }
 
@@ -136,11 +136,11 @@ class Neo4jBookmarkManager implements BookmarkManager {
     return maybeBookmarks
   }
 
-  getBookmarks (database: string): Iterable<string> {
+  async getBookmarks (database: string): Promise<Iterable<string>> {
     const bookmarks = new Set(this._bookmarksPerDb.get(database))
 
     if (typeof this._bookmarksSupplier === 'function') {
-      const suppliedBookmarks = this._bookmarksSupplier(database) ?? []
+      const suppliedBookmarks = await this._bookmarksSupplier(database) ?? []
       for (const bm of suppliedBookmarks) {
         bookmarks.add(bm)
       }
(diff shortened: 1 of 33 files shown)


packages/core/src/bookmark-manager.ts
```diff
@@ -0,0 +1,166 @@
...
+ */
+export default class BookmarkManager {
+  constructor () {
+    throw new Error('Not implemented')
+  }
+
+  /**
+   * Method called when the bookmarks get updated when a transaction finished.
+   *
+   * This method will be called during when auto-commit queries finished and explicit transactions
```
robsdedude:
```suggestion
   * This method will be called when auto-commit queries finish and when explicit transactions
```

packages/core/src/bookmark-manager.ts
```diff
@@ -0,0 +1,166 @@
...
+   * get commited.
+   * @param {string} database The database which the bookmarks belongs to
+   * @param {string[]} previousBookmarks The bookmarks used during the transaction creation
```
robsdedude:
```suggestion
   * @param {string[]} previousBookmarks The bookmarks used when starting the transaction
```

packages/core/src/bookmark-manager.ts
```diff
@@ -0,0 +1,166 @@
...
+   * @param {string[]} newBookmarks The new bookmarks resolved at the end of the transaction.
```
robsdedude:
```suggestion
   * @param {string[]} newBookmarks The new bookmarks received at the end of the transaction.
```

packages/core/src/bookmark-manager.ts
```diff
@@ -0,0 +1,166 @@
...
+   * @returns {void}
+  */
+  updateBookmarks (database: string, previousBookmarks: string[], newBookmarks: string[]): void {
+    throw new Error('Not implemented')
+  }
+
+  /**
+   * Method called by the driver to get the bookmark for one specific database
```
robsdedude:
```suggestion
   * Method called by the driver to get the bookmarks for one specific database
```

packages/core/src/bookmark-manager.ts
```diff
@@ -0,0 +1,166 @@
...
+   *
+   * @param {string} database The database which the bookmarks belongs to
```
robsdedude:
```suggestion
   * @param {string} database The database which the bookmarks belong to
```

packages/core/src/bookmark-manager.ts
```diff
@@ -0,0 +1,166 @@
...
+   * @returns {string[]} The set of bookmarks
+   */
+  getBookmarks (database: string): string[] {
+    throw new Error('Not implemented')
+  }
+
+  /**
+   * Method called by the driver for getting all the bookmarks.
+   *
+   * The return of this method should be all the bookmarks present in the BookmarkManager for all databases.
```
robsdedude:
```suggestion
   * This method should return all bookmarks for all databases present in the BookmarkManager.
```

packages/core/src/bookmark-manager.ts
```diff
@@ -0,0 +1,166 @@
...
+    newBookmarks.forEach(bm => bookmarks.add(bm))
+
+    if (typeof this._bookmarksConsumer === 'function') {
+      this._bookmarksConsumer(database, [...bookmarks])
+    }
+  }
+
+  private _getOrInitializeBookmarks (database: string): Set<string> {
+    let maybeBookmarks = this._bookmarksPerDb.get(database)
+    if (maybeBookmarks == null) {
```
robsdedude: Why `maybeBookmarks == null` over `maybeBookmarks === undefined`?

packages/core/src/bookmark-manager.ts
```diff
@@ -0,0 +1,166 @@
...
+    }
+    return maybeBookmarks
+  }
+
+  getBookmarks (database: string): string[] {
+    const bookmarks = this._bookmarksPerDb.get(database) ?? []
+
+    if (typeof this._bookmarksSupplier === 'function') {
+      const suppliedBookmarks = this._bookmarksSupplier(database) ?? []
+      return [...bookmarks, ...suppliedBookmarks]
```
robsdedude: Should we use a `Set` to get rid of potential duplicates that the bookmark supplier might give us?

packages/core/src/bookmark-manager.ts
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    return [...bookmarks]
+  }
+
+  getAllBookmarks (): string[] {
+    const bookmarks = []
```
robsdedude: Should we use a `Set` here too to also get rid of potential duplicates that the bookmark supplier might give us?

packages/core/src/driver.ts
```diff
@@ -87,13 +88,15 @@ type CreateSession = (args: {
   reactive: boolean
   fetchSize: number
   impersonatedUser?: string
+  bookmarkManager?: BookmarkManager
```
robsdedude, line 91: That's not what the ADR specifies, is it? :thinking:  
I would've expected an `ignoreBookmarkManager` option instead.

Which is also what I see in the implementation below.

packages/core/src/index.ts
```diff
@@ -214,7 +217,9 @@ export type {
   NotificationPosition,
   QueryResult,
   ResultObserver,
-  TransactionConfig
+  TransactionConfig,
+  BookmarkManager,
+  BookmarkManagerConfig
```
robsdedude: Probably too late to mention it here as this is how it's spec'ed out in the ADR, but it feels a slightly off that `BookmarkManagerConfig` is not the config for `BookmarkManager`, but only for the default implementation of the `BookmarkManager` provided by the driver lib.

packages/core/src/session.ts
```diff
@@ -332,6 +338,11 @@ class Session {
     return this._lastBookmarks.values()
   }
 
+  private _bookmarks (): Bookmarks {
+    const bookmarks = this._bookmarkManager?.getAllBookmarks() ?? []
+    return new Bookmarks([...bookmarks, ...this._lastBookmarks])
```
robsdedude: If I read this all correctly, this is not exactly how the ADR describes it:
Once the session received bookmarks from the server, it should only rely on the bookmarks of the BMM, but no longer on its own bookmarks. This will make a difference, if there is another session (`s2`) running for the same db that fetches `s1._lastBookmarks` via the BMM and completes a TX replacing `s1._lastBookmarks` with `s2._lastBookmarks` in the BMM. This code will still send `s1._lastBookmarks` over the wire. It's not incorrect as the server will ignore the outdated bookmarks but it's wasteful still.

packages/core/src/session.ts
```diff
@@ -283,14 +289,14 @@ class Session {
...
-      onBookmarks: this._updateBookmarks.bind(this),
+      onBookmarks: (newBm, oldBm, db) => this._updateBookmarks(newBm, oldBm, db),
       onConnection: this._assertSessionIsOpen.bind(this),
       reactive: this._reactive,
       fetchSize: this._fetchSize,
       lowRecordWatermark: this._lowRecordWatermark,
       highRecordWatermark: this._highRecordWatermark
     })
-    tx._begin(this._lastBookmarks, txConfig)
+    tx._begin(() => this._bookmarks(), txConfig)
```
robsdedude, line 302: Why do we deffer the bookmark fetching?

packages/core/test/bookmark-manager.test.ts
```diff
@@ -0,0 +1,213 @@
+/**
```
robsdedude, line 1: TODO @robsdedude: continue review from here

packages/core/test/bookmark-manager.test.ts
```diff
@@ -0,0 +1,213 @@
...
+        manager.getAllBookmarks(),
+        newBookmarks
+      )
+
+      expect(bookmarksConsumer).toBeCalledWith('neo4j', newBookmarks)
+    })
+  })
+
+  describe('forget()', () => {
+    it('should forgot database', () => {
```
robsdedude:
```suggestion
    it('should forget database', () => {
```

packages/core/test/bookmark-manager.test.ts
```diff
@@ -0,0 +1,213 @@
...
+        'neo4j',
+        manager.getAllBookmarks(),
+        newBookmarks
+      )
+
+      expect(bookmarksConsumer).toBeCalledWith('neo4j', newBookmarks)
+    })
+  })
+
+  describe('forget()', () => {
```
robsdedude, line 253: Consider adding a test for `forget("nonexistent-db")` to see if the BMM can handle that.

packages/core/test/bookmark-manager.test.ts
```diff
@@ -0,0 +1,213 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import {
+  bookmarkManager
+} from '../src/bookmark-manager'
+
+describe('BookmarkManager', () => {
```
robsdedude, line 24: For Python I also added tests to check if duplicate BMs are being removed (duplicates in the initial bookmarks, duplicates from the supplier, duplicates that are both in the BMM and being supplied by the supplier). I'm not sure this even applies to the JS BMM, but if it does, it's be worth adding tests for it.

packages/core/test/bookmark-manager.test.ts
```diff
@@ -0,0 +1,213 @@
...
+          ['system', systemBookmarks]
+        ])
+      })
+
+      const bookmarks = manager.getAllBookmarks()
+
+      expect(bookmarks).toEqual([...neo4jBookmarks, ...systemBookmarks])
+    })
+
+    it('should return empty if there isnt bookmarks for any db', () => {
```
robsdedude:
```suggestion
    it('should return empty if there are no bookmarks for any db', () => {
```

packages/core/test/bookmark-manager.test.ts
```diff
@@ -0,0 +1,213 @@
...
+      const manager = bookmarkManager({})
+
+      const bookmarks = manager.getAllBookmarks()
+
+      expect(bookmarks).toEqual([])
+    })
+
+    it('should return enrich bookmarks list with supplied bookmarks', () => {
```
robsdedude:
```suggestion
    it('should return enriched bookmarks list with supplied bookmarks', () => {
```

packages/core/src/bookmark-manager.ts
```diff
@@ -0,0 +1,166 @@
...
+  bookmarksSupplier?: (database?: string) => string[]
+  bookmarksConsumer?: (database: string, bookmarks: string[]) => void
+}
+
+/**
+ * @typedef {Object} BookmarkManagerConfig
+ * @property {Map<string,string[]>} [initialBookmarks] Defines the initial set of bookmarks. The key is the database name and the values are the bookmarks.
+ * @property {function([database]: string):string[]} [bookmarksSupplier] Called for supplying extra bookmarks to the BookmarkManager
+ * 1. supplying bookmarks from the given database when the default BookmarkManager's `.getBookmarks(database)` gets called.
+ * 2. supplying all the bookmarks when the default BookmarkManager's  `.getAllBookmarks()` gets called
```
robsdedude:
```suggestion
 * 2. supplying all the bookmarks when the default BookmarkManager's `.getAllBookmarks()` gets called
```

packages/core/test/bookmark-manager.test.ts
```diff
@@ -0,0 +1,213 @@
...
+        initialBookmarks: new Map([
+          ['neo4j', neo4jBookmarks],
+          ['system', systemBookmarks]
+        ]),
+        bookmarksSupplier
+      })
+
+      const bookmarks = manager.getAllBookmarks()
+
+      expect(bookmarks.sort()).toEqual(
```
robsdedude: Might be premature optimization but you could do the `.sort()`ed comparison for all tests because we don't really care for the order of the bookmarks. So should that implementation detail ever change, will the future maintainer not wonder if the order is relevant or not. Plus, they won't have to change the test code.

packages/core/test/bookmark-manager.test.ts
```diff
@@ -0,0 +1,213 @@
...
+      })
+
+      manager.getAllBookmarks()
+
+      expect(bookmarksSupplier).toBeCalledWith()
+    })
+  })
+
+  describe('updateBookmarks()', () => {
+    it('should remove previous bookmarks and new bookmarks an existing db', () => {
```
robsdedude:
```suggestion
    it('should remove previous bookmarks and new bookmarks for an existing db', () => {
```

packages/core/test/bookmark-manager.test.ts
```diff
@@ -0,0 +1,213 @@
...
+      const manager = bookmarkManager({
+        initialBookmarks: new Map([
+          ['neo4j', neo4jBookmarks],
+          ['system', systemBookmarks]
+        ])
+      })
+
+      manager.updateBookmarks(
+        'neo4j',
+        manager.getAllBookmarks(),
```
robsdedude: Could be worth adding a previous bookmark that does not exist in the BMM. In the real world this could also happen if another tx has caused the BMM to remove a BM that was still there when the first tx got started.

packages/core/test/session.test.ts
```diff
@@ -310,6 +314,203 @@ describe('session', () => {
...
+          bookmarks: new bookmarks.Bookmarks([...neo4jBookmarks, ...systemBookmarks, ...lastBookmarks])
+        })
+      ])
+    })
+
+    it.each([
+      [undefined],
+      ['neo4j'],
+      ['adb']
+    ])('should not call run updateBookmarks when bookmarks since it didnt have bookmarks', async (metaDb) => {
```
```suggestion
    ])('should not call updateBookmarks when server returns no bookmarks', async (metaDb) => {
```

packages/core/test/session.test.ts
```diff
@@ -310,6 +314,203 @@ describe('session', () => {
...
+      expect(updateBookmarksSpy).not.toBeCalled()
+    })
+  })
+
+  describe('.commit()', () => {
+    it.each([
+      [undefined, 'neo4j'],
+      ['neo4j', 'neo4j'],
+      ['adb', 'adb']
+    ])('should call run updateBookmarks when bookmarks is not empty', async (metaDb, updateDb) => {
```
```suggestion
    ])('should call updateBookmarks when server returns non-empty bookmarks', async (metaDb, updateDb) => {
```

packages/core/test/session.test.ts
```diff
@@ -310,6 +314,203 @@ describe('session', () => {
...
+
+      expect(updateBookmarksSpy).toBeCalledTimes(1)
+      expect(updateBookmarksSpy).toBeCalledWith(updateDb, [...neo4jBookmarks, ...systemBookmarks], customBookmarks)
+    })
+
+    it.each([
+      [undefined],
+      ['neo4j'],
+      ['adb']
+    ])('should not call run updateBookmarks when bookmarks is empty', async (metaDb) => {
```
```suggestion
    ])('should not call updateBookmarks when server returns no bookmarks', async (metaDb) => {
```

packages/core/test/session.test.ts
```diff
@@ -310,6 +314,203 @@ describe('session', () => {
...
+      const connection = mockBeginWithSuccess(newFakeConnection())
+
+      const { session } = setupSession({
+        connection,
+        bookmarkManager: manager,
+        beginTx: false,
+        database: 'neo4j'
+      })
+
+      await session.run('query')
```
robsdedude: Were in `describe('.beginTransaction()'`. What is this test doing here?

packages/core/test/session.test.ts
```diff
@@ -349,45 +550,254 @@ describe('session', () => {
...
+          bookmarks: new bookmarks.Bookmarks([...neo4jBookmarks, ...systemBookmarks, ...lastBookmarks])
+        })
+      )
+    })
+
+    it.each([
+      [undefined, 'neo4j'],
+      ['neo4j', 'neo4j'],
+      ['adb', 'adb']
+    ])('should call run updateBookmarks when bookmarks is not empty', async (metaDb, updateDb) => {
```
```suggestion
    ])('should call updateBookmarks when server returns non-empty bookmarks', async (metaDb, updateDb) => {
```

packages/core/test/session.test.ts
```diff
@@ -349,45 +550,254 @@ describe('session', () => {
...
+      afterComplete({ db: metaDb, bookmark: customBookmarks })
+
+      expect(updateBookmarksSpy).toBeCalledWith(updateDb, [...neo4jBookmarks, ...systemBookmarks], customBookmarks)
+    })
+
+    it.each([
+      [undefined],
+      ['neo4j'],
+      ['adb']
+    ])('should not call run updateBookmarks when bookmarks is empty', async (metaDb) => {
```
```suggestion
    ])('should not call updateBookmarks when server returns no bookmarks', async (metaDb) => {
```

packages/neo4j-driver-lite/src/index.ts
```diff
@@ -199,6 +202,21 @@ const {
  *         logger: (level, message) => console.log(level + ' ' + message)
  *       },
  *
+ *       // Configure a BookmarkManager for in the driver
```
robsdedude:
```suggestion
 *       // Configure a BookmarkManager for the driver to use
```

packages/neo4j-driver-lite/src/index.ts
```diff
@@ -199,6 +202,21 @@ const {
...
+ *       //
+ *       // BookmarkManager is a piece of software responsible for keeping casual concistency between different sessions by sharing bookmarks
```
robsdedude:
```suggestion
 *       // A BookmarkManager is a piece of software responsible for keeping casual consistency between different sessions by sharing bookmarks
```

packages/neo4j-driver-lite/src/index.ts
```diff
@@ -199,6 +202,21 @@ const {
...
+ *       // between the them.
+ *       // Enabling it is done by supplying an BookmarkManager implementation instance to this param.
+ *       // A default implementation could be acquired by calling bookmarkManager factory function.
```
robsdedude:
```suggestion
 *       // A default implementation could be acquired by calling the factory function bookmarkManager.
```

packages/neo4j-driver-lite/src/index.ts
```diff
@@ -199,6 +202,21 @@ const {
...
+ *       //
+ *       // **Warning**: Enabling Bookmark Manager can have a negative impact performance wise since all the queries will wait for the last changes
```
robsdedude:
```suggestion
 *       // **Warning**: Enabling the BookmarkManager can have a negative impact on performance since all the queries will wait for the latest changes
```

packages/neo4j-driver-lite/src/index.ts
```diff
@@ -199,6 +202,21 @@ const {
...
+ *       // being propagated accross the cluster.
+ *       // For keeping concistency between a group of queries, use Session for grouping them.
```
robsdedude:
```suggestion
 *       // For keeping consistency between a group of queries, use Session for grouping them.
```

packages/neo4j-driver/src/index.js
```diff
@@ -185,6 +186,21 @@ const {
  *         logger: (level, message) => console.log(level + ' ' + message)
  *       },
  *
+ *       // Configure a BookmarkManager for in the driver
```
robsdedude: See comments on the same documentation in the lite driver

packages/neo4j-driver-lite/src/index.ts
```diff
@@ -199,6 +202,21 @@ const {
...
  *
+ *       // Configure a BookmarkManager for in the driver
+ *       //
+ *       // BookmarkManager is a piece of software responsible for keeping casual concistency between different sessions by sharing bookmarks
+ *       // between the them.
+ *       // Enabling it is done by supplying an BookmarkManager implementation instance to this param.
+ *       // A default implementation could be acquired by calling bookmarkManager factory function.
+ *       //
+ *       // **Warning**: Enabling Bookmark Manager can have a negative impact performance wise since all the queries will wait for the last changes
+ *       // being propagated accross the cluster.
```
robsdedude:
```suggestion
 *       // being propagated across the cluster.
```

packages/core/src/bookmark-manager.ts
bigmontz: `== null` includes `undefined` (actually, means `null` or `undefined`), but I will change for `=== undefined` since the method `Map.prototype.get()` returns `undefined` for non-existing keys.

packages/core/src/bookmark-manager.ts
bigmontz: Yes, I was thinking about it during the implementation. The methods of `Set` are not that ergonomic as the ones from `Array` in JS and the value will transformed to `Bookmarks`, which removes duplicates. However, for an external user, it could a bit of pain have duplicates in this list.

packages/core/src/driver.ts
bigmontz: This is the internal api used by the `driver.session()`, see `driver.session()` and `driver._newSession`

packages/core/src/index.ts
bigmontz: Yes, it is the config for the `neo4j.bookmarkManager` factory. Different configs could also creates instances of different classes.

packages/core/src/session.ts
bigmontz: Because we are only sure about the bookmarks after acquire the connection, it is a bit of leftover from the version where we needed to `manager.getAllBookmarks(['system', session.database])`

packages/core/src/session.ts
bigmontz: Sure, the ADR was changed after the code is already implemented.

packages/core/test/session.test.ts
bigmontz: I will check it out.

packages/core/src/bookmark-manager.ts
bigmontz: Implementation has been updated.

packages/core/src/session.ts
bigmontz: Actually, now it is needed, because the begin method is not async

packages/core/src/driver.ts
robsdedude: Ah, thanks!

packages/core/test/utils/matchers.ts
```diff
@@ -0,0 +1,43 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    interface Matchers<R> {
+      toBeSortedEqual: (expected: Iterable<string>) => CustomMatcherResult
+    }
+  }
+}
+
+export function installMatchers (): void {
+  expect.extend({
+    toBeSortedEqual (this: jest.MatcherContext, received: Iterable<string>, expected: Iterable<string>): jest.CustomMatcherResult {
```
robsdedude, line 32: could this be
```suggestion
    toBeSortedEqual (this: jest.MatcherContext, received: Iterable<any>, expected: Iterable<any>): jest.CustomMatcherResult {
```

to be of more general use for the *future*™?

robsdedude (approved):
:bookmark: :bowtie:

packages/core/test/utils/matchers.ts
bigmontz: It was causing type issues™ and I'd like to let myself of the future solve it :D

packages/core/src/driver.ts
```diff
@@ -96,7 +96,110 @@ interface DriverConfig {
...
+ * @param {string} defaultAccessMode=WRITE - The access mode of this session, allowed values are {@link READ} and {@link WRITE}.
+ * @param {string|string[]} bookmarks - The initial reference or references to some previous
+ * transactions. Value is optional and absence indicates that that the bookmarks do not exist or are unknown.
+ * @param {number} fetchSize - The record fetch size of each batch of this session.
+ * Use {@link FETCH_ALL} to always pull all records in one batch. This will override the config value set on driver config.
+ * @param {string} database - The database this session will operate on.
+ * @param {string} impersonatedUser - The username which the user wants to impersonate for the duration of the session.
+ * @param {BookmarkManager} [bookmarkManager] = The bookmark manager
+ */
+class SessionConfig {
```
fbiville, line 106: isn't going from a raw object to a class a breaking change? (which is OK for 5.0 anyway, just wondering)

packages/neo4j-driver/test/index.test.js
```diff
@@ -139,7 +139,7 @@ describe('#unit index', () => {
       }
     })
 
-    describe('await rxRession.beginTransaction().toPromise() return', () => {
+    xdescribe('await rxRession.beginTransaction().toPromise() return', () => {
```
fbiville: is that related to the other PR with rx changes?

fbiville (approved):

packages/neo4j-driver/test/index.test.js
bigmontz: ops, it is a leftover. I was skipping the test here for checking for another issues.

packages/core/src/driver.ts
bigmontz: Nope, typescript has loose type. This class will behave an interface actually, I need to define as class to make it possible add the experimental annotations in place. 

See usage example: https://github.com/neo4j/neo4j-javascript-driver/pull/974/files#diff-0ec3f1aeef14ced21b829fd32b84fb3a078e728e5fe474d34b33912dee2b7a1bR124

packages/core/src/internal/connection-holder.ts
```diff
@@ -161,12 +163,16 @@ class ConnectionHolder implements ConnectionHolderInterface {
   }
 
   private async _createConnectionPromise (connectionProvider: ConnectionProvider): Promise<Connection | null> {
+    this._connectionError = undefined
```
fbiville: you do not need to reset it here, do you?

packages/core/src/transaction.ts
```diff
@@ -154,6 +159,10 @@ class Transaction {
         }
         this._onError(error).catch(() => {})
       })
+      // It should make the transaction active anyway
+      // futher errors will be treated by the exiting
```
fbiville: cannot edit for some reason:
```
// further errors will be treated by the existing
```

packages/core/src/transaction.ts
```diff
@@ -347,31 +364,34 @@ const _states = {
         reactive,
         fetchSize,
         highRecordWatermark,
-        lowRecordWatermark
+        lowRecordWatermark,
+        preparationJob
       }: StateTransitionParams
     ): any => {
       // RUN in explicit transaction can't contain bookmarks and transaction configuration
       // No need to include mode and database name as it shall be inclued in begin
```
fbiville: old typo but: should be `included` instead of `inclued`

packages/core/test/transaction.test.ts
```diff
@@ -343,11 +345,36 @@ function testTx<T extends Transaction> (transactionName: string, newTransaction:
...
 
+        tx._begin(async () => Bookmarks.empty(), TxConfig.empty())
+
         const result = tx.run('RETURN 1')
 
         // @ts-expect-error
         expect(result._watermarks).toEqual({ high: 700, low: 300 })
       })
+
+      it('should wait begin message be send', async () => {
```
fbiville: s/send/sent/

packages/core/test/transaction.test.ts
```diff
@@ -376,6 +405,29 @@ function testTx<T extends Transaction> (transactionName: string, newTransaction:
             expect(error).toEqual(expectedError)
           }
         })
+
+        it('should wait begin message be send', async () => {
```
fbiville: s/send/sent/

fbiville (approved):

packages/core/src/internal/connection-holder.ts
bigmontz: I removed the changes in the connection-holder, to avoid effects in other areas.

robsdedude (approved):
:stopwatch:

packages/core/test/transaction.test.ts
robsdedude:
```suggestion
      it('should wait until begin message was sent', async () => {
```

robsdedude: (Once more below)

bigmontz:
Depends on: https://github.com/neo4j-drivers/testkit/pull/513

PR #980, opened 2022-08-29
Fix RxSession in Testkit Backend

bigmontz:
Testkit environment variables without `TEST_` prefix are not forward by the Testkit to the
scripts in the driver. So we need change the session type variable expected by the `backend.py`
to  be `TEST_SESSION_TYPE` and then remove the prefix before redirect to the backend.

This PR also fix the `SessionRun` and `TransactionRun` to not initiate the stream.

**Warn**: `RxSession.beginTransaction` and `RxSession.execute*` were tweaked. The `Observable<RxTransaction>` only emits the `RxTransaction` when the begin is executed with success. The begin still defer the execution to the point observer is subscribed.

### DIFF
--- a/packages/neo4j-driver/src/result-rx.js
+++ b/packages/neo4j-driver/src/result-rx.js
@@ -35,8 +35,9 @@ export default class RxResult {
    * @constructor
    * @protected
    * @param {Observable<Result>} result - An observable of single Result instance to relay requests.
+   * @param {number} state - The streamming state
    */
-  constructor (result) {
+  constructor (result, state) {
     const replayedResult = result.pipe(publishReplay(1), refCount())
 
     this._result = replayedResult
@@ -48,7 +49,7 @@ export default class RxResult {
     this._records = undefined
     this._controls = new StreamControl()
     this._summary = new ReplaySubject()
-    this._state = States.READY
+    this._state = state || States.READY
   }
 
   /**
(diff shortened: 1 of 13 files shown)


bigmontz:
Depends on: https://github.com/neo4j-drivers/testkit/pull/512

packages/neo4j-driver/src/result-rx.js
```diff
@@ -35,8 +35,9 @@ export default class RxResult {
    * @constructor
    * @protected
    * @param {Observable<Result>} result - An observable of single Result instance to relay requests.
+   * @param {number} state - The streamming state
```
fbiville:
```suggestion
   * @param {number} state - The streaming state
```

packages/testkit-backend/src/feature/common.js
```diff
@@ -33,7 +33,7 @@ const features = [
   'Feature:API:ConnectionAcquisitionTimeout',
   'Feature:API:Driver:GetServerInfo',
   'Feature:API:Driver.VerifyConnectivity',
-  'Feature:API:Result.Peek',
```
fbiville, line 36: why was this one removed?

packages/testkit-backend/src/skipped-tests/rx.js
```diff
@@ -4,6 +4,15 @@ const skippedTests = [
   skip(
     'Throws after run insted of the first next because of the backend implementation',
     ifEquals('stub.disconnects.test_disconnects.TestDisconnects.test_disconnect_on_tx_begin')
+  ),
+  skip(
+    'Backend could not support this test',
```
fbiville: what happened with these?

packages/testkit-backend/src/feature/common.js
bigmontz: Because reactive doesn't support Peek. This feature is enable for async at `packages/testkit-backend/src/feature/async.js`

packages/testkit-backend/src/skipped-tests/rx.js
bigmontz: The error occurs in tricky place in the reactive backend. The reactive backend implements a custom subscriber to convert the reactive stream to `AsyncIterator`, the error happens there and it happens in a way it is difficult to align with the testkit premises.

robsdedude: But is there a reasonable way for users to handle this? Because there sure should be a way how users can handle unknown tz ids coming from the server.

bigmontz: It should have, I will double check it again.

bigmontz: It was actually the driver log causing issues. When the debug is turned on, the driver tries to access the broken object and it moves the point where the error occurs (it happens earlier, before the user get access to the record). 

I made the broken object a bit more lenient and I added a especial serialisation for this kind of object in the `json.stringify` wrapper.

fbiville (approved):

PR #983, opened 2022-09-02
Fix serialization of broken objects in the driver logs

bigmontz:
Trying to logging broken objects was causing the driver fails in unexpected moments.
This broken behaviour makes the driver fail different when the `debug` log is enabled.

### DIFF
--- a/packages/core/src/internal/object-util.ts
+++ b/packages/core/src/internal/object-util.ts
@@ -1,0 +1,87 @@
+/**
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+// eslint-disable-next-line @typescript-eslint/naming-convention
+const __isBrokenObject__ = '__isBrokenObject__'
+// eslint-disable-next-line @typescript-eslint/naming-convention
+const __reason__ = '__reason__'
+
+/**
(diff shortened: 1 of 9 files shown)


packages/core/src/internal/object-util.ts
```diff
@@ -0,0 +1,87 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+// eslint-disable-next-line @typescript-eslint/naming-convention
+const __isBrokenObject__ = '__isBrokenObject__'
+// eslint-disable-next-line @typescript-eslint/naming-convention
+const __reason__ = '__reason__'
+
+/**
+ * Creates a object which all method call will throw the given error
```
robsdedude:
```suggestion
 * Creates a object on which all method calls will throw the given error
```

packages/core/src/internal/object-util.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  return new Proxy(object, {
+    get: (_: T, p: string | Symbol): any => {
+      if (p === __isBrokenObject__) {
+        return true
+      } else if (p === __reason__) {
+        return error
+      } else if (p === 'toJSON') {
+        return undefined
+      }
```
robsdedude, line 44: See [Slack discussion (team internal link)](https://neo4j.slack.com/archives/G493NJPF0/p1662376150650769).

robsdedude: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

packages/core/test/json.test.ts
```diff
@@ -0,0 +1,49 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { json, newError } from '../src'
+import { createBrokenObject } from '../src/internal/object-util'
+
+describe('json', () => {
```
robsdedude, line 23: Might be worth opening another PR for adding tests to check the custom json serializer also works with bigint (if the tests don't already exist elsewhere).

robsdedude (approved):
:feather: :page_facing_up:

packages/core/test/json.test.ts
bigmontz: Done

robsdedude: :rocket:

PR #985, opened 2022-09-05
Add stricter string validation to `neo4j.int`

bigmontz:
`neo4j.int` could have some surprising result when used with string. For avoiding this problem,
a configuration option called `strictStringValidation` was added. When enable, `strictStringValidation` will trigger a deeper validation of the string. This option could slow down the conversion.

### DIFF
--- a/packages/core/src/integer.ts
+++ b/packages/core/src/integer.ts
@@ -833,10 +833,12 @@ class Integer {
    * @access private
    * @param {string} str The textual representation of the Integer
    * @param {number=} radix The radix in which the text is written (2-36), defaults to 10
+   * @param {Object} [opts={}] Configuration options
+   * @param {boolean} [opts.strictStringValidation=false] Enable strict validation generated Integer.
    * @returns {!Integer} The corresponding Integer value
    * @expose
    */
-  static fromString (str: string, radix?: number): Integer {
+  static fromString (str: string, radix?: number, { strictStringValidation }: { strictStringValidation?: boolean} = {}): Integer {
     if (str.length === 0) {
       throw newError('number format error: empty string')
     }
@@ -867,7 +869,13 @@ class Integer {
     let result = Integer.ZERO
     for (let i = 0; i < str.length; i += 8) {
       const size = Math.min(8, str.length - i)
-      const value = parseInt(str.substring(i, i + size), radix)
+      const valueString = str.substring(i, i + size)
+      const value = parseInt(valueString, radix)
+
+      if (strictStringValidation === true && !_isValidNumberFromString(valueString, value, radix)) {
+        throw newError(`number format error: "${valueString}" is NaN in radix ${radix}: ${str}`)
+      }
+
       if (size < 8) {
         const power = Integer.fromNumber(Math.pow(radix, size))
         result = result.multiply(power).add(Integer.fromNumber(value))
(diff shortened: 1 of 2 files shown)


packages/core/src/integer.ts
```diff
@@ -867,7 +869,13 @@ class Integer {
     let result = Integer.ZERO
     for (let i = 0; i < str.length; i += 8) {
       const size = Math.min(8, str.length - i)
-      const value = parseInt(str.substring(i, i + size), radix)
+      const valueString = str.substring(i, i + size)
+      const value = parseInt(valueString, radix)
+
+      if (strictStringValidation === true && !_isValidNumberFromString(valueString, value, radix)) {
+        throw newError(`number format error: "${valueString}" is NaN in radix ${radix}: ${str}`)
```
OskarDamkjaer, line 876: Would this raise a generic error? Perhaps there's some validation error code that could be used? (just interested on change needed)

packages/core/test/integer.test.ts
```diff
@@ -256,6 +256,26 @@ describe('Integer', () => {
...
+    expect(() => Integer.fromString('78a91', undefined, { strictStringValidation: true })).toThrow(
+      newError('number format error: "78a91" is NaN in radix 10: 78a91')
+    ))
+
+  test('Integer.fromString("a7891", undefined, { strictStringValidation: true }) toThrow invalid character', () =>
+    expect(() => Integer.fromString('a7891', undefined, { strictStringValidation: true })).toThrow(
+      newError('number format error: "a7891" is NaN in radix 10: a7891')
+    ))
+
+  test('Integer.fromString("7010", 2, { validatedString: true }) toThrow invalid character', () =>
```
OskarDamkjaer:
```suggestion
  test('Integer.fromString("7010", 2, { strictStringValidation: true }) toThrow invalid character', () =>
```

OskarDamkjaer (approved):

packages/core/src/integer.ts
bigmontz: I've kept the error similar the ones we have with other validation errors. It will raise a `Neo4jError` with `code` equals to `N/A`. 

See: https://github.com/neo4j/neo4j-javascript-driver/pull/985/files#diff-654bf635043151007a5b9eecd461ba5d922f5c79bc3e5bc702a75d4d855440eeR843

packages/core/src/integer.ts
```diff
@@ -946,6 +956,20 @@ class Integer {
...
+ * @private
+ * @param theString
+ * @param theNumber
+ * @param radix
+ * @return {boolean} True if valid
+ */
+function _isValidNumberFromString (theString: string, theNumber: number, radix: number): boolean {
+  return !Number.isNaN(theString) &&
+    !Number.isNaN(theNumber) &&
+    theNumber.toString(radix).toLocaleLowerCase() === theString.toLocaleLowerCase()
```
robsdedude, line 970: Why `toLocaleLowerCase`? What non-standard capitalization are we expecting to be returned from `theNumber.toString(radix)`?

robsdedude: Also

shouldn't `_isValidNumberFromString("a", 11, 16)` be `true`?

bigmontz: No, because `a` hex is `10` dec.  But _isValidNumberFromString("a", 10, 16) will be true

robsdedude: Basically saying: please add tests with different radices :upside_down_face:

bigmontz: The `toLocaleLowerCase` was an auto-complete error. I will fix in a next PR since I'd like this fix to be part of the beta.

bigmontz: I tested radix = 2 :D

robsdedude: > No, because `a` hex is `10` dec. But _isValidNumberFromString("a", 10, 16) will be true

:facepalm: doh'!

PR #986, opened 2022-09-06
Fix Zoned DateTime support for dates before common era

bigmontz:
Years previous the common era were being treated as positive numbers causing the dates appear
as positive numbers and breaking the auto-zone offset detection for zoned date times which were
not created with the offset.


packages/neo4j-driver/test/temporal-types.test.js
```diff
@@ -38,7 +38,7 @@ const MAX_TEMPORAL_ARRAY_LENGTH = 1000
  */
 const MAX_DURATION_COMPONENT = 3000000000000
 const MAX_NANO_OF_SECOND = 999999999
-const MAX_YEAR = 999999999
+const MAX_YEAR = 99_999
```
robsdedude, line 41: Nit: feels like `packages/bolt-connection/test/bolt/bolt-protocol-v*.test.js` could also make use of this variable instead of hard coding it into the tests.

robsdedude (approved):

packages/neo4j-driver/test/temporal-types.test.js
bigmontz: They are different packages, so they could not share test code 😢.

PR #987, opened 2022-09-06
Add stricter string validation to `neo4j.int` (#985)

bigmontz:
`neo4j.int` could have some surprising result when used with string. For avoiding this problem,
a configuration option called `strictStringValidation` was added. When enable, `strictStringValidation` will trigger a deeper validation of the string. This option could slow down the conversion.

Co-authored-by: Oskar Damkjaer <OskarDamkjaer@users.noreply.github.com>


bigmontz:
Cherry-picks: https://github.com/neo4j/neo4j-javascript-driver/pull/985 and https://github.com/neo4j/neo4j-javascript-driver/pull/988

PR #989, opened 2022-09-06
Fix Zoned DateTime support for dates before common era (#986)

bigmontz:
Years previous the common era were being treated as positive numbers causing the dates appear
as positive numbers and breaking the auto-zone offset detection for zoned date times which were
not created with the offset.


bigmontz:
Cherry-picks #986

PR #990, opened 2022-09-07
Fix temporal it tests

bigmontz:
New UTC zoned date time has timeZoneOffsetInSeconds values.

PR #991, opened 2022-09-07
Fix DateTime with ZoneId for years between 00-99 #992

bigmontz:
The error was happening because Date.UTC factory doesn't treats dates between 00-99 as 1900-1999.

Removing the usage of this factory makes the code slighty faster while resolves the issue.


bigmontz:
Cherry picks: https://github.com/neo4j/neo4j-javascript-driver/pull/990 and https://github.com/neo4j/neo4j-javascript-driver/pull/992

robsdedude (approved):
:tophat:

PR #992, opened 2022-09-08
Fix DateTime with ZoneId for years between 00-99

bigmontz:
The error was happening because Date.UTC factory doesn't treats dates between 00-99 as 1900-1999.

Removing the usage of this factory makes the code slighty faster while resolves the issue.

### DIFF
--- a/packages/neo4j-driver/test/examples.test.js
+++ b/packages/neo4j-driver/test/examples.test.js
@@ -708,7 +708,7 @@ describe('#integration examples', () => {
   it('service unavailable example', done => {
     const console = consoleOverride
     const consoleLoggedMsg = consoleOverridePromise
-    const uri = `bolt://${sharedNeo4j.hostname}:7688` // wrong port
+    const uri = `bolt://${sharedNeo4j.hostname}:7686` // wrong port
     const password = 'wrongPassword'
 
     // tag::service-unavailable[]
--- a/packages/neo4j-driver/test/temporal-types.test.js
+++ b/packages/neo4j-driver/test/temporal-types.test.js
@@ -1439,7 +1439,24 @@ describe('#integration temporal-types', () => {
       expect(records.length).toEqual(1)
 
       const value = records[0].get(0)
-      expect(value).toEqual(expectedValue)
+
+      if (
+        expectedValue.timeZoneId != null &&
+        value.timeZoneOffsetSeconds != null &&
+        neo4j.isDateTime(value) &&
+        neo4j.isDateTime(expectedValue)) {
+        expect(value).toEqual(jasmine.objectContaining({
+          year: expectedValue.year,
+          month: expectedValue.month,
+          day: expectedValue.day,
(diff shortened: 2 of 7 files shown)


packages/neo4j-driver/test/examples.test.js
```diff
@@ -708,7 +708,7 @@ describe('#integration examples', () => {
   it('service unavailable example', done => {
     const console = consoleOverride
     const consoleLoggedMsg = consoleOverridePromise
-    const uri = `bolt://${sharedNeo4j.hostname}:7688` // wrong port
+    const uri = `bolt://${sharedNeo4j.hostname}:7686` // wrong port
```
robsdedude, line 711: :eyes: why did it change?

packages/neo4j-driver/test/temporal-types.test.js
```diff
@@ -1439,7 +1439,24 @@ describe('#integration temporal-types', () => {
...
       const value = records[0].get(0)
-      expect(value).toEqual(expectedValue)
+
+      if (
+        expectedValue.timeZoneId != null &&
+        value.timeZoneOffsetSeconds != null &&
+        neo4j.isDateTime(value) &&
+        neo4j.isDateTime(expectedValue)) {
+        expect(value).toEqual(jasmine.objectContaining({
+          year: expectedValue.year,
```
robsdedude, line 1449: Maybe add a comment here. I'm not 100% sure about the intention of this special treatment. (same for the change further down in this file)

robsdedude (approved):
:tophat:

packages/neo4j-driver/test/examples.test.js
bigmontz: The `7688` port is the server side routing port. Since fabric is enabled by default in 5.0, this error is not happening consistently. So I move it to an unused port.

packages/neo4j-driver/test/temporal-types.test.js
bigmontz: The DateTime with zone are created without the offset, but they return with the offset from the utc patch servers. This changes have the intention of ignoring the offset in this comparison.

robsdedude: What if it was 
```
        expectedValue.timeZoneId != null &&
        expectedValue.timeZoneOffsetSeconds != null &&
```
? Would that do the same thing? If so, I think it feels cleaner to know in advance what the result should be and not decide how to check the result depending on what it looks like (whenever possible, ofc.)

bigmontz:
```
expectedValue.timeZoneId != null &&
        expectedValue.timeZoneOffsetSeconds != null &&
```

There is not such test case in the old integration test suite, because the value we send is the value we expect and the test were created with the old implementation in mind.

robsdedude: K. Let's roll with it :+1:

PR #997, opened 2022-09-16
Add property-based testing to temporal-types conversion

bigmontz:
Add this type of testing to `.toStandardDate()` ('should be the reverse operation of fromStandardDate but losing time information') helps to cover corner cases and solve special cases such:

* Negative date time not being serialised correctly in the iso standard. Years should always have 6 digits and the signal in front for working correctly with negative years and high numbers. This also avoids the year 2000 problem. See, https://en.wikipedia.org/wiki/ISO_8601
* `Date.fromStandardDate` factory was not taking in consideration the `seconds` contribution in the timezone offset. This is not a quite common scenario, but there are dates with timezone offset of for example `50 minutes` and `20 seconds`.
* Fix `Date.toStandardDate` for dates with offsets of seconds. Javascript Date constructor doesn't create dates from iso strings with seconds in the offset. For instance, `new Date("2010-01-12T14:44:53+00:00:10")`. So, the date should be re-created from the UTC timestamp.

### DIFF
--- a/packages/core/src/internal/temporal-util.ts
+++ b/packages/core/src/internal/temporal-util.ts
@@ -288,7 +288,7 @@ export function dateToIsoString (
   month: NumberOrInteger | string,
   day: NumberOrInteger | string
 ): string {
-  const yearString = formatNumber(year, 6, { usePositiveSign: true })
+  const yearString = formatYear(year)
   const monthString = formatNumber(month, 2)
   const dayString = formatNumber(day, 2)
   return `${yearString}-${monthString}-${dayString}`
@@ -578,6 +578,19 @@ function formatSecondsAndNanosecondsForDuration (
 function formatNanosecond (value: NumberOrInteger | string): string {
   value = int(value)
   return value.equals(0) ? '' : '.' + formatNumber(value, 9)
+}
+
+/**
+ *
+ * @param {Integer|number|string} year The year to be formatted
+ * @return {string} formatted year
+ */
+function formatYear (year: NumberOrInteger | string): string {
+  const yearInteger = int(year)
+  if (yearInteger.isNegative() || yearInteger.greaterThan(9999)) {
+    return formatNumber(yearInteger, 6, { usePositiveSign: true })
+  }
+  return formatNumber(yearInteger, 4)
 }
 
 /**
--- a/packages/core/src/internal/temporal-util.ts
+++ b/packages/core/src/internal/temporal-util.ts
@@ -315,6 +315,15 @@ export function toStandardDate (utc: number): Date {
 }
 
 /**
+ * Shortcut for creating a new StandardDate
+ * @param date
+ * @returns {Date} the standard date
+ */
+export function newDate (date: string | number | Date): Date {
+  return new Date(date)
+}
+
+/**
  * Get the total number of nanoseconds from the milliseconds of the given standard JavaScript date and optional nanosecond part.
  * @param {global.Date} standardDate the standard JavaScript date.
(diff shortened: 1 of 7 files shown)


packages/core/test/temporal-types.test.ts
```diff
@@ -33,15 +34,33 @@ describe('Date', () => {
...
+            return
+          }
+
+          const date = Date.fromStandardDate(standardDate)
+          const receivedDate = date.toStandardDate()
+
+          const hour = standardDate.setHours(0, -1 * receivedDate.getTimezoneOffset())
+
+          // In some situations, the setHours result in a NaN hour.
+          // In this case, the test should be discarded
```
robsdedude: What are those cases? Is it when the localized Date exceeds the range of valid Dates?

Also, isn't the return value of `setHours` `msSinceEpoch` or something like that?

packages/core/src/internal/temporal-util.ts
```diff
@@ -288,16 +288,7 @@ export function dateToIsoString (
...
-  const isNegative = year.isNegative()
-  if (isNegative) {
-    year = year.multiply(-1)
-  }
-  let yearString = formatNumber(year, 4)
-  if (isNegative) {
-    yearString = '-' + yearString
-  }
-
+  const yearString = formatNumber(year, 6, { usePositiveSign: true })
```
robsdedude: Is there a reason not to use `formatNumber(year, 4)` for years between 0 and 9999? This would increase the compatibility with other ISO implementations.

bigmontz: The issue:

```
> new Date('0099-07-04')
Sat Jul 04 0099 00:50:20 GMT+0050 (Central European Summer Time)
> new Date('+0099-07-04')
Sun Jul 04 1999 00:00:00 GMT+0200 (Central European Summer Time)
> new Date('-0099-07-04')
Sun Jul 04 1999 00:00:00 GMT+0200 (Central European Summer Time)

packages/core/test/temporal-types.test.ts
bigmontz: Case: `new Date(NaN)`

robsdedude: Is that the only one? In that case, I think it might be more robust to check for that explicitly rather than using `setHours` as proxy.

packages/core/src/internal/temporal-util.ts
robsdedude: Oh. I didn't know JS also did this with strings. Some JS design decisions just make we want to shake my head.

robsdedude (approved):
:clock10:

packages/core/src/internal/temporal-util.ts
bigmontz: Yes, but I amend the PR for making the conversion to 6 digits and sign only when it is really needed.

packages/core/test/temporal-types.test.ts
bigmontz: That's it the funny thing, the library generates this value in a way that `isNaN(date)` is false when usually is true when `new Date(NaN)` is created. The error is only catch when I try to change something in the date. I will check to see if I have other more simple scenario which blocks this kind of date.

Screenshot 2022-09-19 at 11 56 09

bigmontz: I don't manage to fix it, unfortunately.

robsdedude (approved):
:rocket:

packages/core/test/temporal-types.test.ts
bigmontz: Fixed xD

PR #998, opened 2022-09-19
Add property-base tests to `Integer`

bigmontz:


thelonelyvulpes (approved):

packages/core/test/integer.test.ts
```diff
@@ -355,6 +356,121 @@ describe('Integer', () => {
...
+      fc.property(
+        fc.bigInt({ max: Integer.MAX_SAFE_VALUE.toBigInt(), min: Integer.MIN_SAFE_VALUE.toBigInt() }),
+        bigint => bigint === int(bigint).toBigInt()
+      )
+    )
+  })
+
+  test('int(string) should be reversed by Integer.toString', () => {
+    fc.assert(fc.property(fc.integer().map(num => num.toString()), str => str === int(str).toString()))
+  })
```
robsdedude: I think you could replace this single test with these two tests which in sum (due to transitivity of ===, which hopefully holds for every sane implementation of equality) asserts the same thing + more fine-grained transformations.

```suggestion
  test('int.toString() should match Integer.toString()', () => {
    fc.assert(fc.property(fc.integer(), i => i.toString() === int(i).toString()))
  })

  test('int(string) should match int(Integer)', () => {
    fc.assert(fc.property(fc.integer(), i => int(i) === int(i.toString())))
  })
```

This also covers the original test iff. we assume that `i === j` => `i.toString() === j.toString()` for `i`, `j` of type `int`. Alternatively, if we don't want to make that assumption, we can have all three tests or add a test along the lines of

```typescript
  test('int.toString() matches for equal numbers', () => {
    fc.assert(fc.property(fc.integer(), i => int(i).toString() === int(i).toString()))
  })
```
But yeah, that one is weird looking because it tests such a basic assumption.

packages/core/test/integer.test.ts
```diff
@@ -355,6 +356,121 @@ describe('Integer', () => {
...
+
+  test('Integer.add should be Distributive', () => {
+    fc.assert(fc.property(arbitraryInteger(), arbitraryInteger(), arbitraryInteger(),
+      (a, b, c) => a.multiply(b.add(c)).equals(a.multiply(b).add(a.multiply(c)))))
+  })
+
+  test('Integer.subtract should be Distributive', () => {
+    fc.assert(fc.property(arbitraryInteger(), arbitraryInteger(), arbitraryInteger(),
+      (a, b, c) => a.multiply(b.subtract(c)).equals(a.multiply(b).subtract(a.multiply(c)))))
+  })
```
robsdedude, line 413: You could also add tests for neutral elements (`a + 0 === a`, `a * 1 === a`).

packages/core/test/integer.test.ts
```diff
@@ -355,6 +356,121 @@ describe('Integer', () => {
...
+
+  describe('For Same Signal Integers', () => {
```
robsdedude:
```suggestion
  describe('with same sign', () => {
```

packages/core/test/integer.test.ts
```diff
@@ -1188,6 +1304,20 @@ function wellFormedNumbersAndRadix (): Array<[string, number]> {
   ]
 }
 
+function arbitraryInteger (): fc.Arbitrary<Integer> {
+  return fc.record({ low: fc.integer(), high: fc.integer() })
+    .map(({ low, high }) => new Integer(low, high))
+}
+
+function arbitrarySameSignalIntegers (): fc.Arbitrary<{ a: Integer, b: Integer }> {
```
robsdedude:
```suggestion
function arbitrarySameSignIntegers (): fc.Arbitrary<{ a: Integer, b: Integer }> {
```

packages/core/test/integer.test.ts
```diff
@@ -1188,6 +1304,20 @@ function wellFormedNumbersAndRadix (): Array<[string, number]> {
...
+  return fc.record({ a: arbitraryInteger(), b: arbitraryInteger() })
+    .map(({ a, b }) => a.isPositive() === b.isPositive() ? { a, b } : { a, b: b.negate() })
+}
+
+function arbitraryDiffSignalIntegers (): fc.Arbitrary<{ a: Integer, b: Integer }> {
```
robsdedude:
```suggestion
function arbitraryDiffSignIntegers (): fc.Arbitrary<{ a: Integer, b: Integer }> {
```

packages/core/test/integer.test.ts
```diff
@@ -355,6 +356,121 @@ describe('Integer', () => {
...
+    })
+
+    test('Integer.lessThanOrEqual should return true if a - b is ZERO or negative', () => {
+      fc.assert(fc.property(
+        arbitrarySameSignalIntegers(),
+        ({ a, b }) => a.subtract(b).isNegative() ? a.lessThan(b) : !a.lessThan(b)))
+    })
+  })
+
+  describe('For Different Signal Integers', () => {
```
robsdedude:
```suggestion
  describe('with different sign', () => {
```

packages/core/test/integer.test.ts
```diff
@@ -355,6 +356,121 @@ describe('Integer', () => {
...
+        ({ a, b }) => a.subtract(b).isPositive() ? a.greaterThan(b) : !a.greaterThan(b)))
+    })
+
+    test('Integer.greaterThanOrEqual should return true if a - b is positive or ZERO', () => {
+      fc.assert(fc.property(
+        arbitrarySameSignalIntegers(),
+        ({ a, b }) => a.subtract(b).isPositive() || a.subtract(b).isZero() ? a.greaterThanOrEqual(b) : !a.greaterThanOrEqual(b)))
+    })
+
+    test('Integer.equals should return true if a - b is ZERO', () => {
```
robsdedude: Couldn't this test just be outside the case differentiation? This holds true regardless if the numbers have the same or a different sign. Or are there issues with overflows?

robsdedude (approved):
:test_tube:

packages/core/test/integer.test.ts
bigmontz: We don't have, we could just move it to outside.

packages/core/test/integer.test.ts
```diff
@@ -412,66 +412,75 @@ describe('Integer', () => {
...
+
+  test('Integer.subtract should have 0 as identity', () => {
+    fc.assert(fc.property(arbitraryInteger(), (a) => a.subtract(0).equals(a)))
+  })
+
+  test('Integer.multiply should have 0 as identity', () => {
+    fc.assert(fc.property(arbitraryInteger(), (a) => a.multiply(1).equals(a)))
+  })
+
+  test('Integer.div should have 0 as identity', () => {
```
robsdedude, line 427:
```suggestion
  test('Integer.div should have 1 as identity', () => {
```
please don't divide by 0 :wink:

packages/core/test/integer.test.ts
```diff
@@ -412,66 +412,75 @@ describe('Integer', () => {
...
 
+  test('Integer.add should have 0 as identity', () => {
+    fc.assert(fc.property(arbitraryInteger(), (a) => a.add(0).equals(a)))
+  })
+
+  test('Integer.subtract should have 0 as identity', () => {
+    fc.assert(fc.property(arbitraryInteger(), (a) => a.subtract(0).equals(a)))
+  })
+
+  test('Integer.multiply should have 0 as identity', () => {
```
robsdedude, line 423:
```suggestion
  test('Integer.multiply should have 1 as identity', () => {
```

robsdedude (approved):

packages/core/test/integer.test.ts
bigmontz: I'm creating a new math.

PR #1001, opened 2022-09-20
Add property-based testing to temporal-types conversion (#997)

bigmontz:
Add this type of testing to `.toString()` ('should be serialize string which can be loaded by new Date') helps to cover corner cases and solve special cases such:

* Negative date time not being serialised correctly in the iso standard. Years should always have 6 digits and the signal in front for working correctly with negative years and high numbers. This also avoids the year 2000 problem. See, https://en.wikipedia.org/wiki/ISO_8601
* `Date.fromStandardDate` factory was not taking in consideration the `seconds` contribution in the timezone offset. This is not a quite common scenario, but there are dates with timezone offset of for example `50 minutes` and `20 seconds`.
* Fix `Date.toString` for dates with offsets of seconds. Javascript Date constructor doesn't create dates from iso strings with seconds in the offset. For instance, `new Date("2010-01-12T14:44:53+00:00:10")`. So, this tests should be skipped.

### DIFF
--- a/packages/core/test/temporal-types.test.ts
+++ b/packages/core/test/temporal-types.test.ts
@@ -1,0 +1,96 @@
+/**
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { StandardDate } from '../src/graph-types'
+import { LocalDateTime, Date, DateTime } from '../src/temporal-types'
+import { temporalUtil } from '../src/internal'
+import fc from 'fast-check'
+
+const MIN_UTC_IN_MS = -8_640_000_000_000_000
+const MAX_UTC_IN_MS = 8_640_000_000_000_000
+const ONE_DAY_IN_MS = 86_400_000
+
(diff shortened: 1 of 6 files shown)


bigmontz:
Adapted cherry pick from: https://github.com/neo4j/neo4j-javascript-driver/pull/997

packages/core/test/temporal-types.test.ts
```diff
@@ -0,0 +1,96 @@
...
+import { temporalUtil } from '../src/internal'
+import fc from 'fast-check'
+
+const MIN_UTC_IN_MS = -8_640_000_000_000_000
+const MAX_UTC_IN_MS = 8_640_000_000_000_000
+const ONE_DAY_IN_MS = 86_400_000
+
+describe('Date', () => {
+  describe('.toString()', () => {
+    it('should be serialize string which can be loaded by new Date', () => {
```
robsdedude:
```suggestion
    it('should return a string which can be loaded by new Date', () => {
```
or `should serialize to a string which ...`

packages/core/test/temporal-types.test.ts
```diff
@@ -0,0 +1,96 @@
...
+            expect(receivedDate.getMinutes()).toEqual(adjustedDateTime.getMinutes())
+          })
+      )
+    })
+  })
+})
+
+describe('LocalDateTime', () => {
+  describe('.toString()', () => {
+    it('should be serialize string which can be loaded by new Date', () => {
```
robsdedude:
```suggestion
    it('should return a string which can be loaded by new Date', () => {
```

robsdedude (approved):
While this is a breaking API change, it's also a fix. Not sure about the severity and if it's worth potentially breaking users.

Lastly, I checked the API documentation and that function isn't documented at all. So one could always argues it's fair game to break undocumented API, especially if it fixes things.

packages/core/test/temporal-types.test.ts
```diff
@@ -0,0 +1,96 @@
...
+          expect(receivedDate).toEqual(date)
+        })
+      )
+    })
+  })
+})
+
+describe('DateTime', () => {
+  describe('.toString()', () => {
+    it('should be serialize string which can be loaded by new Date', () => {
```
robsdedude:
```suggestion
    it('should return a string which can be loaded by new Date', () => {
```

PR #1004, opened 2022-10-05
Commit `neo4j-driver-deno` to the repository

bigmontz:
The `deno driver` is generated using a custom script made for this repository.
This script doesn't have any tests linked to it.
So the only way this changes are tested is by running the `deno driver` test suite.
For guarantee the behaviour of this driver, we should version commit the generated driver to the repository.

This changes introduce the first committed version of the driver.
Adding the verification if the committed deno driver is in sync with the lite driver is also part of the this scope.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
   "scripts": {
     "clean": "lerna clean -y && lerna run clean",
     "build": "lerna bootstrap --ci",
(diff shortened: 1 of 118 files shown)


AndyHeap-NeoTech (approved):
As requested just looked at the first smaller commit. Looks sensible to me.

PR #1006, opened 2022-10-10
Introduce `Driver.executeQuery`

bigmontz:
This method gives the user a simple interface and obvious place to start with the driver. Behind this method the full retry mechanism will be present. The results are eagerly returned and in memory. With this, we have removed the need for the user to have knowledge of transactions, routing control, streaming of results and cursor lifetimes, and any of the more complex concepts that are exposed when using the session object.

Running a simple write query:
```typescript
const { keys, records, summary } = await driver.executeQuery(
    'CREATE (p:Person{ name: $name }) RETURN p', 
    { name: 'Person1'}
)
```

Running a read query:
```typescript
const { keys, records, summary } = await driver.executeQuery(
   'MATCH (p:Person{ name: $name }) RETURN p',
   { name: 'Person1'},
   { routing: neo4j.routing.READERS}
)
```

Running a read query and transforming the result:
```typescript
const person1 = await driver.executeQuery(
   'MATCH (p:Person{ name: $name }) RETURN p',
   { name: 'Person1'},
   {
     routing: neo4j.routing.READERS,
     resultTransformer: neo4j.resultTransformers.mappedResultTransformer({
       map(record) {
         return record.get('p')
       },
       collect(personArray) {
         return personArray[0]
       }
     })
   }
)
```

⚠️ This API is experimental


bigmontz:
Depends on: https://github.com/neo4j-drivers/testkit/pull/531

packages/core/src/driver.ts
```diff
@@ -193,6 +200,87 @@ class SessionConfig {
...
 }
 
+type RoutingControl = 'WRITERS' | 'READERS'
+const WRITERS: RoutingControl = 'WRITERS'
+const READERS: RoutingControl = 'READERS'
+/**
+ * @typedef {'WRITERS'|'READERS'} RoutingControl
+ */
+/**
+ * Constant that represents writers routing control.
```
robsdedude: Only writers?

packages/core/src/driver.ts
```diff
@@ -193,6 +200,87 @@ class SessionConfig {
...
+     * The username which the user wants to impersonate for the duration of the query.
+     *
+     * @type {string|undefined}
+     */
+    this.impersonatedUser = undefined
+
+    /**
+     * Configure a BookmarkManager for the session to use
+     *
+     * A BookmarkManager is a piece of software responsible for keeping casual consistency between different sessions by sharing bookmarks
```
robsdedude:
```suggestion
     * A BookmarkManager is a piece of software responsible for keeping casual consistency between different pieces of work by sharing bookmarks
```
Or something? Not sure. I'd avoid mentioning sessions as this config option has (at the surface) nothing to do with sessions.

packages/core/src/driver.ts
```diff
@@ -193,6 +200,87 @@ class SessionConfig {
...
+     * between the them.
+     *
+     * By default, it uses the drivers non mutable driver level bookmark manager.
```
robsdedude:
```suggestion
     * By default, it uses the driver's non mutable driver level bookmark manager.
```

robsdedude: Maybe also mention/link the getter function of that BMM here.

packages/core/src/driver.ts
```diff
@@ -250,6 +343,80 @@ class Driver {
...
+   * @returns {BookmarkManager}
+   */
+  get queryBookmarkManager (): BookmarkManager {
+    return this._queryBookmarkManager
+  }
+
+  /**
+   * Executes a query in a retriable context and returns a {@link EagerResult}.
+   *
+   * This method is a shortcut for a transaction function
```
robsdedude:
```suggestion
   * This method is a shortcut for a transaction function.
```

packages/core/src/driver.ts
```diff
@@ -250,6 +343,80 @@ class Driver {
...
+   *
+   * @example
+   * // Run a read query
+   * const { keys, records, summary } = await driver.executeQuery(
+   *                  'MATCH (p:Person{ name: $name }) RETURN p',
+   *                  { name: 'Person1'},
+   *                  { routing: neo4j.routing.READERS})
+   *
+   * @example
+   * // this lines
```
robsdedude:
```suggestion
   * // these lines
```

packages/core/src/driver.ts
```diff
@@ -250,6 +343,80 @@ class Driver {
...
+   *       <PARAMETERS>,
+   *       QueryConfig {
+   *           routing: neo4j.routing.WRITERS,
+   *           resultTransformer: transformer,
+   *           database: "<DATABASE>",
+   *           impersonatedUser: "<USER>",
+   *           bookmarkManager: bookmarkManager
+   *       }
+   *   )
+   * // are equivalent to this ones
```
robsdedude:
```suggestion
   * // are equivalent to those
```

packages/core/src/driver.ts
```diff
@@ -250,6 +343,80 @@ class Driver {
...
+   * This method is a shortcut for a transaction function
+   *
+   * @example
+   * // Run a simple write query
+   * const { keys, records, summary } = await driver.executeQuery('CREATE (p:Person{ name: $name }) RETURN p', { name: 'Person1'})
+   *
+   * @example
+   * // Run a read query
+   * const { keys, records, summary } = await driver.executeQuery(
+   *                  'MATCH (p:Person{ name: $name }) RETURN p',
```
robsdedude: This indentation looks weird to me. Is it on purpose?

packages/core/src/driver.ts
```diff
@@ -250,6 +343,80 @@ class Driver {
...
+   *           routing: neo4j.routing.WRITERS,
+   *           resultTransformer: transformer,
+   *           database: "<DATABASE>",
+   *           impersonatedUser: "<USER>",
+   *           bookmarkManager: bookmarkManager
+   *       }
+   *   )
+   * // are equivalent to this ones
+   * const session = driver.session({
+   *            database: "<DATABASE>",
```
robsdedude: Indentation?

packages/core/src/driver.ts
```diff
@@ -250,6 +343,80 @@ class Driver {
...
+   *    const transformedResult = await session.executeWrite(tx => {
+   *        const result = tx.run("<QUERY>", <PARAMETERS>)
+   *        return transformer(result)
+   *    })
+   * } finally {
+   *    await session.close()
+   * }
+   *
+   * @public
+   * @param {string| {text:string, parameters?: object}} query - Cypher query to execute
```
robsdedude:
```suggestion
   * @param {string | {text: string, parameters?: object}} query - Cypher query to execute
```

packages/core/src/driver.ts
```diff
@@ -250,6 +343,80 @@ class Driver {
...
+   * @param {Object} parameters - Map with parameters to use in query
```
robsdedude:
```suggestion
   * @param {Object} parameters - Map with parameters to use in the query
```

packages/core/src/driver.ts
```diff
@@ -193,6 +200,87 @@ class SessionConfig {
...
+
+Object.freeze(routing)
+
+type ResultTransformer<T> = (result: Result) => Promise<T>
+
+/**
+ * The query configuration
+ * @interface
+ */
+class QueryConfig<Entries extends Dict = Dict, T = EagerResult<Entries>> {
```
robsdedude: I'm curious as this exceeds my TypeScript foo by a lot.

Why is this dance with two type variables needed while `QueryConfig` is only ever instantiated with `Dict, ...`?

Why doesn't this (or something similar) work?

```typescript
class QueryConfig<T = EagerResult<Dict>> { ... }
```

packages/core/src/driver.ts
```diff
@@ -250,6 +343,80 @@ class Driver {
...
+   * }
+   *
+   * @public
+   * @param {string| {text:string, parameters?: object}} query - Cypher query to execute
+   * @param {Object} parameters - Map with parameters to use in query
+   * @param {QueryConfig<T>} config - The query configuration
+   * @returns {Promise<T>}
+   */
+  async executeQuery<T> (query: Query, parameters?: any, config: QueryConfig<Dict, T> = {}): Promise<T> {
+    const bookmarkManager = config.bookmarkManager === null ? undefined : config.bookmarkManager ?? this.queryBookmarkManager
```
robsdedude: I'm not a TS dev, so I don't know how obvious this is to other TS devs, but I had to google the operator precedence here. Both (ternary and nullish) seemed to be low precedence OPs, but which one would be lower... :shrug: So maybe parentheses would help readability.

packages/core/src/driver.ts
```diff
@@ -250,6 +343,80 @@ class Driver {
...
+    const resultTransformer: ResultTransformer<T> = (config.resultTransformer ?? createEagerResultFromResult) as unknown as ResultTransformer<T>
```
robsdedude: > ```typescript
> ... as unknown as ResultTransformer<T>
> ```

WAT? please explain this magic to me.

packages/core/src/internal/query-executor.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  public async execute<T>(config: ExecutionConfig<T>, query: Query, parameters?: any): Promise<T> {
+    const session = this._createSession({
+      database: config.database,
+      bookmarkManager: config.bookmarkManager,
+      impersonatedUser: config.impersonatedUser
+    })
+    try {
+      const execute: TransactionFunction<T> = config.routing === 'READERS'
```
robsdedude: a) I went a step further and raised an exception if users pass an invalid routing option. I don't think it's necessary, just fyi.
b) I think the variable name is unfortunate because it could be confused with `this.execute`.

packages/core/src/result-eager.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+}
+
+/**
+ * Creates a {@link EagerResult} from a given {@link Result} by
+ * consuming all the stream.
+ *
+ * @private
+ * @param {Result} result The result to be consumed
+ * @returns A promise of a EagerResult
```
robsdedude:
```suggestion
 * @returns A promise of an EagerResult
```

packages/core/src/result-eager.ts
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Field summary
+     * @type {ResultSummary}
+     */
+    this.summary = summary
+  }
+}
+
+/**
+ * Creates a {@link EagerResult} from a given {@link Result} by
```
robsdedude:
```suggestion
 * Creates an {@link EagerResult} from a given {@link Result} by
```

packages/core/src/result-eager.ts
```diff
@@ -0,0 +1,74 @@
...
+ * consuming all the stream.
```
robsdedude:
```suggestion
 * consuming the whole stream.
```

packages/core/test/internal/query-executor.test.ts
```diff
@@ -0,0 +1,390 @@
...
+      })
+
+      await expect(queryExecutor.execute(baseConfig, 'query', { a: 'b' })).rejects.toThrow(error)
+
+      expect(sessionsCreated.length).toEqual(1)
+      const [{ spyOnClose }] = sessionsCreated
+      expect(spyOnClose).toHaveBeenCalled()
+    })
+
+    xit('should give precedence to errors during session close', async () => {
```
robsdedude: Why is it being skipped?

packages/core/test/internal/query-executor.test.ts
```diff
@@ -0,0 +1,390 @@
...
+          spy.mockRejectedValue(error)
+        },
+        mockSessionClose (spy) {
+          spy.mockRejectedValue(closeError)
+        }
+      })
+
+      try {
+        await queryExecutor.execute(baseConfig, 'query', { a: 'b' })
+        fail('code should be not reachable')
```
robsdedude:
```suggestion
        fail('code should be unreachable')
```

packages/core/test/internal/query-executor.test.ts
```diff
@@ -0,0 +1,390 @@
...
+      })
+
+      await expect(queryExecutor.execute(baseConfig, 'query', { a: 'b' })).rejects.toThrow(error)
+
+      expect(sessionsCreated.length).toEqual(1)
+      const [{ spyOnClose }] = sessionsCreated
+      expect(spyOnClose).toHaveBeenCalled()
+    })
+
+    xit('should give precedence to errors during session close', async () => {
```
robsdedude: Added skipped test?

packages/core/test/internal/query-executor.test.ts
```diff
@@ -0,0 +1,390 @@
...
+        resultObservers.push(resultObserver)
+        const result = new Result(
+          Promise.resolve(resultObserver),
+          query,
+          parameters
+        )
+        results.push(result)
+        return result
+      }
+    } as ManagedTransaction
```
robsdedude, line 379: I don't understand why you have to tell TS the type of `managedTransaction` twice.

packages/core/test/result-eager.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    const eagerResult = new EagerResult(keys, records, summary)
+
+    expect(eagerResult.keys).toBe(keys)
+    expect(eagerResult.records).toBe(records)
+    expect(eagerResult.summary).toBe(summary)
+  })
+})
+
+describe('createEagerResultFromResult', () => {
+  describe('when a valid result', () => {
```
robsdedude:
```suggestion
  describe('with a valid result', () => {
```
?

robsdedude (approved):

packages/core/src/driver.ts
bigmontz: This dance allows me to do type-safety records:

https://github.com/neo4j/neo4j-javascript-driver/pull/1006/files#diff-0ec3f1aeef14ced21b829fd32b84fb3a078e728e5fe474d34b33912dee2b7a1bR345

bigmontz: It seems like it was not need, the type inference is working fine with the records as we can see in https://github.com/neo4j/neo4j-javascript-driver/pull/1006/files#diff-0ec3f1aeef14ced21b829fd32b84fb3a078e728e5fe474d34b33912dee2b7a1bR345-R380

packages/core/src/driver.ts
bigmontz: Because of the `Dict` thing, the typing was not matching 100%. In this kind scenario, you only can force cast if you cast to unknown first<sup>1</sup>. But it is already fixed.

<sup><sup>1</sup>you can actually cast, but the linter just screams in your face :D </sup>

packages/core/src/internal/query-executor.ts
bigmontz: At this point, I assume the configuration is already valid and the defaults are already set. I could add a validation on the `Driver.executeQuery` though

bigmontz: Done

packages/core/test/internal/query-executor.test.ts
bigmontz: It is because I don't defined the `_run` in the object, which is expected by the `ManagedTransaction`. I solved it by removing the the `: ManagedTransaction` in the definition (which is the preferred way, but it is not possible because of the `_run`).

packages/core/test/internal/query-executor.test.ts
bigmontz: I forget to un-skip

packages/core/test/internal/query-executor.test.ts
bigmontz: I forget to un-skip

packages/core/src/driver.ts
```diff
@@ -207,10 +207,10 @@ const READERS: RoutingControl = 'READERS'
  * @typedef {'WRITERS'|'READERS'} RoutingControl
  */
 /**
- * Constant that represents writers routing control.
+ * Constants that represents routing mode.
```
robsdedude:
```suggestion
 * Constants that represents routing modes.
```

packages/core/test/driver.test.ts
```diff
@@ -441,7 +441,21 @@ describe('Driver', () => {
...
-      function extendsDefaultWith<T = EagerResult<Dict>> (config: QueryConfig<Dict, T>) {
+      it('should validate the routing configuration', async () => {
+        const expectedError = newError('Illegal query routing config: "GO FIGURE"')
+
+        const query = 'Query'
+        const params = {}
+
+        const output = driver?.executeQuery<string>(query, params, {
+          // @ts-expect-error
+          routing: 'GO FIGURE'
```
robsdedude, line 452: Soon :tm: (maybe) :rofl:

robsdedude (approved):
:art:

packages/core/src/driver.ts
```diff
@@ -288,6 +382,106 @@ class Driver {
...
+   * @returns {BookmarkManager}
+   */
+  get queryBookmarkManager (): BookmarkManager {
+    return this._queryBookmarkManager
+  }
+
+  /**
+   * Executes a query in a retriable context and returns a {@link EagerResult}.
+   *
+   * This method is a shortcut for a transaction function.
```
fbiville: maybe document that `CALL {} IN TRANSACTIONS` and the older `... USING PERIODIC COMMIT` queries won't work here.

packages/core/src/driver.ts
```diff
@@ -288,6 +382,106 @@ class Driver {
...
+   *
+   * @example
+   * // Run a read query
+   * const { keys, records, summary } = await driver.executeQuery(
+   *    'MATCH (p:Person{ name: $name }) RETURN p',
+   *    { name: 'Person1'},
+   *    { routing: neo4j.routing.READERS})
+   *
+   * @example
+   * // Run a read query return a Person Node
```
fbiville:
```suggestion
   * // Run a read query returning a Person Node
```

packages/core/src/driver.ts
```diff
@@ -288,6 +382,106 @@ class Driver {
...
+   * const person1 = await driver.executeQuery(
+   *    'MATCH (p:Person{ name: $name }) RETURN p',
+   *    { name: 'Person1'},
+   *    {
+   *      resultTransformer: neo4j.resultTransformers.mappedResultTransformer({
+   *        map(record) {
+   *          return record.get('p')
+   *        },
+   *        collect(personArray) {
+   *          return personArray[0]
```
fbiville: not sure this is the best collect example ^^
maybe index the persons by element ID or something?

packages/core/src/driver.ts
```diff
@@ -288,6 +382,106 @@ class Driver {
...
+   *    const transformedResult = await session.executeWrite(tx => {
+   *        const result = tx.run("<QUERY>", <PARAMETERS>)
+   *        return transformer(result)
+   *    })
+   * } finally {
+   *    await session.close()
+   * }
+   *
+   * @public
+   * @experimental
```
fbiville: will this annotation be enough to tell users this API can change anytime? or is an explicit mention worth it?

packages/core/src/result-eager.ts
```diff
@@ -0,0 +1,59 @@
...
+  constructor (
+    keys: string[],
+    records: Array<Record<Entries>>,
+    summary: ResultSummary
+  ) {
+    /**
+     * Field keys, in the order the fields appear in the records.
+     * @type {string[]}
+     */
+    this.keys = keys
```
fbiville, line 47: is there a reason you explicitly assign them here, versus declaring the constructor params `public` and adding the docs there?

packages/core/src/result-transformers.ts
```diff
@@ -0,0 +1,190 @@
...
+   * by consuming the whole stream.
+   *
+   * This is the default implementation used in {@link Driver#executeQuery}
+   *
+   * @example
+   * // This:
+   * const { keys, records, summary } = await driver.executeQuery('CREATE (p:Person{ name: $name }) RETURN p', { name: 'Person1'}, {
+   *   resultTransformer: neo4j.resultTransformers.eagerResultTransformer()
+   * })
+   * // equivalent to:
```
fbiville:
```suggestion
   * // is equivalent to:
```

packages/core/src/result-transformers.ts
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  /**
+   * Creates a {@link ResultTransformer} which maps the {@link Record} in the result and collects it
+   * along with the {@link ResultSummary} and {@link Result#keys}.
+   *
+   * NOTE: The config object requires map or/and collect to be valid.
+   *
+   * @example
+   * // Mapping the records
```
fbiville, line 81: it would be interesting to explain here what the default collect does

packages/core/src/result-transformers.ts
```diff
@@ -0,0 +1,190 @@
...
+   * @param {function(records:R[], summary:ResultSummary, keys:string[]):T} [config.collect=function(records, summary, keys) { return { records, summary, keys }}] Method called for mapping
+   * the result data to the transformer output.
+   * @returns {ResultTransformer<T>} The result transformer
+   * @see {@link Driver#executeQuery}
+   * @see https://github.com/neo4j/neo4j-javascript-driver/discussions/1052
+   */
+  mappedResultTransformer <
+    R = Record, T = { records: R[], keys: string[], summary: ResultSummary }
+  >(config: { map?: (rec: Record) => R, collect?: (records: R[], summary: ResultSummary, keys: string[]) => T }): ResultTransformer<T> {
+    if (config == null || (config.collect == null && config.map == null)) {
```
fbiville, line 141: so you allow a `null` map (if collect is defined)? what happens then?

fbiville: not sure to understand what `state.records.push(record as unknown as R)` does

packages/core/src/driver.ts
bigmontz: I can add a text not also.

packages/core/src/result-eager.ts
bigmontz: If i declare with the constructor assignment, I can not use esdocs since it is TS formation, not a ES formation. 😢

packages/core/src/result-transformers.ts
bigmontz: > so you allow a null map (if collect is define)? what happens then?

The collect will receive the list of records, summary and keys for creating the final output. 

> not sure to understand what state.records.push(record as unknown as R) does

At that point, the record is typed as Record, however the generics is defining `R` (which is probably Record, but it could be something else if you force it). In theory, we could just do `record as R`. However, it's a good practice enforced by the linter to cast to `unknown` before. 

`unknown` is like `any`, the difference is that it causes type error when an unknown type value is assigned to a type which is not `any | unknown`, but an unknown variable can receive value of any types just like `any`

packages/core/src/result-transformers.ts
bigmontz: The default collector will return:

```
interface {
   records: R[]
   summary: ResultSummary
   keys: string[]
}
```

So, it's quite similar to the eager result. However, the records are already mapped to `R` type defined in the `map` function.

fbiville (approved):

PR #1010, opened 2022-10-25
Add generics and type mapping to Result, Session.run and Transaction.run

bigmontz:
The new generic typing allow mapping the records of these running queries to type-mapped Record.

Given the following Person and Friendship definitions.

```typescript
interface Person {
  age: Integer
  name: string
}

interface Friendship {
  since: Integer
}

interface PersonAndFriendship {
  p: Node<number, Person>
  f: Relationship<number, Friendship>
}
```

The new type-mapping allow safe access the properties of query which return `Record<Person>`

```typescript
const { records } = await session.run<Person>('MATCH (p:Person) RETURN p.name as name, p.age as age')

for (const person of records) {
    const age: Integer = person.get('age')
    const name: string = person.get('name')

    // @ts-expect-error
    const nameInt: Integer = person.get('name')
}
```

The type-mapping can be also extended for `Node` and `Relationship`.

```typescript
const { records } = await session.run<PersonAndFriendship>('MATCH (p:Person)-[f:Friendship]-() RETURN p, f')

for (const r of records) {
    const person = r.get('p')

    const age: Integer = person.properties.age
    const name: string = person.properties.name

    // @ts-expect-error
    const nameInt: Integer = person.properties.name

    // @ts-expect-error
    const err: string = person.properties.err

    const friendship = r.get('f')

    const since: Integer = friendship.properties.since

    // @ts-expect-error
    const sinceString: string = friendship.properties.since

    // @ts-expect-error
    const err2: string = friendship.properties.err
}
```

The usage in combination with `executeRead` and `executeWrite` is also possible.

```typescript
const { records } = await session.executeRead(tx => tx.run<Person>('MATCH (p:Person) RETURN p.name as name, p.age as age'))

for (const person of records) {
    const age: Integer = person.get('age')
    const name: string = person.get('name')

    // @ts-expect-error
    const nameInt: Integer = person.get('name')
}
```

With some async iterator usage:

```typescript 
const personList = await session.executeRead(async (tx) => {
    const result = tx.run<Person>('MATCH (p:Person) RETURN p.name as name, p.age as age')
    const objects: Person[] = []
    // iterate while streaming the objects
    for await (const record of result) {
        objects.push(record.toObject())
    }
    return objects
})

for (const person of personList) {
    const age: Integer = person.age
    const name: string = person.name

    // @ts-expect-error
    const nameInt: Integer = person.name

    // @ts-expect-error
    const nome: string = person.nome
}
```

### DIFF
--- a/packages/core/src/result.ts
+++ b/packages/core/src/result.ts
@@ -65,7 +65,7 @@ interface QueryResult<Entries extends Dict = Dict> {
  * Interface to observe updates on the Result which is being produced.
  *
  */
-interface ResultObserver {
+interface ResultObserver<Entries=Dict> {
   /**
    * Receive the keys present on the record whenever this information is available
    *
@@ -77,7 +77,7 @@ interface ResultObserver {
    * Receive the each record present on the {@link @Result}
    * @param {Record} record The {@link Record} produced
    */
-  onNext?: (record: Record) => void
+  onNext?: (record: Record<Entries>) => void
 
   /**
    * Called when the result is fully received
@@ -391,7 +391,7 @@ class Result<Entries extends Dict = Dict> implements Promise<QueryResult<Entries
    * @param {function(error: {message:string, code:string})} observer.onError - handle errors.
    * @return {void}
    */
-  subscribe (observer: ResultObserver): void {
+  subscribe (observer: ResultObserver<Entries>): void {
     this._subscribe(observer)
       .catch(() => {})
   }
--- a/packages/core/src/result.ts
+++ b/packages/core/src/result.ts
@@ -20,7 +20,7 @@
 /* eslint-disable @typescript-eslint/promise-function-async */
 
 import ResultSummary from './result-summary'
-import Record from './record'
+import Record, { Dict } from './record'
 import { Query, PeekableAsyncIterator } from './types'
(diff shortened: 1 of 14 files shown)


packages/core/src/result.ts
```diff
@@ -56,16 +56,16 @@ const DEFAULT_ON_KEYS = (keys: string[]): void => {}
  * The query result is the combination of the {@link ResultSummary} and
  * the array {@link Record[]} produced by the query
  */
-interface QueryResult {
-  records: Record[]
+interface QueryResult<Entries extends Dict = Dict> {
```
robsdedude: nit: shouldn't it be called `Entry` everywhere?

robsdedude (approved):
:rocket:

packages/core/src/result.ts
bigmontz: I kept the same naming convention already used in the Record

robsdedude: Is this name exposed to the user or is it just internal? I was thinking this is a type and a type is usually singular, like you write `Dict` and not `Dicts`. But I don't mind either way. Just a nit pick here.

bigmontz: It exposed in the auto-complete. I might call it `RecordShape` so.

PR #1014, opened 2022-11-08
Add generic types for Node#labels, Relationship#type and UnboundRelationship#type

bigmontz:
This generic types enable defining a list string like values accepted for an specific node labels list or relationship type.

Node example:

```typescript
type PersonLabels = 'Person' | 'Actor'

// get the person from a non-typed record. If the record is correctly type-mapped
// the Node type can be omitted since it will be inferred.
const person: Node<Integer, PersonProps, PersonLabels> = record.get('person')

// get labels
// the type inferred for the labels will be PersonLabels[]
const labels = person.labels

// @ts-exepct-error
const wrongLabels: 'Car'|'Bicycle'[] = person.labels
```

Relationship/UnboundRelationship example:

```typescript
type ActedInType = 'ACTED_IN'

// get the relationship from a non-typed record. If the record is correctly type mapped
// the Relationship type can be omitted since it will be inferred.
const actedIn: Relationship<Integer, ActedInProps, ActedInType> = record.get('actedIn')

// get type
// the type inferred for the relType will be 'ACTED_IN'
const relType = actedIn.type

// @ts-expect-error
const directed: 'DIRECTED' = actedIn.type
```

### DIFF
--- a/packages/core/src/graph-types.ts
+++ b/packages/core/src/graph-types.ts
@@ -47,9 +47,9 @@ function hasIdentifierProperty (obj: any, property: string): boolean {
 /**
  * Class for Node Type.
  */
-class Node<T extends NumberOrInteger = Integer, P extends Properties = Properties> {
+class Node<T extends NumberOrInteger = Integer, P extends Properties = Properties, Label extends string = string> {
   identity: T
-  labels: string[]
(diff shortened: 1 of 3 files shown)


robsdedude (approved):
:star2:

PR #1015, opened 2022-11-10
Fix Record#get type checking

bigmontz:
This method was not correctly type checking the key argument. Keys are not in the Entries where being accepted without trigger typescript errors.

The error was being cause because the method was not relying in the keys set in the Entries, but the ones came from the constructor and from the field lookup.

The keys came from constructor and field lookup are not meant to be used in the client code, since they are internal.

This way, the error was solving by strict `get` method key type for consider only indexes (number) and Key originated from the Entries.

Example:

```typescript
interface Person {
   age: Integer
   name: string
}

const p: Record<Person> = // get record form somewhere

const age: Integer = p.get('age')
const name: string = p.get('name')

// @ts-expected-error This error was not being point out before
const nonExistingKey = p.get('non-existing-key')
```

### DIFF
--- a/packages/core/src/record.ts
+++ b/packages/core/src/record.ts
@@ -68,7 +68,7 @@ function generateFieldLookup<
 class Record<
   Entries extends Dict = Dict,
   Key extends keyof Entries = keyof Entries,
-  FieldLookup extends Dict<string, number> = Dict<string, number>
+  FieldLookup extends Dict<keyof Entries, number> = Dict<keyof Entries, number>
 > {
   keys: Key[]
   length: number
@@ -198,11 +198,10 @@ class Record<
   get <K extends keyof Entries = keyof Entries> (key: K | number): any {
     let index: number
     if (!(typeof key === 'number')) {
-      // @ts-expect-error
       index = this._fieldLookup[key]
       if (index === undefined) {
         throw newError(
-          `This record has no field with key '${key.toString()}', available key are: [` +
+          `This record has no field with key '${key.toString()}', available keys are: [` +
             this.keys.toString() +
             '].'
         )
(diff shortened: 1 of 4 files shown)


packages/core/src/record.ts
```diff
@@ -187,24 +187,22 @@ class Record<
...
    *
    * @param {string|Number} key Field key, or the index of the field.
    * @returns {*}
    */
-  get (key: string | number): any {
-    let index
+  get <K extends keyof Entries = keyof Entries> (key: K | number): any {
+    let index: number
     if (!(typeof key === 'number')) {
+      // @ts-expect-error
```
robsdedude: I tried to undertand why this is needed.

Sure, I see that `K extends keyof Entries` does not fit `string` which `this._fieldLookup[_]` expects, but what I mean is that I couldn't track down why `keyof Entries` couldn't be typed to be of type `string`. What else could there be in the Record?

packages/core/src/record.ts
```diff
@@ -187,24 +187,22 @@ class Record<
...
       index = this._fieldLookup[key]
       if (index === undefined) {
         throw newError(
-          "This record has no field with key '" +
-            key +
-            "', available key are: [" +
+          `This record has no field with key '${key.toString()}', available key are: [` +
```
robsdedude:
```suggestion
          `This record has no field with key '${key.toString()}', available keys are: [` +
```
+ adjust tests

robsdedude (approved):
:shrimp:

packages/core/src/record.ts
bigmontz: The _fieldLookup type is defined in the constructor and it doesn't deriver from the `Entries`. I can fix it.

PR #1016, opened 2022-11-14
Clear receive timeout when close browser and Deno channels

bigmontz:
The receive timeout was not being clearer when channel get closed.
With the timeout running, receive timeouts events can still be notified to the existing observers.
This was causing some tests failing and possible issues in production code.

This bug was found in the test `package/neo4j-driver/test/result.test.js (should handle missing onCompleted)`. This test was calling `done` twice because of a late failing coming to the observer.

Node channels are not affected by this error since its timeout implementation is controlled to the socket.

### DIFF
--- a/packages/bolt-connection/test/channel/browser/browser-channel.test.js
+++ b/packages/bolt-connection/test/channel/browser/browser-channel.test.js
@@ -412,6 +412,31 @@ describe('WebSocketChannel', () => {
         expect(fakeSetTimeout.clearedTimeouts).toEqual([])
       })
 
+      it('should be cleared when connection closes', async () => {
+        webSocketChannel.startReceiveTimeout()
+
+        expect(fakeSetTimeout._timeoutIdCounter).toEqual(1)
+        expect(fakeSetTimeout.calls.length).toEqual(1)
+        expect(fakeSetTimeout.calls[0][1]).toEqual(receiveTimeout)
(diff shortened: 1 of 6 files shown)


packages/bolt-connection/test/channel/browser/browser-channel.test.js
```diff
@@ -412,6 +412,31 @@ describe('WebSocketChannel', () => {
...
+        expect(fakeSetTimeout.clearedTimeouts).toEqual([])
+
+        await webSocketChannel.close()
+
+        expect(fakeSetTimeout._timeoutIdCounter).toEqual(1)
+        expect(fakeSetTimeout.calls.length).toEqual(1)
+        expect(fakeSetTimeout.clearedTimeouts).toEqual([0])
+      })
+
+      it('should call not setTimeout(receiveTimeout) when connection is closed', async () => {
```
robsdedude, line 430: I tried to understand if this can happen in the driver but I got lost in the callback jungle trying to trace back potential invocations of `startReceiveTimeout` on a closed channel.

robsdedude (approved):
:stopwatch: :hammer:

packages/bolt-connection/test/channel/browser/browser-channel.test.js
bigmontz: I could happen when the `driver.close` happens just after session acquires a connection from the pool for running a query, for instance.

PR #1017, opened 2022-11-22
Update Bookmark manager for no longer tracking per database

bigmontz:
Changes in the BookmarkManagerConfig:

* `initialBookmarks` type changed from `Map<string,Iterable<string>>` to `Iterable<string>`
* `bookmarksSupplier` type changed from `(db?: string) => Promise<Iterable<string>>` to `() => Promise<Iterable<string>>`
* `bookmarksConsumer` type changed from `(db: string, bookmarks: Iterable<string>) => Promise<void>` to `(bookmarks: Iterable<string>) => Promise<void>`

Changes in the BookmarkManager:

* `getAllBookmarks` and `forget` were removed
* `updateBookmarks` signature changed to `updateBookmarks (previousBookmarks: Iterable<string>, newBookmarks: Iterable<string>): Promise<void>`
* `getBookmarks` signature changed to `getBookmarks (): Promise<Iterable<string>>`

⚠️ This is a experimental feature.

### DIFF
--- a/packages/core/src/bookmark-manager.ts
+++ b/packages/core/src/bookmark-manager.ts
@@ -36,54 +36,30 @@ export default class BookmarkManager {
    * Method called when the bookmarks get updated when a transaction finished.
    *
    * This method will be called when auto-commit queries finish and when explicit transactions
-   * get commited.
+   * get committed.
    *
-   * @param {string} database The database which the bookmarks belongs to
    * @param {Iterable<string>} previousBookmarks The bookmarks used when starting the transaction
    * @param {Iterable<string>} newBookmarks The new bookmarks received at the end of the transaction.
    * @returns {void}
   */
-  async updateBookmarks (database: string, previousBookmarks: Iterable<string>, newBookmarks: Iterable<string>): Promise<void> {
+  async updateBookmarks (previousBookmarks: Iterable<string>, newBookmarks: Iterable<string>): Promise<void> {
     throw new Error('Not implemented')
   }
 
   /**
-   * Method called by the driver to get the bookmarks for one specific database
-   *
-   * @param {string} database The database which the bookmarks belong to
-   * @returns {Iterable<string>} The set of bookmarks
-   */
-  async getBookmarks (database: string): Promise<Iterable<string>> {
-    throw new Error('Not implemented')
-  }
-
-  /**
-   * Method called by the driver for getting all the bookmarks.
-   *
-   * This method should return all bookmarks for all databases present in the BookmarkManager.
+   * Method called by the driver to get the bookmarks.
    *
    * @returns {Iterable<string>} The set of bookmarks
    */
-  async getAllBookmarks (): Promise<Iterable<string>> {
-    throw new Error('Not implemented')
-  }
-
-  /**
-   * Forget the databases and its bookmarks
-   *
-   * This method is not called by the driver. Forgetting unused databases is the user's responsibility.
-   *
-   * @param {Iterable<string>} databases The databases which the bookmarks will be removed for.
-   */
-  async forget (databases: Iterable<string>): Promise<void> {
+  async getBookmarks (): Promise<Iterable<string>> {
     throw new Error('Not implemented')
   }
 }
 
 export interface BookmarkManagerConfig {
-  initialBookmarks?: Map<string, Iterable<string>>
-  bookmarksSupplier?: (database?: string) => Promise<Iterable<string>>
-  bookmarksConsumer?: (database: string, bookmarks: Iterable<string>) => Promise<void>
+  initialBookmarks?: Iterable<string>
+  bookmarksSupplier?: () => Promise<Iterable<string>>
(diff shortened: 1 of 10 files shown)


bigmontz:
Depends on: https://github.com/neo4j-drivers/testkit/pull/534

packages/core/test/bookmark-manager.test.ts
```diff
@@ -157,136 +81,62 @@ describe('BookmarkManager', () => {
...
-      await manager.getAllBookmarks()
+      await manager.getBookmarks()
 
       expect(bookmarksSupplier).toBeCalledWith()
     })
   })
 
   describe('updateBookmarks()', () => {
-    it('should remove previous bookmarks and new bookmarks for an existing db', async () => {
+    it('should remove previous bookmarks and new bookmarks', async () => {
```
robsdedude:
```suggestion
    it('should replace previous bookmarks with new bookmarks', async () => {
```

packages/core/src/bookmark-manager.ts
```diff
@@ -36,66 +36,40 @@ export default class BookmarkManager {
...
  * @typedef {Object} BookmarkManagerConfig
  *
  * @since 5.0
  * @experimental
- * @property {Map<string,Iterable<string>>} [initialBookmarks@experimental] Defines the initial set of bookmarks. The key is the database name and the values are the bookmarks.
- * @property {function([database]: string):Promise<Iterable<string>>} [bookmarksSupplier] Called for supplying extra bookmarks to the BookmarkManager
- * 1. supplying bookmarks from the given database when the default BookmarkManager's `.getBookmarks(database)` gets called.
- * 2. supplying all the bookmarks when the default BookmarkManager's `.getAllBookmarks()` gets called
- * @property {function(database: string, bookmarks: Iterable<string>): Promise<void>} [bookmarksConsumer] Called when the set of bookmarks for database get updated
+ * @property {Iterable<string>} [initialBookmarks@experimental] Defines the initial set of bookmarks. The key is the database name and the values are the bookmarks.
```
bigmontz:
```suggestion
 * @property {Iterable<string>} [initialBookmarks] Defines the initial set of bookmarks. The key is the database name and the values are the bookmarks.
```

packages/core/test/bookmark-manager.test.ts
```diff
@@ -30,106 +30,33 @@ describe('BookmarkManager', () => {
...
+    it('should return empty if there are no bookmarks', async () => {
       const manager = bookmarkManager({})
 
-      const bookmarks = await manager.getAllBookmarks()
+      const bookmarks = await manager.getBookmarks()
 
       expect(bookmarks).toBeSortedEqual([])
     })
 
     it('should return enriched bookmarks list with supplied bookmarks', async () => {
```
robsdedude, line 51: Is there a test to check that the supplied bookmarks don't leak into the state of the BMM?

packages/core/src/bookmark-manager.ts
robsdedude:
```suggestion
 * @property {Iterable<string>} [initialBookmarks] Defines the initial set of bookmarks. The key is the database name and the values are the bookmarks.
```

robsdedude (approved):
:book: :exclamation: :briefcase:

packages/core/src/driver.ts
```diff
@@ -159,10 +159,10 @@ class SessionConfig {
      * Enabling it is done by supplying an BookmarkManager implementation instance to this param.
      * A default implementation could be acquired by calling the factory function {@link bookmarkManager}.
      *
-     * **Warning**: Share the same BookmarkManager instance accross all session can have a negative impact
+     * **Warning**: Share the same BookmarkManager instance across all session can have a negative impact
```
thelonelyvulpes:
```suggestion
     * **Warning**: Sharing the same BookmarkManager instance across multiple sessions can have a negative impact
```

packages/core/test/bookmark-manager.test.ts
bigmontz: I just added the test

packages/core/src/bookmark-manager.ts
```diff
@@ -119,69 +91,36 @@ export function bookmarkManager (config: BookmarkManagerConfig = {}): BookmarkMa
...
-    const bookmarks = this._getOrInitializeBookmarks(database)
+  async updateBookmarks (previousBookmarks: Iterable<string>, newBookmarks: Iterable<string>): Promise<void> {
+    const bookmarks = this._bookmarks
     for (const bm of previousBookmarks) {
       bookmarks.delete(bm)
     }
     for (const bm of newBookmarks) {
       bookmarks.add(bm)
     }
     if (typeof this._bookmarksConsumer === 'function') {
```
thelonelyvulpes, line 109: this check could be moved to constructor, if it's not a function raise an error. avoids needing to validate the type each call. though you will still need to validate if it is truthy, so maybe not worth it.

bigmontz: But I think it can be worthy to add some validation in factory. I will do it a next PR.

PR #1023, opened 2022-11-25
Fix RxResult performance issue in Browser

### LINKED ISSUE #1019: Lost performance with neo4j-driver 5.2.0 on the browser
Hello! I'm really sorry because I really don't know how to state the issue so I'll try to give as more detail as possible.

I have a web-project that uses neo4j-driver to connect to a database neo4j database (currently version 4.4.6) and it works flawlessly.
This is the code in question: 

```javascript
 rxSession.run(
            `MATCH (n:${label}) return n order by toLower(n.Nombre)`
        ).records().pipe(
            map( record => record.get('n')),
            concatWith(rxSession.close())
        ).subscribe({
            next: (data) => {
                this[label].items.push(data);
            },
            complete: () => {
                this[label].saving = false;
                this[label].loading = false;
                if(this[label].selected.identity != -1) this.select(label, this[label].items.find(v => v.identity.toInt() == this[label].selected.identity.toInt()));
            },
            error: (error) => this.errors.push(error)
        });
```

With driver version 4.4.6 and database version 4.4.6 it loads 760 nodes in milliseconds.
Now that I updated the driver to the 5.2.0 version, the same code loads 760 nodes in 5 seconds!

bigmontz:
`setImmediate` was slowing down the result consumption since it postpone the execution to the next cycle. Since the cycle in Browser is longer, the issue was more perceptive there.

The remove of `setImmediate` doesn't cause any issue with the back-pressure mechanics.

### DIFF
--- a/packages/neo4j-driver/src/result-rx.js
+++ b/packages/neo4j-driver/src/result-rx.js
@@ -251,7 +251,8 @@ function createFullyControlledSubject (
       } else {
         subject.next(value)
         if (!streamControl.paused) {
-          setImmediate(async () => await pushNextValue(iterator.next()))
+          pushNextValue(iterator.next())
+            .catch(() => {})
         }
       }
     } catch (error) {


AndyHeap-NeoTech (approved):
Looks ok to me, might want someone with more JavaScript experience to look as well.

fbiville (approved):

PR #1037, opened 2022-12-05
Add and Improve type guards for `is` methods

bigmontz:
The following methods were update to include type-guards and improve the generics over when using the type-guards:

* isNode
* isRelationship
* isUnboundRelationship
* isPath
* isPathSegment
* isPoint
* isDuration
* isLocalTime
* isTime
* isDate
* isLocalDateTime
* isDateTime

Usage example without generics:

```typescript
function handleDuration(duration: unknown) {
   if (isDuration(duration)) {
      // from this point i can access the duration properties
      // and methods without issue
      const months: Integer = duration.months
   } else {
      // this is not a duration object
      // @ts-expect-error
      const months = duration.months
   }
}
```

Usage example with generics:

```typescript
function handleDuration(duration: unknown) {
   if (isDuration<number>(duration)) {
      // from this point i can access the duration properties
      // and methods without issue
      const months: number = duration.months
      // @ts-expect-error
      const intMonths: Integer = duration.months
   } else {
      // this is not a duration object
      // @ts-expect-error
      const months = duration.months
   }
}
```

⚠️ Generic types miss-match in runtime will not trigger errors.

### DIFF
--- a/packages/core/src/spatial-types.ts
+++ b/packages/core/src/spatial-types.ts
@@ -93,6 +93,7 @@ Object.defineProperty(Point.prototype, POINT_IDENTIFIER_PROPERTY, {
  * @param {Object} obj the object to test.
  * @return {boolean} `true` if given object is a {@link Point}, `false` otherwise.
  */
-export function isPoint (obj?: any): obj is Point {
-  return obj != null && obj[POINT_IDENTIFIER_PROPERTY] === true
+export function isPoint<T extends NumberOrInteger = Integer> (obj: unknown): obj is Point<T> {
+  const anyObj: any | null | undefined = obj
+  return obj != null && anyObj[POINT_IDENTIFIER_PROPERTY] === true
 }
--- a/packages/core/test/spatial-types.test.ts
+++ b/packages/core/test/spatial-types.test.ts
@@ -94,14 +94,34 @@ describe('isPoint', () => {
     new Point(CARTESIAN_3D_CRS_CODE, 19.24, 100.29, 20.22222),
     new Point(CARTESIAN_3D_CRS_CODE, 19.24, 100.29),
     new Point(0, 19.24, 100.29, 20.22222)
-  ])('isPoint(%s) should be truthy', point =>
+  ])('isPoint(%s) should be truthy', (point: unknown) => {
     expect(isPoint(point)).toBeTruthy()
-  )
+
+    if (isPoint(point)) {
+      const typedPoint: Point = point
+      expect(typedPoint).toEqual(point)
+    } else {
+      // @ts-expect-error
+      const typedPoint: Point = point
+      expect(typedPoint).toEqual(point)
+    }
+  })
 
   test.each([
     { srid: CARTESIAN_3D_CRS_CODE, x: 18.24, y: 13.8, z: 124 },
     { srid: 0, x: 18.24, y: 13.8 },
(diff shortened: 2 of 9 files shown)


packages/core/src/spatial-types.ts
```diff
@@ -93,6 +93,7 @@ Object.defineProperty(Point.prototype, POINT_IDENTIFIER_PROPERTY, {
  * @param {Object} obj the object to test.
  * @return {boolean} `true` if given object is a {@link Point}, `false` otherwise.
  */
-export function isPoint (obj?: any): obj is Point {
-  return obj != null && obj[POINT_IDENTIFIER_PROPERTY] === true
+export function isPoint<T extends NumberOrInteger = Integer> (obj: unknown): obj is Point<T> {
+  const anyObj: any | null | undefined = obj
```
robsdedude, line 97: What's the benefit of this intermittent `unknown` type? Why does `function isPoint<T extends NumberOrInteger = Integer> (obj: any): obj is Point<T>` not work?

packages/core/test/spatial-types.test.ts
```diff
@@ -94,14 +94,34 @@ describe('isPoint', () => {
     new Point(CARTESIAN_3D_CRS_CODE, 19.24, 100.29, 20.22222),
     new Point(CARTESIAN_3D_CRS_CODE, 19.24, 100.29),
     new Point(0, 19.24, 100.29, 20.22222)
-  ])('isPoint(%s) should be truthy', point =>
+  ])('isPoint(%s) should be truthy', (point: unknown) => {
     expect(isPoint(point)).toBeTruthy()
```
robsdedude: Why not checking more strictly for it shouldn't to be `true` (like the other type guards are tested).

bigmontz: I think I bring it from other test, I will make more strict.

packages/core/src/spatial-types.ts
bigmontz: We can attribute objects of all types to `any` and `unknown`. However, `unknown` can only be assign to `any` or `unknown`. Apart of this, the use of `any` in the function signature is code smell in typescript, so it should be avoid. 

For this case in particular, the usage of `any` in the tests means the `const p: Point = obj` will also work in the else, since I can attribute `any` to all the types.

packages/core/test/spatial-types.test.ts
bigmontz:
```suggestion
    expect(isPoint(point)).toBe(true)
```

packages/core/test/spatial-types.test.ts
```diff
@@ -94,14 +94,34 @@ describe('isPoint', () => {
...
+  })
 
   test.each([
     { srid: CARTESIAN_3D_CRS_CODE, x: 18.24, y: 13.8, z: 124 },
     { srid: 0, x: 18.24, y: 13.8 },
     ['srid', CARTESIAN_3D_CRS_CODE, 'x', 18.24, 'y', 12.8, 'z', 124],
     'Point(1, 2, 3, 4)'
-  ])('isPoint(%s) should be falsy', point => expect(isPoint(point)).toBeFalsy())
+  ])('isPoint(%s) should be true', (point: unknown) => {
+    expect(isPoint(point)).toBe(true)
```
robsdedude:
```suggestion
  ])('isPoint(%s) should be false', (point: unknown) => {
    expect(isPoint(point)).toBe(false)
```
no?

robsdedude (approved):
:mag_right:

PR #1041, opened 2022-12-23
Fix exported types

bigmontz:
Types accessed by `neo4j.types.` or `neo4j.` where not being able to perform `instanceof` because of the wrong exporting.


bigmontz:
https://trello.com/c/adogNEbR/4361-fix-exported-types

PR #1045, opened 2023-01-02
Fix Integer primitive type representation

bigmontz:
Given the scenario where:

```javascript
const one = neo4j.int(1)
const two = neo4j.int(2)
```

The following operations was resulting in string concatenation:

```javascript
console.log( one + two ) // '12'
console.log( 1 + one ) // '11'

console.log( one.add(two).toInt() ) // 3, which is correct
console.log( '1' + one ) // '11', correct
```

This is issue is solved by defining the method `Integer.valueOf` which defines the primitive type representation. The primitive representation of `Integer` is set as `BigInt`.

This way:

```javascript
console.log( one + two ) // 3n
console.log( 1 + one ) // thrown an exception since we can not add bigint to number
console.log( 1n + one ) // 2n
console.log( -one ) // -1n
console.log( '1' + one ) // '11', correct
```

### DIFF
--- a/packages/core/test/integer.test.ts
+++ b/packages/core/test/integer.test.ts
@@ -398,6 +398,70 @@ describe('Integer', () => {
     )
   })
 
+  test('Integer should be able to use + operator as bigint', () => {
+    fc.assert(
+      fc.property(
+        fc.bigInt({ max: Integer.MAX_SAFE_VALUE.toBigInt(), min: Integer.MIN_SAFE_VALUE.toBigInt() }),
+        fc.bigInt({ max: Integer.MAX_SAFE_VALUE.toBigInt(), min: Integer.MIN_SAFE_VALUE.toBigInt() }),
+        (num1, num2) =>
+          // @ts-expect-error
+          // eslint-disable-next-line  @typescript-eslint/restrict-plus-operands
+          num1 + int(num2) === num1 + num2 &&
+          // @ts-expect-error
+          // eslint-disable-next-line  @typescript-eslint/restrict-plus-operands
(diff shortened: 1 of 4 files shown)


packages/core/test/integer.test.ts
```diff
@@ -370,6 +370,34 @@ describe('Integer', () => {
...
+    fc.assert(
+      fc.property(
+        fc.bigInt({ max: Integer.MAX_SAFE_VALUE.toBigInt(), min: Integer.MIN_SAFE_VALUE.toBigInt() }),
+        // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
+        num => 'string' + int(num) + 'str' === 'string' + int(num).toString() + 'str'
+      )
+    )
+  })
+
+  test('Integer should be able to used in the string interpolation', () => {
```
fbiville:
```suggestion
  test('Integer should be able to be used in the string interpolation', () => {
```

packages/core/test/integer.test.ts
```diff
@@ -370,6 +370,34 @@ describe('Integer', () => {
...
+  test('Integer.valueOf should be equivalent to the Integer.toBigInt', () => {
+    fc.assert(
+      fc.property(
+        fc.bigInt({ max: Integer.MAX_SAFE_VALUE.toBigInt(), min: Integer.MIN_SAFE_VALUE.toBigInt() }),
+        num => int(num).toBigInt() === int(num).valueOf()
+      )
+    )
+  })
+
+  test('Integer should concatenate with a string', () => {
```
fbiville, line 382: shouldn't you test the addition as well?

fbiville (approved):

Issue #993, opened 2022-09-10
[Feature Request] No ability to refresh authentication token


bmax:
I am unsure whether to call this a bug or a feature request. We're using opencypher with AWS Neptune and want to use the neo4j driver with the bolt protocol to allow us in the future to make an easier switch to neo4j enterprise. We are generating an AWS V4 Signature and passing it in as a basic auth to the driver. 

Unfortunately, we can't figure out a way to gracefully handle the signature expiring (whether it's 5m or 24h). I imagine a lot of things would be involved in handling this gracefully (lost queries, reconnecting, transactions, etc), we would love for the driver to be able to handle the regeneration of the signature for us. 

Thanks in advance!

- Neo4j version: AWS Neptune OpenCypher
- Neo4j Mode: N/A
- Driver version: JS Driver 4.4.7
- Operating system: Linux

### Steps to reproduce
Here is the code that we've used to generate the neo4j driver w/ an AWS V4 Signature:
```typescript
import neo4j, { Driver } from 'neo4j-driver'
import { HttpRequest } from '@aws-sdk/protocol-http'
import { defaultProvider } from '@aws-sdk/credential-provider-node'
import { SignatureV4 } from '@aws-sdk/signature-v4'
const { Sha256 } = require('@aws-crypto/sha256-js')
import { Credentials, LogLevel } from '@aws-sdk/types'
import { STS } from '@aws-sdk/client-sts'

const region = 'us-east-1'
const serviceName = 'neptune-db'
const host = process.env.AWS_NEPTUNE_URL
const port = 8182
const protocol = 'bolt+s'
const hostPort = host + ':' + port
const url = protocol + '://' + hostPort

const log_level = process.env.LOG as Exclude<LogLevel, 'all' | 'log' | 'off'>

async function assume(params): Promise<Credentials> {
  const sts = new STS({ region })
  const result = await sts.assumeRoleWithWebIdentity(params)
  if (!result.Credentials) {
    throw new Error('unable to assume credentials - empty credential object')
  }
  return {
    accessKeyId: String(result.Credentials.AccessKeyId),
    secretAccessKey: String(result.Credentials.SecretAccessKey),
    sessionToken: result.Credentials.SessionToken,
  }
}

async function signedHeader() {
  const req = new HttpRequest({
    method: 'GET',
    protocol: protocol,
    hostname: host,
    port: port,
    headers: {
      Host: hostPort,
    },
  })

  const signer = new SignatureV4({
[...]

PR #1050, opened 2023-01-09
Re-authentication

bigmontz:

This changes introduce two ways of changing the connection credentials in a driver instance, each of them solving a different use case. 

### Token Expiration / Change Credentials for the whole driver instance

This use case is related to the issue #993 in the repository. For solving this, the driver is now able to receive a `AuthTokenManager` in the driver creation. This interface enables the user code provide new auth tokens to the driver and be notified by token expiration failures. 

For simplifying the usage, the driver also provides a default implementation of `AuthTokenManager` which can be created with `neo4j. expirationBasedAuthTokenManager` and receives a function for renewing the auth token as parameters. 

Example:

```typescript
import neo4j, { AuthToken } from 'neo4j-driver'

/**
 * Method called whenever the driver needs to refresh the token.
 *
 * The refresh will happen if the driver is notified by the server
 * about a token expiration or if the `Date.now() > tokenData.expiry`
 *
 * Important, the driver will block all the connections creation until
 * this function resolves the new auth token.
 */
async function fetchAuthTokenFromMyProvider () {
   const bearer: string = await myProvider.getBearerToken()
   const token: AuthToken = neo4j.auth.bearer(bearer)
   const expiration: Date = myProvider.getExpiryDate()  
   return {
      token,
      // if expiration is not provided, 
      // the driver will only fetch a new token when a failure happens
      expiration 
   }
}

const driver = neo4j.driver(
    'neo4j://localhost:7687', 
    neo4j.expirationBasedAuthTokenManager({ 
        tokenProvider: fetchAuthTokenFromMyProvider 
    })
)
```

### User Switching

In this scenario, different credentials can be configured in a session providing a way for change the user context for the session. For using this feature, it needed to check if your server supports session auth by calling `driver.supportsSessionAuth()`.

Example:

```typescript
import neo4j from 'neo4j-driver'


const driver =  neo4j.driver(
    'neo4j://localhost:7687', 
    neo4j.auth.basic('neo4j', 'password')
)


const sessionWithUserB = driver.session({
  database: 'neo4j',
  auth: neo4j.auth.basic('userB', 'userBpassword')
})


try {
  // run some queries as userB
  const result = await sessionWithUserB.executeRead(tx => tx.run('RETURN 1'))
} finally {
  // close the session as usual
  await sessionWithUserB.close()
}
```


packages/bolt-connection/test/bolt/bolt-protocol-v5x1.test.js
```diff
@@ -0,0 +1,1092 @@
+/**
```
ConorNeo: Not for this PR but i wonder if we could have these work off each other somehow, on the server we have Annotations that say run this test on x,y,z versions to stop duplication

packages/bolt-connection/test/bolt/bolt-protocol-v4x0.test.js
```diff
@@ -589,6 +589,15 @@ describe('#unit BoltProtocolV4x0', () => {
       expect(unpacked).toEqual(object)
     })
   })
+
+  describe('Bolt v5.1', () => {
```
ConorNeo: Do we not need the other message (login)

packages/bolt-connection/src/bolt/bolt-protocol-v5x1.js
```diff
@@ -0,0 +1,132 @@
...
+   *
+   * @param {Object} param
+   * @param {Object} param.authToken the authentication token.
+   * @param {function(err: Error)} param.onError the callback to invoke on error.
+   * @param {function()} param.onComplete the callback to invoke on completion.
+   * @param {boolean} param.flush whether to flush the buffered messages.
+   *
+   * @returns {StreamObserver} the stream observer that monitors the corresponding server response.
+   */
+  login ({ authToken, onComplete, onError, flush } = {}) {
```
ConorNeo: The Server Calls this Logon, i don't know what the other drivers do and if it should be logon or login but i suspect either the server should change or the drivers so we all align nicely

packages/bolt-connection/test/bolt/bolt-protocol-v4x0.test.js
bigmontz: I only need to check outside the protocol level if driver supports logoff (which means also it support logon). I might be the case I just called `supportReauth`.

packages/bolt-connection/src/bolt/bolt-protocol-v5x1.js
bigmontz: I've started with login because the driver already uses observer called `LoginObserver`, but it is fair enough keep the message with same name as in the server since it will make easier to people read the logs and follow the code, for example.

packages/bolt-connection/test/bolt/bolt-protocol-v5x1.test.js
bigmontz: I don't bother to have duplication in the test code, it make it easier to address differences and make the tests totally decoupled.

Maybe it worth to add subset of re-usable tests per capability, it can be a interesting re-factory to do.

ConorNeo (approved):

tristanls:
Hi, apologies if this is the wrong place to ask. Thank you, been looking forward to this work. It looks like this missed the 5.7 release. Is it feasible to put out 5.8 release so we can test out this API preview?

ConorNeo:
Hi, yes this is merged into dev and (unless something changes) will be in 5.8

jhanggi:
I know this feature is in preview, so happy to wait. I wanted to try this out since we've had problems in the past with `Neo4jError: LDAP authorization info expired`. We've worked around it by catching that particular error and retrying, but this felt like it might be a cleaner solution.

I believe the typescript types may not all be updated. It seems the second argument to `neo4j.driver` still only accepts AuthToken.

`Argument of type 'AuthTokenManager' is not assignable to parameter of type 'AuthToken'. Type 'AuthTokenManager' is missing the following properties from type 'AuthToken': scheme, credentials`

```
const driver = neo4j.driver(
  'neo4j://localhost:7687',
  neo4j.expirationBasedAuthTokenManager({
    // tokenProvider rather than getAuthData
    tokenProvider: () => {
      return neo4j.basic('username', 'password')
    }),
  }),
);
```

bigmontz:
@jhanggi, thanks for spotting the issue. 

I've updated the type definitions in the PR: https://github.com/neo4j/neo4j-javascript-driver/pull/1089

I've also updated description of this PR with the correct example.

bigmontz:
Fix released in 5.9.0

PR #1051, opened 2023-01-13
Add the ability to Refresh Auth Token Credentials

vkagamlyk:
Add the ability to Refresh Auth Token Credentials

Currently when credentials expire, driver will continue trying to connect with the outdated credentials. There is no ability to update or refresh login credentials (See #993).

This PR introduces the ability to pass an optional `credentialsRefresher` function to an authToken object which will change credentials as they become obsolete.

There are no breaking changes introduced in this PR.

There is a test which validates these modifications by using a simple `credentialsRefresher` not breaks serialization.

Also tested with AWS Neptune, new credentials were correctly used to create a new connection.

Intended use
```
async function createDriver() {
  let authToken = { scheme: "basic", realm: "realm", principal: "username", credentialsRefresher: signedHeader };

  return neo4j.driver(url, authToken, {
      encrypted: "ENCRYPTION_ON",
      trust: "TRUST_SYSTEM_CA_SIGNED_CERTIFICATES",
      maxConnectionPoolSize: 2,
      maxConnectionLifetime: 2000
    }
  );
}
```

I'm a first-time contributor and have just signed the CLA

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ coverage
 *.code-workspace
 /testkit/CAs
 /testkit/CustomCAs
+/.vs
--- a/packages/bolt-connection/src/connection/connection-channel.js
+++ b/packages/bolt-connection/src/connection/connection-channel.js
@@ -175,11 +175,11 @@ export default class ChannelConnection extends Connection {
     const self = this
     // credentialsRefresher is function to refresh credentials
     if (authToken && authToken.credentialsRefresher) {
-        var maybePromise = authToken.credentialsRefresher(authToken)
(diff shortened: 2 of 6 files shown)


bigmontz:
Hi and thank you very much for taking the time to work on this PR.

I am really sorry, but I have to close this PR. We've been working on a more general solution for this and other use-cases. There is no estimate yet as to when it will be released but you can see its current state here: https://github.com/neo4j/neo4j-javascript-driver/pull/1050

This is planned across all official drivers. Again, I'm sorry for this. Consider contacting us before starting to work a feature PR to help avoid such unlucky situations.

PR #1055, opened 2023-01-31
Fix `executeQuery` default return type

bigmontz:
The executeQuery method was returning `Promise<unknown>` when no transformer or typing was provided like in the example bellow:

```typescript
// works
const { records } = await driver.executeQuery<EegerResult>(`query`)

// works
const { records }: EagerResult = await driver.executeQuery(`query`)

// doesn't work, but it is fixed by this change
const { records } = await driver.executeQuery(`query`)
```

Defining a default generic type solves the issue.

### DIFF
--- a/packages/core/test/driver.test.ts
+++ b/packages/core/test/driver.test.ts
@@ -342,6 +342,32 @@ describe('Driver', () => {
         }, query, params)
       })
 
+      it('should be able to destruct the result in records, keys and summary', async () => {
+        const query = 'Query'
+        const params = {}
+        const spiedExecute = jest.spyOn(queryExecutor, 'execute')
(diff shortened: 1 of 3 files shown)


robsdedude (approved):
well typed

packages/core/test/driver.test.ts
```diff
@@ -342,6 +342,32 @@ describe('Driver', () => {
         }, query, params)
       })
 
+      it('should be able to destruct the result in records, keys and summary', async () => {
```
fbiville, line 345:
```suggestion
      it('should be able to destructure the result in records, keys and summary', async () => {
```

fbiville (approved):
eager to see this merged 🥁

PR #1059, opened 2023-02-14
Comply to ExecuteQuery ADR latest revision

fbiville:
Align naming of the default bookmark manager getter with the prescribed one in the latest version of the ADR.

Allow filtering of records by skipping `undefined` results. 
This avoid loading all mapped records into memory when some should be discarded.


bigmontz (approved):
🦄

PR #1065, opened 2023-03-09
Introduce Bolt Protocol 5.1

bigmontz:
This version of the protocol introduces two new messages for manage credentials.

* LOGIN with a given auth token.
* LOGOFF of the current auth token.

The HELLO message is not more responsible for negotiating credentials in this version of the protocol.

### DIFF
--- a/packages/bolt-connection/src/bolt/bolt-protocol-v1.js
+++ b/packages/bolt-connection/src/bolt/bolt-protocol-v1.js
@@ -24,7 +24,7 @@ import {
 // eslint-disable-next-line no-unused-vars
 import { Chunker } from '../channel'
 import { structure, v1 } from '../packstream'
-import RequestMessage from './request-message'
+import RequestMessage, { SIGNATURES } from './request-message'
 import {
   LoginObserver,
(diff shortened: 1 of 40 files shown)


packages/bolt-connection/src/bolt/bolt-protocol-v1.js
```diff
@@ -391,15 +468,15 @@ export default class BoltProtocol {
...
       if (flush) {
         this._chunker.flush()
       }
     }
   }
 
-  isLastMessageLogin () {
-    return this._lastMessageSignature === 0x01
+  isLastMessageLogon () {
+    return this._lastMessageSignature === 0x01 ||
```
ConorNeo: For my own understanding, Could we not use modify request-message.js a little and use the consts there to do the checking to remove these literals?

ConorNeo (approved):

packages/bolt-connection/src/bolt/bolt-protocol-v1.js
bigmontz: Sure, I will do that.

PR #1067, opened 2023-03-20
Improve record shapping mapping in `Driver#executeQuery`

bigmontz:
The changes allows the `RecordShape` to be passed through the `executeQuery` method. This way, removing the need of access the `EagerResult` type for passing the type.

Transforming:

```typescript
const result: EagerResult<Person> = driver.executeQuery("My Query")
const person = result.records.map(record => record.toObject()) // the type infered is Person[]

```

to:

```typescript
const result = driver.executeQuery<Person>("My Query") // the type infered is EagerResult<Person>
const person = result.records.map(record => record.toObject()) // the type infered is Person[]
```

Altough, this a breaking changing on the generic types for `executeQuery`. It isn't possible to define the return from the transformer without defining the record shape. However, this is a more advanced use case where the user wants to define their own result transformer.

### DIFF
--- a/packages/core/src/driver.ts
+++ b/packages/core/src/driver.ts
@@ -47,6 +47,7 @@ import resultTransformers, { ResultTransformer } from './result-transformers'
 import QueryExecutor from './internal/query-executor'
 import { newError } from './error'
 import NotificationFilter from './notification-filter'
+import { Dict } from './record'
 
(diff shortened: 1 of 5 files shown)


bigmontz:
Closing since the API experience will be worse than the existing one.

PR #1069, opened 2023-03-30
Fix integration test `CREATE USER` statement

bigmontz:
The statement was failing in newer database version since minimum password length changed to 8. Adjusting the password in the test fix the issue.

The drop user routine was also fixed since it can hide information about the original failure.

### DIFF
--- a/packages/neo4j-driver/test/rx/summary.test.js
+++ b/packages/neo4j-driver/test/rx/summary.test.js
@@ -16,7 +16,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import neo4j from '../../src'
+import neo4j, { Neo4jError } from '../../src'
 // eslint-disable-next-line no-unused-vars
 import RxSession from '../../src/session-rx'
 // eslint-disable-next-line no-unused-vars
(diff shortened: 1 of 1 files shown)


packages/neo4j-driver/test/rx/summary.test.js
```diff
@@ -344,15 +344,23 @@ describe('#integration-rx summary', () => {
...
       await verifySystemUpdates(
         runnable,
-        "CREATE USER foo SET PASSWORD 'bar'",
+        "CREATE USER foo SET PASSWORD 'barizon1'",
         {},
         1
       )
+    } catch (e) {
+      // the user should not be dropped if their creation fails
+      dropUser = !(e instanceof Neo4jError)
```
robsdedude, line 358: Does `Neo4jError` also cover connectivity errors. Maybe it should be something line 

```js
    try {
      await verifySystemUpdates(
        runnable,
        "CREATE USER foo SET PASSWORD 'barizon1'",
        {},
        1
      )
    } catch (e) {
      // the user should not be dropped if their creation fails
      dropUser = (e instanceof AssertionError) // or whatever error type you expect when the failure does not come from the driver but the tests in `verifySystemUpdates`
      throw e
    } finally {
      if (dropUser) {
        await verifySystemUpdates(runnable, 'DROP USER foo', {}, 1)
      }
    }
```

robsdedude (approved):

packages/neo4j-driver/test/rx/summary.test.js
bigmontz: The `Neo4jError` also covers the connectivity error.

PR #1071, opened 2023-03-30
Add default info for `fetchSize` session config

stefano-ottolenghi:
The current undefined is a bit misleading, as it doesn't give away what the actual value is. I would then assume that no value => fetch all. I'd even advocate for putting 1000 into the docstring there, but this is a decision for @bigmontz


bigmontz (approved):
🔢 

Please, update branch with 5.0 changes and run `npm run build::deno` for sync deno code.

PR #1074, opened 2023-04-04
Enables Deno specific tests in Core

bigmontz:
Core features which are environment dependent will need to be implemented. For testing this features, environment dependent tests need to be setup.

The `deno` tests can be run by `npm run test::deno` and they need to be located under `./packages/core/test/deno`. The testkit pipeline is configured to run these tests only when configured for test `deno`.

### DIFF
--- a/packages/core/jest.config.ts
+++ b/packages/core/jest.config.ts
@@ -152,9 +152,10 @@ export default {
   // ],
 
   // An array of regexp pattern strings that are matched against all test paths, matched tests are skipped
-  // testPathIgnorePatterns: [
-  //   "/node_modules/"
-  // ],
+  testPathIgnorePatterns: [
+    "/node_modules/",
+    "/test/deno/"
+  ],
 
   // The regexp pattern or array of patterns that Jest uses to detect test files
   // testRegex: [],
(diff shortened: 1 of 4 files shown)


packages/core/jest.config.ts
```diff
@@ -152,9 +152,10 @@ export default {
   // ],
 
   // An array of regexp pattern strings that are matched against all test paths, matched tests are skipped
-  // testPathIgnorePatterns: [
-  //   "/node_modules/"
-  // ],
+  testPathIgnorePatterns: [
+    "/node_modules/",
+    "/test/deno/"
```
ConorNeo, line 157: so would want to remove this when i add tests?

bigmontz: No, this is the jest configuration. If you remove it, jest will pick the tests under this folder and the code will explode since `Deno` will not be defined.

ConorNeo: ahhh with you sorry!

packages/core/package.json
```diff
@@ -10,6 +10,7 @@
     "test": "jest",
     "test::watch": "jest --watch",
     "test::unit": "npm run test",
+    "test::deno": "deno test --allow-none ./test/deno/",
```
bigmontz, line 13: It might worth to remove the flag `--allow-none` after you create testing code under `./test/deno/`.

MaxAake:
The idea is good but the changes here are out of date and small, so it'll be easier to start from scratch

PR #1076, opened 2023-04-05
Bolt agent added to hello metadata

ConorNeo:


packages/bolt-connection/src/bolt/bolt-protocol-v3.js
```diff
@@ -78,7 +78,7 @@ export default class BoltProtocol extends BoltProtocolV2 {
     // passing notification filter on this protocol version throws an error
     assertNotificationFilterIsEmpty(notificationFilter, this._onProtocolError, observer)
 
-    this.write(RequestMessage.hello(userAgent, authToken), observer, true)
+    this.write(RequestMessage.hello(userAgent === '' || userAgent == null ? boltAgent : userAgent, authToken), observer, true)
```
bigmontz: Should we treat `''` as null?

packages/bolt-connection/src/bolt/request-message.js
```diff
@@ -191,6 +191,38 @@ export default class RequestMessage {
...
+  /**
+   * Create a new HELLO message.
+   * @param {string} userAgent the user agent.
+   * @param {string} boltAgent the bolt agent.
+   * @param {NotificationFilter} notificationFilter the notification filter configured
+   * @param {Object} routing server side routing, set to routing context to turn on server side routing (> 4.1)
+   * @return {RequestMessage} new HELLO message.
+   */
+  static hello5x3 (userAgent, boltAgent, notificationFilter = null, routing = null) {
+    const metadata = { user_agent: userAgent, bolt_agent: boltAgent }
```
bigmontz: maybe we should omit `user_agent` if it's `null` or `undefined`.

packages/core/src/internal/bolt-agent/browser/BoltAgent.ts
```diff
@@ -0,0 +1,21 @@
+/**
```
bigmontz: the file names are created lowercase and separated by `-`. So, this file should be called `bolt-agent.ts` (or move this implementation to be directly in the `index.ts`)

packages/neo4j-driver-deno/generate.ts
```diff
@@ -129,15 +129,15 @@ async function copyAndTransform(inDir: string, outDir: string) {
...
       // Replace the "node channel" with the "deno channel", since Deno supports different APIs
-      if (inPath.endsWith("channel/index.js")) {
+      // Replace the "node boltAgent" with the "deno boltAgent", since Deno supports different APIs
+      if (inPath.endsWith("channel/index.js") || inPath.endsWith("bolt-agent/index.ts")) {
         contents = contents.replace(
-          `export * from './node/index.js'`,
-          `export * from './deno/index.js'`,
+          `export * from './node/index'`,
+          `export * from './deno/index'`,
         );
```
bigmontz: Did you test the deno driver in testkit? I don't know if the `export` replace will work without the extension.

packages/neo4j-driver-lite/src/index.ts
```diff
@@ -376,7 +379,7 @@ async function hasReachableServer (url: string, config?: Pick<Config, 'logging'>
   }
 }
 
-const USER_AGENT: string = 'neo4j-javascript/' + VERSION
+const BOLT_AGENT = internal.boltAgent.BoltAgent.fromVersion('neo4j-javascript/' + VERSION)
```
bigmontz: Maybe move the constant definition before it usage. And I think this line might be `internal.boltAgent.fromVersion`

packages/neo4j-driver/src/index.js
```diff
@@ -342,7 +343,7 @@ async function hasReachableServer (url, config) {
   }
 }
 
-const USER_AGENT = 'neo4j-javascript/' + VERSION
+const BOLT_AGENT = internal.boltAgent.fromVersion('neo4j-javascript/' + VERSION)
```
bigmontz: Maybe move the definition to done before the usage (i know this was already this way, but I think it worth the small refactory).

bigmontz (commented):
This is looking good so far. 

I think it might worth added tests for old protocols to check if they are using the `boltAgent` as fallback for the case the user is not setting the `userAgent`.

packages/bolt-connection/src/bolt/bolt-protocol-v3.js
ConorNeo: if i dont it causes a testkit test to fail

packages/bolt-connection/src/bolt/request-message.js
ConorNeo: yeah makes sense!

bigmontz (approved):
🥇

PR #1089, opened 2023-05-25
Fix re-auth `neo4j.driver` typescript type

bigmontz:
The type signature was missing inform the second param accepts either AuthToken or AuthTokenManager.


AndyHeap-NeoTech (approved):
Approved

PR #1093, opened 2023-06-08
Add `notificationFilter` to the Config type

bigmontz:
The `notificationFilter` config was not mapped correctly to the `Config` type. So, the driver was not offering the correct auto-complete and type support for this configuration on the `neo4j.driver` factory.

The disconnection between the configuration object and the documentation was the root cause of the issue, since the filter was correctly documented.

For solving this issue, the `Config` documentation was moved to live along with its definition. The interface was promoted to class to make it happen. However, the class constructor can only be used internally and for the api point of view, the `Config` still behaving as an interface.


packages/core/src/types.ts
```diff
@@ -68,13 +75,220 @@ export interface Config {
...
+     *
+     * **Default**: ```60 * 60 * 1000 // 1 hour```
+     *
+     * @type {number|undefined}
+     */
+    this.maxConnectionLifetime = 60 * 60 * 1000 // 1 hour
+
+    /**
+     * The maximum amount of time to wait to acquire a connection from the pool (to either create a new
+     * connection or borrow an existing one.
```
stefano-ottolenghi:
```suggestion
     * connection or borrow an existing one).
```

packages/core/src/types.ts
```diff
@@ -68,13 +75,220 @@ export interface Config {
...
+     *
+     * Pooled connections older than this
+     * threshold will be closed and removed from the pool. Such discarding happens during connection acquisition
+     * so that new session is never backed by an old connection. Setting this option to a low value will cause
+     * a high connection churn and might result in a performance hit. It is recommended to set maximum lifetime
+     * to a slightly smaller value than the one configured in network equipment (load balancer, proxy, firewall,
+     * etc. can also limit maximum connection lifetime). No maximum lifetime limit is imposed by default. Zero
+     * and negative values result in lifetime not being checked.
+     *
+     * **Default**: ```60 * 60 * 1000 // 1 hour```
```
stefano-ottolenghi: This is probably my ignorance, but how does the `//` get parsed? If it stays raw, then maybe `=` is more appropriate?

packages/core/src/types.ts
```diff
@@ -68,13 +75,220 @@ export interface Config {
...
+     * @type {number|undefined}
+     */
+    this.connectionAcquisitionTimeout = 60000 // 1 minute
+
+    /**
+     * Specify the maximum time in milliseconds transactions are allowed to retry via
+     * {@link Session#executeRead} and {@link Session#executeWrite} functions.
+     *
+     * These functions will retry the given unit of work on `ServiceUnavailable`, `SessionExpired` and transient
+     * errors with exponential backoff using initial delay of 1 second.
```
stefano-ottolenghi:
```suggestion
     * errors with exponential backoff using an initial delay of 1 second.
```

packages/core/src/types.ts
```diff
@@ -68,13 +75,220 @@ export interface Config {
...
+
+    /**
+     * Make this driver always return native JavaScript numbers for integer values, instead of the
+     * dedicated {@link Integer} class.
+     *
+     * Values that do not fit in native number bit range will be represented as `Number.NEGATIVE_INFINITY` or `Number.POSITIVE_INFINITY`.
+     *
+     * **Warning:** {@link ResultSummary} It is not always safe to enable this setting when JavaScript applications are not the only ones
+     * interacting with the database. Stored numbers might in such case be not representable by native
+     * `Number` type and thus driver will return lossy values. This might also happen when data was
```
stefano-ottolenghi:
```suggestion
     * `Number` type and thus the driver will return lossy values. This might also happen when data was
```

packages/core/src/types.ts
```diff
@@ -68,13 +75,220 @@ export interface Config {
...
+     */
+    this.disableLosslessIntegers = false
+
+    /**
+     * Make this driver always return native Javascript `BigInt` for integer values,
+     * instead of the dedicated {@link Integer} class or `Number`.
+     *
+     * **Warning:** `BigInt` doesn't implement the method `toJSON`. In manner of serialize it as `json`,
+     * this's needed to add a custom implementation of the `toJSON` on the
+     * `BigInt.prototype`. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#use_within_json.
```
stefano-ottolenghi, line 228: I don't understand the meaning of the sentence starting with `In manner` :thinking: 

Is this correct?

> In you want to serialize a `BigInt` value to json, you need a custom implementation of the method `toJSON` on `BigInt.prototype`.

packages/core/src/types.ts
bigmontz: Something like this. I was trying to keep similar with what we have before. 

Screenshot 2023-06-08 at 16 00 47

packages/core/src/types.ts
```diff
@@ -68,13 +75,220 @@ export interface Config {
...
+     * @type {boolean|undefined}
+     */
+    this.disableLosslessIntegers = false
+
+    /**
+     * Make this driver always return native Javascript `BigInt` for integer values,
+     * instead of the dedicated {@link Integer} class or `Number`.
+     *
+     * **Warning:** `BigInt` doesn't implement the method `toJSON`. In manner of serialize it as `json`,
+     * this's needed to add a custom implementation of the `toJSON` on the
```
bigmontz:
```suggestion
     * **Warning:** `BigInt` doesn't implement the method `toJSON`. To serialize it as `json`,
     * it's needed to add a custom implementation of the `toJSON` on the
```

packages/core/src/types.ts
bigmontz: I add a suggestion in comment above (or bellow, it depends on the where you looking the comments)

packages/core/src/types.ts
bigmontz: We can move to be something like ```60000 (1 minute)``` or ```60000``` (1 minute).

packages/core/src/types.ts
stefano-ottolenghi: This looks good! :)

packages/core/src/types.ts
stefano-ottolenghi: I think it would look nicer, so that `// 1 minute` doesn't seem part of the default. Of course nobody will truly believe so, and anybody who believes so would be better off not writing software, but it still requires an extra second for appropriate parsing to the human eye that we could help avoid. It only makes sense if we change the format for all options though, not only for this one.

ConorNeo (approved):

packages/core/src/types.ts
bigmontz: I had addressed the comment and improve all the default values format.

stefano-ottolenghi (approved):
:superhero:

PR #1103, opened 2023-06-27
Fix Driver.getServerInfo api docs

bigmontz:
The method returns `Promise<ServerInfo>` instead of `Promise<void>`


thelonelyvulpes (approved):

robsdedude (approved):
:writing_hand:

PR #1108, opened 2023-07-05
Handle when a Sub Millisecond Tx Timeout is specified

bigmontz:
If a user sets a timeout with a fractional millisecond value it will be rounded up to the next integer value. e.g

> 0.1 → 1
> 1.4 → 2.0
> 1.8 → 2.0

A warning that a rounding has taken place should be logged under info level.

### DIFF
--- a/packages/core/src/internal/tx-config.ts
+++ b/packages/core/src/internal/tx-config.ts
@@ -66,7 +66,7 @@ const EMPTY_CONFIG = new TxConfig({})
 function extractTimeout (config: any): Integer | null {
   if (util.isObject(config) && config.timeout != null) {
     util.assertNumberOrInteger(config.timeout, 'Transaction timeout')
-    const timeout = int(config.timeout)
+    const timeout = int(config.timeout, { ceilFloat: true })
     if (timeout.isNegative()) {
       throw newError('Transaction timeout should not be negative')
     }
--- a/packages/core/src/internal/tx-config.ts
+++ b/packages/core/src/internal/tx-config.ts
@@ -20,6 +20,7 @@
 import * as util from './util'
 import { newError } from '../error'
 import Integer, { int } from '../integer'
+import { Logger } from './logger'
 
 /**
  * Internal holder of the transaction configuration.
@@ -35,9 +36,9 @@ export class TxConfig {
    * @constructor
    * @param {Object} config the raw configuration object.
    */
-  constructor (config: any) {
+  constructor (config: any, log?: Logger) {
     assertValidConfig(config)
-    this.timeout = extractTimeout(config)
+    this.timeout = extractTimeout(config, log)
     this.metadata = extractMetadata(config)
   }
 
@@ -63,9 +64,13 @@ const EMPTY_CONFIG = new TxConfig({})
 /**
  * @return {Integer|null}
  */
-function extractTimeout (config: any): Integer | null {
+function extractTimeout (config: any, log?: Logger): Integer | null {
   if (util.isObject(config) && config.timeout != null) {
     util.assertNumberOrInteger(config.timeout, 'Transaction timeout')
+    if (isTimeoutFloat(config) && log?.isInfoEnabled() === true) {
+      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+      log?.info(`Transaction timeout expected to be an integer, got: ${config.timeout}. The value will be round up.`)
+    }
     const timeout = int(config.timeout, { ceilFloat: true })
     if (timeout.isNegative()) {
       throw newError('Transaction timeout should not be negative')
@@ -73,6 +78,10 @@ function extractTimeout (config: any): Integer | null {
     return timeout
   }
   return null
+}
+
+function isTimeoutFloat (config: any): boolean {
+  return typeof config.timeout === 'number' && !Number.isInteger(config.timeout)
 }
 
 /**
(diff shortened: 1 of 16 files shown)


packages/core/test/session.test.ts
```diff
@@ -474,6 +476,78 @@ describe('session', () => {
...
+            await session.beginTransaction({ timeout })
+
+            expect(connection.seenBeginTransaction[0][0].txConfig.timeout)
+              .toEqual(int(Math.ceil(timeout)))
+          }
+        )
+      )
+    })
+
+    it('should log a warning for timeout configurations with sub milliseconds', async () => {
```
```suggestion
    it('should log when a timeout is configured with sub milliseconds', async () => {
```
it's not a warning... well sort of, but not on warning level, either way ;)

packages/core/src/internal/tx-config.ts
```diff
@@ -63,10 +64,14 @@ const EMPTY_CONFIG = new TxConfig({})
...
  * @return {Integer|null}
  */
-function extractTimeout (config: any): Integer | null {
+function extractTimeout (config: any, log?: Logger): Integer | null {
   if (util.isObject(config) && config.timeout != null) {
     util.assertNumberOrInteger(config.timeout, 'Transaction timeout')
-    const timeout = int(config.timeout)
+    if (isTimeoutFloat(config) && log?.isInfoEnabled() === true) {
+      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+      log?.info(`Transaction timeout expected to be an integer, got: ${config.timeout}. The value will be round up.`)
```
robsdedude:
```suggestion
      log?.info(`Transaction timeout expected to be an integer, got: ${config.timeout}. The value will be rounded up.`)
```

packages/core/test/session.test.ts
```diff
@@ -474,6 +476,78 @@ describe('session', () => {
...
+              beginTx: false,
+              database: 'neo4j'
+            })
+
+            await session.beginTransaction({ timeout })
+
+            expect(loggerFunction).not.toBeCalledWith(
+              'info',
+              `Transaction timeout expected to be an integer, got: ${timeout}. The value will be round up.`
+            )
```
robsdedude: It seems this could be quire fragile. If you change the log message but not this test, it wouldn't start to fail, I believe.

packages/core/test/driver.test.ts
```diff
@@ -625,6 +625,8 @@ describe('Driver', () => {
       mode: 'WRITE',
       reactive: false,
       impersonatedUser: undefined,
+      // @ts-expect-error
+      log: driver?._log,
```
robsdedude, line 629: Why is this? I don't understand it.

packages/neo4j-driver/test/rx/session.test.js
```diff
@@ -442,6 +442,111 @@ describe('#unit rx-session', () => {
     })
   })
 
+  describe('.run()', () => {
+    it('should redirect run to the underline session', () => {
```
robsdedude:
```suggestion
    it('should redirect run to the underlying session', () => {
```

packages/neo4j-driver/test/rx/session.test.js
```diff
@@ -442,6 +442,111 @@ describe('#unit rx-session', () => {
...
+            async summary () {
+              return summary
+            }
+          }
+        }
+      }
+      const session = new RxSession({ session: _session })
+      const query = 'the query'
+      const params = {}
+      const txConfig = { timeout: 1000 }
```
robsdedude: Does it make sense to use a non-int timeout here to verify it gets passed down unchanged?

packages/neo4j-driver/test/rx/session.test.js
robsdedude, line 1: I don't quite understand the difference in design here: why does `RxSession.run` get passed down to `_session` unchanged but `RxSession._beginTransaction` does not?

packages/core/test/driver.test.ts
bigmontz: `_driver` is not exposed as public api, so I have to tell typescript this is an type error, but this is okay.

packages/core/test/session.test.ts
bigmontz: Yes, but for me this is fine. A change in behaviour might trigger a change in the test code. 

Of course, I will not be that specific in a testkit scenario since the change there will be expensive

packages/neo4j-driver/test/rx/session.test.js
bigmontz: I just want to check if the value is send to `run` since I known Session.run knows how to handle this value

packages/neo4j-driver/test/rx/session.test.js
bigmontz: The session run depends only on the Result, so the rxsession can re-use most of the code. 

`RxSession._beginTransaction` is called by the Rx execute methods and beginTx and call `Session._beginTransaction` which doesn't validate any anything. 

This is not the most beautiful design, but 🤷🏼

packages/core/test/session.test.ts
bigmontz: Got it, this about the `.not`. I will fix it.

packages/neo4j-driver/test/rx/session.test.js
robsdedude: Fair enough. I was just thinking it might be good to test that RxSession.run does not change the timeout even if it has sub-ms information. But it's no big deal either way.

robsdedude (approved):
:stopwatch: :running_shirt_with_sash:

PR #1111, opened 2023-07-10
Remove broken TZ from Javascript Driver unit tests

bigmontz:
The removed timezone  doesn’t work properly in a test scenario since the algorithm used for asserting is not able to do the correct math.

The driver code itself is working properly. This is just the matter of fixing the test to avoid flakyness.


robsdedude (approved):
:boot: :boom:

ConorNeo (approved):

PR #1112, opened 2023-07-10
Testkit: skip tests for prevent commit/rollback after tx termination

bigmontz:
These driver needs to implement the feature and the test need to adjusted with the correct exception.


robsdedude (approved):
:carousel_horse:

PR #1113, opened 2023-07-18
Add error logs for connecting to HTTP server

bigmontz:
The driver already throws an appropriated erro when the server returns HTTP. Add this information to the log helps with future debugging process.

### DIFF
--- a/packages/bolt-connection/test/bolt/index.test.js
+++ b/packages/bolt-connection/test/bolt/index.test.js
@@ -115,7 +115,7 @@ describe('#unit Bolt', () => {
       channel.onmessage(packedHandshakeMessage(httpMagicNumber))
     })
 
-    it('should log error if the server responds with http server', async () => {
+    it('should log error if the server responds with http payload', async () => {
       const { channel, handshakePromise, log } = subject()
       const httpMagicNumber = 1213486160
       const logErrorSpy = jest.spyOn(log, 'error')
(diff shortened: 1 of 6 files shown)


packages/bolt-connection/test/bolt/index.test.js
```diff
@@ -114,6 +114,29 @@ describe('#unit Bolt', () => {
 
       channel.onmessage(packedHandshakeMessage(httpMagicNumber))
     })
+
+    it('should log error if the server responds with http server', async () => {
```
fbiville:
```suggestion
    it('should log error if the server responds with http payload', async () => {
```

packages/bolt-connection/test/bolt/index.test.js
```diff
@@ -114,6 +114,29 @@ describe('#unit Bolt', () => {
...
+      const httpMagicNumber = 1213486160
+      const logErrorSpy = jest.spyOn(log, 'error')
+
+      channel.onmessage(packedHandshakeMessage(httpMagicNumber))
+
+      await expect(handshakePromise).rejects.toThrow()
+      expect(logErrorSpy).toHaveBeenCalledWith('Handshake failed since server responded with HTTP.')
+    })
+
+    it('should not log error if the server responds valid protocol version', async () => {
```
fbiville:
```suggestion
    it('should not log error if the server responds with a valid protocol version', async () => {
```

fbiville (approved):

PR #1117, opened 2023-07-26
Export `RecordShape` type

### LINKED ISSUE #1116: [HELP] Type 'T' does not satisfy the constraint 'Dict'.
I started using Neo4j driver from TypeScript and I have this problem:
`Type 'T' does not satisfy the constraint 'Dict'.` when I try to abstract away session usage within a generic function.

My goal is to have typed queries like here:
https://dev.to/adamcowley/using-typescript-with-neo4j-478c

I tried extending T with Dict but I am not able to see Dict outside of Neo4J modules.

Here is my function:
```
export async function executeCypherQueryOnSession<T>(action: (session: Session) => Promise<QueryResult<T>>
) {
  const session = driver.session();

  try {
    let result = action(session);
    session.close();

    return result;
  } catch (err) {
    throw Error('Error occured in executeCypherQueryOnSession: ${err}');
  }
}
```
I am a complete beginner in typescript / javascript and neo4J.
Can you please help me?

Thank you

Edit:
To clarify this is the usage:

```
return executeCypherQueryOnSession<ActorActedInMovie>(async (session) => {
            return await session.executeRead((tx) => { 
                return tx.run<ActorActedInMovie >(
                    "MATCH (actor: Actor {name: ${name}})-[acts:ACTS]-(movie: Movie) return movie ")
            })
        })
```

If i don't abstract away T and just pass the interface like in the article straight to tx.run<SomeInterface> it is accepted by neo4j API.

bigmontz:
This type is the base type used to define the shape of the Record. Exporting this type enables user to use it for defining custom generic function to work with result and records without have to redefine the shape type.

The type was renamed from `Dict` to `RecordShape` for having a more meaningful name for external consumers.


packages/core/src/result.ts
```diff
@@ -86,7 +86,7 @@ interface ResultObserver<RecordShape extends Dict =Dict> {
   onCompleted?: (summary: ResultSummary) => void
 
   /**
-   * Called when some error occurs during the result proccess or query execution
+   * Called when some error occurs during the result process or query execution
```
injectives: processing maybe?

injectives (approved):

packages/core/src/result.ts
bigmontz:
```suggestion
   * Called when some error occurs during the result processing or query execution
```

PR #1119, opened 2023-08-01
Reduce number of `ROUTE` requests when routing table is stale

bigmontz:
The driver was starting a rediscovery process of each `acquireConnection` causing more load than needed in the cluster. Keeping track of the ongoing requests and use it when possible reduces the load in the process and speeds up the connection acquisition.

PR #1121, opened 2023-08-03
Reduce number of `ROUTE` requests when routing table is stale (#1119)

bigmontz:
The driver was starting a rediscovery process of each acquireConnection causing more load than needed in the cluster. Keeping track of the ongoing requests and use it when possible reduces the load in the process and speeds up the connection acquisition.

Backports #1119


robsdedude (approved):
:camera_flash: :printer:

PR #1122, opened 2023-08-04
Introduce `testcontainers` for integration tests

bigmontz:
`testcontainers` replaces `boltkit` as library to start docker containers.

The container should be able to start when `npm run test`, `npm run test::integration` and `npm run test::browser` are run if methods are not called from testkit.

Running the tests direct from the IDE should be also an supported usecase.

* [x] `npm run test::integration`
* [x] `npm run test`
* [x] `npm run test::browser`
* [x] Don't conflict with testkit
* [x] Start container for run tests from IDE (lite-driver)
* [ ] Start container for run tests from IDE (driver)


packages/neo4j-driver-lite/test/integration/config.ts
```diff
@@ -16,15 +16,50 @@
...
  */
+import Neo4jContainer from './neo4j-container'
 
 const env = process.env
 
 const username = env.TEST_NEO4J_USER ?? 'neo4j'
 const password = env.TEST_NEO4J_PASS ?? 'password'
 const hostname = env.TEST_NEO4J_HOST ?? 'localhost'
 const scheme = env.TEST_NEO4J_SCHEME ?? 'bolt'
+const version = env.TEST_NEO4J_VERSION ?? '5.8'
```
ConorNeo: I wonder if we can make this default latest version somehow.

ConorNeo (approved):

packages/neo4j-driver-lite/test/integration/config.ts
bigmontz: Yes, done by configuring '5' as the default version. So, it will selects the latest 5 version.

PR #1124, opened 2023-08-15
Driver.executeQuery optimization

bigmontz:
Pipeline begin and run when using `executeQuery` to reduce the number of round trips.


StephenCathcart (approved):
Looks good to me! 🎉

PR #1125, opened 2023-08-16
Remove `node-static` dependency from `testkit-backend`

bigmontz:
This library has security vulnerabilities and it's not actively maintained.

Their usage was replaced by a very basic implementation of static file server.

See: https://github.com/neo4j/neo4j-javascript-driver/security/dependabot/134

### DIFF
--- a/packages/testkit-backend/src/infrastructure/http-static.server.js
+++ b/packages/testkit-backend/src/infrastructure/http-static.server.js
@@ -48,6 +48,7 @@ export class HttpStaticServer {
 
       fs.readFile(filePath, 'utf-8', (err, message) => {
         if (err) {
+          console.error(`HttpStaticServer: Error reading file ${filePath}`, err)
           writeError(response, 500, '500 Internal Server Error!')
           return
         }
--- a/packages/testkit-backend/src/infrastructure/http-static.server.js
+++ b/packages/testkit-backend/src/infrastructure/http-static.server.js
@@ -1,0 +1,66 @@
+import fs from 'fs'
+
+const DEFAULT_MIME_TYPES = {
+  html: 'text/html',
+  js: 'text/javascript'
+}
+
+const DEFAULT_MIME_TYPE = 'text/plain'
+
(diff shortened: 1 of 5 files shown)


packages/testkit-backend/src/infrastructure/http-static.server.js
```diff
@@ -0,0 +1,66 @@
...
+      }
+
+      if (!fs.existsSync(filePath)) {
+        writeError(response, 404, '404 Not Found!')
+        return
+      }
+
+      fs.readFile(filePath, 'utf-8', (err, message) => {
+        if (err) {
+          writeError(response, 500, '500 Internal Server Error!')
```
robsdedude, line 52: Maybe wanna log the error internally for easier debugging.

packages/testkit-backend/src/infrastructure/http-static.server.js
```diff
@@ -0,0 +1,66 @@
...
+    this._indexFile = indexFile || DEFAULT_INDEX_FILE
+    this._defaultMimeType = defaultMimeType || DEFAULT_MIME_TYPE
+  }
+
+  serve (request, response) {
+    if (request.method === 'GET') {
+      const file = request.url !== '/' ? request.url : this._indexFile
+      const filePath = `${this._basePath}${file}`
+      const [, extension, ...rest] = file.split('.')
+      if (rest.length > 0) {
```
robsdedude, line 39: A bit simplistic :D doesn't allow valid usage of `..` (staying withing `basePath`) nor files or folders that make the path contain multiple `.` (e.g., `my_file.txt.bak`). But should be good enough for TestKit if the limitations are well understood.

robsdedude (approved):
:wastebasket:

packages/testkit-backend/src/infrastructure/http-static.server.js
bigmontz: Exactly, the idea was just block anyone to move out of the base directory and use a bit of the knowledge of the path content to make the check simple.

PR #1131, opened 2023-08-17
Update `karma` dependencies on `neo4j-driver`

bigmontz:
This PR also fix some tests which are using done/async in a inconsistent way.

This solves: https://github.com/neo4j/neo4j-javascript-driver/security/dependabot/107


packages/neo4j-driver/test/bolt-v3.test.js
```diff
@@ -218,7 +218,8 @@ describe('#integration Bolt V3 API', () => {
       if (
         e.code !== 'Neo.ClientError.Transaction.TransactionTimedOut' &&
         e.code !== 'Neo.TransientError.Transaction.LockClientStopped' &&
-        e.code !== 'Neo.ClientError.Transaction.LockClientStopped'
+        e.code !== 'Neo.ClientError.Transaction.LockClientStopped' &&
+        e.code !== 'Neo.ClientError.Transaction.TransactionTimedOutClientConfiguration'
```
robsdedude, line 222: What does the error code have to do with `karma`?

packages/neo4j-driver/test/internal/connection-channel.test.js
```diff
@@ -96,58 +96,61 @@ describe('#integration ChannelConnection', () => {
...
-        done()
-      })
+  it('should provide error message when connecting to http-port', async () => {
+    let fail = false
+    try {
+      await createConnection(`bolt://${sharedNeo4j.hostnameWithHttpPort}`, {
+        encrypted: false
+      }, null, new Logger('error', () => {}))
+    } catch (error) {
+      fail = true
```
robsdedude: Is there no `expect(promise).toFailWith(...)` or something like it?

packages/neo4j-driver/test/internal/connection-channel.test.js
```diff
@@ -278,31 +273,32 @@ describe('#integration ChannelConnection', () => {
     testQueueingOfObserversWithBrokenConnection(resetAction, done)
   })
 
-  it('should reset and flush when SUCCESS received', async done => {
-    connection = await createConnection(`bolt://${sharedNeo4j.hostnameWithBoltPort}`)
+  it('should reset and flush when SUCCESS received', done => {
+    createConnection(`bolt://${sharedNeo4j.hostnameWithBoltPort}`).then(conn => {
```
robsdedude, line 271: why is this one not `async` and using `await`?

robsdedude (approved):

packages/neo4j-driver/test/bolt-v3.test.js
bigmontz: This one start to happen with lastest version of the server deployed to docker.

packages/neo4j-driver/test/internal/connection-channel.test.js
bigmontz: There is something like this, but we have some extra checks on error which is only done in the `testUtils.isServer() = true`. I can try to quick refactory it.

packages/neo4j-driver/test/internal/connection-channel.test.js
bigmontz: there is no await here. The async/await was removed from this test since we need to work with promises to be able to call `resetAndFlush` and simulate some protocol messages before the resolution of the promise. There are other tests doing something similar. 

The mix of `async/await` with the callback `done` is not supported anymore in newer versions of karma, the is actually good.

packages/neo4j-driver/test/internal/connection-channel.test.js
robsdedude: Looks cleaner :rocket:

PR #1137, opened 2023-09-06
Fix flaky tests on TransactionExecutor suite

bigmontz:
The TransactionExecutor tests were mocking the global setTimeout and clearTimeout. This mocks were conflicting with other calls to this api in the browser tests. So, the calls to the set/clear were not being the expected in some situations.

Replace the global mock for injecting the functions resolve this issue in a less intrusive way.

### DIFF
--- a/packages/core/src/internal/transaction-executor.ts
+++ b/packages/core/src/internal/transaction-executor.ts
@@ -33,19 +33,30 @@ type Resolve<T> = (value: T | PromiseLike<T>) => void
 type Reject = (value: any) => void
 type Timeout = ReturnType<typeof setTimeout>
 
+interface Dependencies {
+  setTimeout: typeof setTimeout
+  clearTimeout: typeof clearTimeout
+}
+
 export class TransactionExecutor {
   private readonly _maxRetryTimeMs: number
   private readonly _initialRetryDelayMs: number
   private readonly _multiplier: number
   private readonly _jitterFactor: number
   private _inFlightTimeoutIds: Timeout[]
+  private readonly _setTimeout: typeof setTimeout
+  private readonly _clearTimeout: typeof clearTimeout
   public pipelineBegin: boolean
 
   constructor (
     maxRetryTimeMs?: number | null,
     initialRetryDelayMs?: number,
     multiplier?: number,
-    jitterFactor?: number
+    jitterFactor?: number,
+    dependencies: Dependencies = {
+      setTimeout,
+      clearTimeout
+    }
   ) {
     this._maxRetryTimeMs = _valueOrDefault(
       maxRetryTimeMs,
@@ -63,6 +74,9 @@ export class TransactionExecutor {
       jitterFactor,
       DEFAULT_RETRY_DELAY_JITTER_FACTOR
     )
+
+    this._setTimeout = dependencies.setTimeout
+    this._clearTimeout = dependencies.clearTimeout
 
     this._inFlightTimeoutIds = []
     this.pipelineBegin = false
@@ -99,7 +113,7 @@ export class TransactionExecutor {
 
   close (): void {
     // cancel all existing timeouts to prevent further retries
-    this._inFlightTimeoutIds.forEach(timeoutId => clearTimeout(timeoutId))
+    this._inFlightTimeoutIds.forEach(timeoutId => this._clearTimeout(timeoutId))
     this._inFlightTimeoutIds = []
   }
 
@@ -119,7 +133,7 @@ export class TransactionExecutor {
 
     return new Promise<T>((resolve, reject) => {
       const nextRetryTime = this._computeDelayWithJitter(retryDelayMs)
-      const timeoutId = setTimeout(() => {
+      const timeoutId = this._setTimeout(() => {
         // filter out this timeoutId when time has come and function is being executed
         this._inFlightTimeoutIds = this._inFlightTimeoutIds.filter(
           id => id !== timeoutId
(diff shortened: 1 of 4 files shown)


packages/core/src/internal/transaction-executor.ts
```diff
@@ -32,20 +32,41 @@ type TransactionWork<T, Tx = Transaction> = (tx: Tx) => T | Promise<T>
...
   public pipelineBegin: boolean
 
   constructor (
     maxRetryTimeMs?: number | null,
     initialRetryDelayMs?: number,
     multiplier?: number,
-    jitterFactor?: number
+    jitterFactor?: number,
+    dependencies: Dependencies = {
+      setTimeout: setTimeoutWrapper,
```
thelonelyvulpes, line 67: why do you need the wrapper function? you could have `global.setTimeout` it is no longer being manipulated in timers-util

packages/neo4j-driver/test/internal/timers-util.js
```diff
@@ -15,47 +15,37 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+*/
+
+/**
+ * This is a liter mock which only creates mocked functions to work
```
thelonelyvulpes:
```suggestion
 * This is a lighter mock which only creates mocked functions to work
```

packages/neo4j-driver/test/internal/timers-util.js
```diff
@@ -15,47 +15,37 @@
...
     }
+    return this._timeoutIdCounter++
+  }
 
-    return this
+  clearTimeout (id) {
+    this.clearedTimeouts.push(id)
   }
 
   pause () {
```
thelonelyvulpes: It took me a while to realise what pause does, pause implies all timeouts can be resumed but the calls to setTimeout will discard their `code` parameter? would `disableTimeoutCallbacks` be more indicative?

thelonelyvulpes (approved):

packages/neo4j-driver/test/internal/timers-util.js
```diff
@@ -15,47 +15,37 @@
...
-
-    this._originalClearTimeout = global.clearTimeout
-    global.clearTimeout = id => {
-      this.clearedTimeouts.push(id)
+  setTimeout (code, delay) {
+    if (!this._paused) {
+      code()
+      this.invocationDelays.push(delay)
     }
+    return this._timeoutIdCounter++
```
robsdedude:
```suggestion
    let timeoutId = this._timeoutIdCounter++
    this.invocationDelays.push(delay)
    if (!this._paused) {
      code()
    }
    return timeoutId
```

I'm not sure whether there's a reason not to do it like this, but this would've been my first intuition: record the call (including the delay) regardless if `code()` throws or weather callbacks are `_paused`.

packages/neo4j-driver/test/internal/transaction-executor.test.js
```diff
@@ -320,178 +315,177 @@ describe('#unit TransactionExecutor', () => {
...
+function createTransactionExecutorWithFakeTimeout (...args) {
+  const fakeSetTimeout = new TimeoutsMock()
+  const dependencies = {
+    setTimeout: fakeSetTimeout.setTimeout,
+    clearTimeout: fakeSetTimeout.clearTimeout
+  }
+
+  // args should have at least 5 elements for place the dependencies
+  if (args.length < 5) {
+    args.length = 5
```
robsdedude: TIL: in JS you can just assign a length to an array  :exploding_head:

But please help me understand what it changes (other than the length attribute). `args[4]` will be `undefined` either way, won't it?

And apparently (tested with node), assigning elements out of bounds is no issue
```javascript
> l = [1, 2]
[ 1, 2 ]
> l[4] = 1
1
> l
[ 1, 2, <2 empty items>, 1 ]

```

packages/neo4j-driver/test/internal/transaction-executor.test.js
```diff
@@ -320,178 +315,177 @@ describe('#unit TransactionExecutor', () => {
...
+
+  // dependencies not set, so we should set
+  if (args[4] === undefined) {
+    args[4] = dependencies
+  // dependencies are an object, so we should join both
+  } else if (typeof args[4] === 'object') {
+    args[4] = { ...dependencies, ...args[4] }
+  } else {
+    throw new TypeError(
+      `Expected  object or undefined as args[4] but got ${typeof args[4]}`)
```
robsdedude:
```suggestion
      `Expected object or undefined as args[4] but got ${typeof args[4]}`)
```

robsdedude (approved):

packages/core/src/internal/transaction-executor.ts
bigmontz: In browser, the `setTimeout` can only be called from the window object, otherwise it raises an `illegal invocation`. 

I can solve it also by do something like `setTimeout: setTimeout.bind(globalThis)`. However, work with the globalThis is always painful.

packages/neo4j-driver/test/internal/timers-util.js
bigmontz: The same happens to me, but I didn't realise a better name until the code was done. I will go with you suggestion.

packages/neo4j-driver/test/internal/transaction-executor.test.js
bigmontz: The other arguments will be undefined. I didn't know I can assign elements out of bound. So, I will do it instead. The code will looks cleaner.

PR #1139, opened 2023-09-07
Hide protocol object behind the Connection

bigmontz:
The connection in the driver is exposing the protocol object, which is not great. Improve this part of the code can be done by make the connection object have methods to do high level requests to the server.

List of calls the core driver does using the protocol, by passing the connection:

* `version`
* `run`
* `beginTransaction`
* `commitTransaction`
* `rollbackTransaction`

Methods is present in the `Connection` interface (used by core in **bold**):

* `id`
* `database`
* `server`
* `authToken`
* `address`
* `version`
* `supportsReAuth`
* **`isOpen`**
* **`protocol`**
* `connect`
* `write`
* **`resetAndFlush`**
* **`hasOngoingObservableRequests`**
* **`_release`**

So, `isOpen`, `resetAndFlush` and `hasOngoingObservableRequests`  are the methods which will stay in the connection along with the new methods. The method `release` will move the a `Releasable` interface, which will be composed with `Connection` when returning the connection from the provider.

The `Releasable` interface is also defined to enable the `ConnectionProvider` returns a connection which can be released back to the pool.

Internally, `bolt-connection` can keep exposing the internal of the connection outside the connection in a first moment. The full encapsulation of the `protocol` should be done in the next phase of refactoring.

### DIFF
--- a/packages/core/src/result.ts
+++ b/packages/core/src/result.ts
@@ -513,7 +513,7 @@ class Result<R extends RecordShape = RecordShape> implements Promise<QueryResult
           connectionHolder
             .releaseConnection()
             .then(() =>
-              connection?.protocol()?.version
+              connection?.getProtocolVersion()
             ),
         // onRejected:
         _ => undefined
--- a/packages/core/src/session.ts
+++ b/packages/core/src/session.ts
@@ -75,7 +75,7 @@ class Session {
   private readonly _results: Result[]
   private readonly _bookmarkManager?: BookmarkManager
   private readonly _notificationFilter?: NotificationFilter
-  private readonly _log?: Logger
+  private readonly _log: Logger
   /**
    * @constructor
    * @protected
@@ -132,7 +132,8 @@ class Session {
       connectionProvider,
       impersonatedUser,
       onDatabaseNameResolved: this._onDatabaseNameResolved,
-      getConnectionAcquistionBookmarks: this._getConnectionAcquistionBookmarks
+      getConnectionAcquistionBookmarks: this._getConnectionAcquistionBookmarks,
+      log
     })
     this._writeConnectionHolder = new ConnectionHolder({
       mode: ACCESS_MODE_WRITE,
@@ -142,7 +143,8 @@ class Session {
       connectionProvider,
       impersonatedUser,
       onDatabaseNameResolved: this._onDatabaseNameResolved,
-      getConnectionAcquistionBookmarks: this._getConnectionAcquistionBookmarks
+      getConnectionAcquistionBookmarks: this._getConnectionAcquistionBookmarks,
+      log
     })
     this._open = true
     this._hasTx = false
--- a/packages/core/src/session.ts
+++ b/packages/core/src/session.ts
@@ -187,7 +187,7 @@ class Session {
     const result = this._run(validatedQuery, params, async connection => {
       const bookmarks = await this._bookmarks()
       this._assertSessionIsOpen()
-      return (connection as Connection).protocol().run(validatedQuery, params, {
+      return (connection as Connection).run(validatedQuery, params, {
         bookmarks,
         txConfig: autoCommitTxConfig,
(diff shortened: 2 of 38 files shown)


packages/core/src/result.ts
```diff
@@ -513,7 +513,7 @@ class Result<R extends RecordShape = RecordShape> implements Promise<QueryResult
           connectionHolder
             .releaseConnection()
             .then(() =>
-              connection?.protocol()?.version
+              connection?.getProtocolVersion()
```
robsdedude, line 516: Was the `?` after `.protocol()` redundant? Because looking into the implementations of `Connection`, this was effectively turned from `connection?.protocol()?.version` to `connection?.protocol().version`.

packages/core/src/session.ts
```diff
@@ -187,7 +187,7 @@ class Session {
     const result = this._run(validatedQuery, params, async connection => {
       const bookmarks = await this._bookmarks()
       this._assertSessionIsOpen()
-      return (connection as Connection).protocol().run(validatedQuery, params, {
+      return (connection as Connection).run(validatedQuery, params, {
```
robsdedude: (Somewhat unrelated:) I tried to understand why this cast is necessary. 
I found the type definition `type ConnectionConsumer = (connection: Connection | null) => any | undefined | Promise<any> | Promise<undefined>` and wondered why connection could be `null` in that type but here in the callback it is assumed/asserted it's not.

packages/core/src/transaction.ts
```diff
@@ -150,13 +150,13 @@ class Transaction {
               if (events != null) {
                 events.onError(error)
               }
-              return this._onError(error)
+              this._onError(error).catch(() => {})
```
robsdedude, line 153: Why swallow the error? Are they handled in a different place or could this potentially make pinpointing a programming error in the driver (specifically in the error handler) quite hard?

packages/core/test/result.test.ts
```diff
@@ -466,14 +466,15 @@ describe('Result', () => {
...
-            const connectionMock = {
-              protocol: () => {
-                return { version }
-              }
-            }
+            const connectionMock = new FakeConnection()
+            // converting to accept number as undefined
+            // this test is considering the situation where protocol version
+            // is undefined, which it should not happen in normal driver
+            // operation.
```
robsdedude:
```suggestion
            // converting to accept undefined as number
            // this test is considering the situation where protocol version
            // is undefined, which should not happen during normal driver
            // operation.
```

packages/core/test/result.test.ts
```diff
@@ -679,14 +680,15 @@ describe('Result', () => {
...
```
robsdedude:
```suggestion
            // converting to accept undefined as number
            // this test is considering the situation where protocol version
            // is undefined, which should not happen during normal driver
            // operation.
```

robsdedude (approved):
As an added bonus, I like how the new structure makes the mocking in tests cleaner as well :raised_hands:

packages/core/src/result.ts
bigmontz: Redundancy. Probably I just kept chain `?` by accident.

packages/core/src/transaction.ts
bigmontz: This error will never happen since `ConnectionHolder.releaseConnection` catch `Connection.release` errors. I've added logs (as warning) to failures on release connection in the `ConnectionHolder`. 

During the investigation, I realise a bug in the `Pool._release` and fixed. The Pool was not updating the resource count and resume pending requests if an error happens while destroying an resource. This behaviour can cause we run out of available connections in the pool if connections start to fail to be closed.

packages/core/src/session.ts
bigmontz: The connection can be null if `ConnectionHolder` is not initialised with success.

PR #1142, opened 2023-09-20
Introduce API Metrics/Telemetry

bigmontz:
Neo4j would like to be able to track more information about driver apis usage, so that smarter decisions can be made on improving drivers and their stack.

The collection of metrics must be respectful of users. Thus, the collected metrics are:
1. impossible to be tied back the customer or user
1. transparent to the users that inspect
1. restrained to required metrics

The metric collection is disabled by default in Neo4j. It can be enabled in the server by setting `server.bolt.telemetry.enabled` to `true`.

However, the metric collection is enabled by default in the drivers. It can be disabled in the driver by configuring the driver with `telemetryDisabled` to `true`. Default: `false`.

**Metrics are only collected when enabled both in server and driver instances.**

Disabling metrics on driver:
```typescript
const driver = neo4j.driver(
    'neo4j://localhost:7687', neo4j.auth.basic('neo4j', 'password'), { telemetryDisabled: true })
```


packages/bolt-connection/src/bolt/bolt-protocol-v1.js
```diff
@@ -454,6 +455,24 @@ export default class BoltProtocol {
     return observer
   }
 
+  /**
+   * Send a TELEMETRY through the underlying connection.
+   *
+   * @param {object} param0 Message params
+   * @param {number} param0.api The API called
+   * @param {object} param1 Configuration and callbacks callbacks
```
robsdedude:
```suggestion
   * @param {object} param1 Configuration and callbacks
```
Isn't is just callbacks, actually?

packages/bolt-connection/src/bolt/stream-observers.js
```diff
@@ -526,6 +526,38 @@ class ResetObserver extends StreamObserver {
...
+    this.onError(
+      newError('Received RECORD when logging off ' + json.stringify(record))
+    )
+  }
+
+  onError (error) {
+    if (this._onError) {
+      this._onError(error)
+    }
+  }
```
robsdedude, line 552: According to the ADR, this message is infallible. If the server replies anything but `SUCCESS`, this is a protocol violation.

packages/bolt-connection/src/bolt/bolt-protocol-v1.js
```diff
@@ -454,6 +455,24 @@ export default class BoltProtocol {
...
+   * @param {function()} param1.onCompleted Called when completed
+   * @param {function()} param1.onError Called when error
+   * @return {StreamObserver} the stream observer that monitors the corresponding server response.
+   */
+  telemetry ({ api }, { onError, onCompleted } = {}) {
+    const observer = new CompletedObserver()
+    if (onCompleted) {
+      onCompleted()
```
robsdedude, line 471: I'm not sure this poses a problem, but I'd suggest not to call `onCompleted`, if the protocol doesn't support telemetry. That way, a retry loop will try to transmit the telemetry data again and might succeed if a rolling upgrade is ongoing.

If this change would cause big issues, it's not a big deal an can be left as is, as telemetry is on a best-effort basis anyway.

packages/bolt-connection/test/bolt/__snapshots__/bolt-protocol-v5x4.test.js.snap
```diff
@@ -0,0 +1,61 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`#unit BoltProtocolV5x4 .packable() should pack not pack graph types (Node) 1`] = `"It is not allowed to pass nodes in query parameters, given: (c:a {a:"b"})"`;
```
robsdedude: > .packable() should pack not pack 

:joy:

packages/core/src/internal/constants.ts
```diff
@@ -37,6 +37,16 @@ const BOLT_PROTOCOL_V5_0: number = 5.0
...
+const BOLT_PROTOCOL_V5_4: number = 5.4
+
+const TELEMETRY_APIS = {
+  MANAGED_TRANSACTION: 0,
+  UNMANAGED_TRANSACTION: 1,
+  AUTO_COMMIT_TRANSACTION: 2,
+  EXECUTE_QUERY: 3
+} as const
+
+export type TelemetryApis = typeof TELEMETRY_APIS[keyof typeof TELEMETRY_APIS]
```
robsdedude, line 49: This type definition is wild :hot_face: 

Do I understand it right that `TelemetryApis` is the union of all `TELEMETRY_APIS` values? So currently `0 | 1 | 2 | 3`?

packages/core/src/transaction-promise.ts
```diff
@@ -69,7 +69,8 @@ class TransactionPromise extends Transaction implements Promise<Transaction> {
     impersonatedUser,
     highRecordWatermark,
     lowRecordWatermark,
-    notificationFilter
+    notificationFilter,
+    apiTelemetryConfig
```
robsdedude, line 74: Parameter not documented

packages/core/src/transaction.ts
```diff
@@ -89,7 +94,8 @@ class Transaction {
     impersonatedUser,
     highRecordWatermark,
     lowRecordWatermark,
-    notificationFilter
+    notificationFilter,
+    apiTelemetryConfig
```
robsdedude, line 99: Undocumented param

packages/core/src/types.ts
```diff
@@ -285,6 +286,16 @@ export class Config {
...
      */
     this.userAgent = undefined
+
+    /**
+     * Specify if telemetry collection is disabled.
+     *
+     * Metrics are only collected when enabled both in server and driver instances.
+     *
+     * **Default**: ```false```
+     * @type {boolean}
```
robsdedude:
```suggestion
     * **Default**: ```false```
     * 
     * @type {boolean}
```
uniform formatting

packages/core/src/types.ts
```diff
@@ -285,6 +286,16 @@ export class Config {
...
+     */
+    this.telemetryDisabled = undefined
```
robsdedude: What's the benefit of it being `undefined` instead of `false`? Especially sinde the `@type` says `boolean` and not `boolean | undefined`?

packages/core/test/session.test.ts
```diff
@@ -87,6 +88,22 @@ describe('session', () => {
     expect(result._watermarks).toEqual({ high: Number.MAX_VALUE, low: Number.MAX_VALUE })
   })
 
+  it('run should set expected apiTelemetryConfig with api equals to TELEMETRY_APIS.AUTO_COMMIT_TRANSACTION', async () => {
```
```suggestion
  it('run should set protocol option apiTelemetryConfig.api to TELEMETRY_APIS.AUTO_COMMIT_TRANSACTION', async () => {
```
or something like it. I don't quite understand the sentence that was there.

packages/core/test/session.test.ts
```diff
@@ -87,6 +88,22 @@ describe('session', () => {
...
+    const result = session.run('RETURN 1')
+    await result
+
+    expect(connection.seenProtocolOptions[0]).toMatchObject({
+      apiTelemetryConfig: {
+        api: TELEMETRY_APIS.AUTO_COMMIT_TRANSACTION
+      }
+    })
+    // @ts-expect-error
+    expect(result._watermarks).toEqual({ high: 700, low: 300 })
```
robsdedude: Nit: I'm surprised to see this assertion here. Doesn't seem to have anything to do with what's being tested.

packages/core/test/session.test.ts
```diff
@@ -689,6 +731,22 @@ describe('session', () => {
...
+      // @ts-expect-error
+      jest.spyOn(Transaction.prototype, 'run').mockImplementation(async () => await Promise.resolve())
+      const query = 'RETURN $a'
+      const params = { a: 1 }
+
+      await execute(session)(async (tx: ManagedTransaction) => {
+        await tx.run(query, params)
+      })
+
+      expect(connection.seenBeginTransaction[0][0].apiTelemetryConfig.api)
```
robsdedude: I've seen 3 different ways of asserting the `apiTelemetryConfig.api` in this file.

 * [`.toMatchObject`](https://github.com/neo4j/neo4j-javascript-driver/pull/1142/files#diff-565952dc4fafc3763a292d42986d3bed115b34365d9e4f7bb7b927a487b61384R98-R102)
  * [`expect.objectContaining`](https://github.com/neo4j/neo4j-javascript-driver/pull/1142/files#diff-565952dc4fafc3763a292d42986d3bed115b34365d9e4f7bb7b927a487b61384R588-R594)
  * `.toEqual` on the property (here)
  
  What are the differences, whey are all these different methods needed? Is there a difference in how permissive/strict they are?

packages/neo4j-driver/test/internal/transaction-executor.test.js
```diff
@@ -314,6 +314,107 @@ describe('#unit TransactionExecutor', () => {
...
+          })
+
+          await executor.execute(transactionCreator, async (x) => 1)
+
+          expect(transactionCreator.calls.length).toBe(2)
+          expect(transactionCreator.calls[0][0].api).toBe(telemetryApi)
+          expect(transactionCreator.calls[1][0]).toBe(undefined)
+        })
+
+        it('should send metric on the retry when metrics sent without success success', async () => {
```
robsdedude:
```suggestion
        it('should send metrics on the retry when metrics sent without success', async () => {
```

packages/core/src/types.ts
```diff
@@ -285,6 +286,16 @@ export class Config {
      * @type {string|undefined}
      */
     this.userAgent = undefined
+
+    /**
+     * Specify if telemetry collection is disabled.
+     *
+     * Metrics are only collected when enabled both in server and driver instances.
```
robsdedude: I suggest to be more explicit in the API docs. What is being collected?

For inspiration https://github.com/neo4j/neo4j-python-driver/pull/965/files#diff-b83b53d2e790a152ae629a3756c2232706f9a9660ed385196b17f63362df2c87R668-R689

packages/bolt-connection/src/bolt/bolt-protocol-v1.js
bigmontz: Just for now, but It might change with configurations like `flush` in the future. I'm just give some tip for myself of the future.

packages/bolt-connection/src/bolt/bolt-protocol-v1.js
bigmontz: I don't think it can cause any harm to the existing code to not call the onComplete, but it is a quite surprising behaviour the request never finished. So, for keeping the same external of the version which actual send the message, I prefer to call the onCompleted

packages/bolt-connection/src/bolt/stream-observers.js
bigmontz: I will transform the erro in a `PROTOCOL_ERROR`

packages/bolt-connection/test/bolt/__snapshots__/bolt-protocol-v5x4.test.js.snap
bigmontz: my dyslexia always causing me problems, that's why I like types.

packages/core/src/internal/constants.ts
bigmontz: Yes, how amazing is it? :D

packages/core/src/types.ts
bigmontz: It doesn't matter much in this case, this type is not exposed to the end user and the value I set here is not used for nothing. This class definition is just a gymnastics to make the esdoc capture the documentation, but the user will never instantiate it with new. The will only use as an interface. 

About the type, actually is `boolean | undefined` since the user don't need to set in the configuration object. However, the driver will treat undefined as false in this case.

packages/bolt-connection/test/bolt/__snapshots__/bolt-protocol-v5x4.test.js.snap
bigmontz: Actually the message is correct since the generated function is called pack.

packages/core/test/session.test.ts
bigmontz: `toMatchObject` matchs a object containing the properties, same with expect.objectContaining, but the second one can be used inside arrays and with `toBeCalledWith`. The third way of doing was my self deviating. I will fix it.

robsdedude (approved):
:tractor:

PR #1150, opened 2023-10-24
Release Re-Auth and AuthTokenManager

bigmontz:
Removes the @experimental tags form the APIs.
The API is consider stable.

### DIFF
--- a/packages/core/src/auth-token-manager.ts
+++ b/packages/core/src/auth-token-manager.ts
@@ -26,7 +26,7 @@ export type SecurityErrorCode = `Neo.ClientError.Security.${string}`
 /**
  * Interface for the piece of software responsible for keeping track of current active {@link AuthToken} across the driver.
  * @interface
- * @since 5.8
+ * @since 5.14
  */
 export default class AuthTokenManager {
   /**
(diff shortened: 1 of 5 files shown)


packages/core/src/auth-token-manager.ts
```diff
@@ -26,7 +26,6 @@ export type SecurityErrorCode = `Neo.ClientError.Security.${string}`
 /**
  * Interface for the piece of software responsible for keeping track of current active {@link AuthToken} across the driver.
  * @interface
- * @experimental Exposed as preview feature.
  * @since 5.8
```
thelonelyvulpes:
```suggestion
 * @since 5.14
```
Should this be when it was GA'd?

thelonelyvulpes (approved):
I just had a question about the use of `@since`, Otherwise, that's a nice clean mechanism in js for marking experimental features, which makes this change very simple.

packages/core/src/auth-token-manager.ts
bigmontz: That makes sense.

bigmontz:
[Domain Object Mapping [Javascript]](https://trello.com/c/Krl7PBEB/304-domain-object-mapping-javascript)

PR #1159, opened 2023-11-08
Introduce Result, Record and Graph types mapping 

bigmontz:
⚠️ This a preview feature

## Getting Started

Let's say we have the following Cypher query:

```cypher
 MATCH (p:Person)-[:ACTED_IN]->(m:Movie)<-[:ACTED_IN]-(c:Person) 
 WHERE id(p) <> id(c) 
 RETURN p AS person, m AS movie, COLLECT(c) AS costars
```

and we are going to load each of the records into Typescript objects like these:

```typescript
class ActingJobs {
    constructor(
        public readonly person: Person,
        public readonly movie: Movie,
        public readonly costars: Person[]
    ) {
    }
}

class  Movie {
    constructor(
        public readonly title: string,
        public readonly released?: number,
        public readonly tagline?: string
    ){
    }
}

class Person {
    constructor (
        public readonly name: string,
        public readonly born?: number
    ) {

    }
}
```

Each record in the results will result in an instance of `ActingJob` with the properties populated from the query results and type validation.

To do this at present, you would write something like the following:

```typescript 
const { records: actingJobsList } = await driver.executeQuery(QUERY, undefined, {
    database: 'neo4j',
    resultTransformer: neo4j.resultTransformers.mappedResultTransformer({ map: (record: Record): ActingJobs | undefined => {
        const person = fromNodeToPerson(record.get('person'))
        const movie = fromNodeToMovie(record.get('movie'))
        const costars = fromNodeListToPersonList(record.get('costars'))

        return new ActingJobs(person, movie, costars)
    } })
})

function fromNodeToPerson (node: any): Person {
    if (isNode(node)) {
        if (typeof node.properties.name !== 'string') {
            throw Error('Person.name is not a string')
        }
        
        if (node.properties.born != null &&  typeof node.properties.born !== 'bigint') {
            throw Error('Person.born is not a number')
        } 
        
        return new Person(node.properties.name, Number(node.properties.born))
    }
    throw Error('Person is not a valid node')
}

function fromNodeToMovie (node: any): Movie {
    if (isNode(node)) {
        if (typeof node.properties.title !== 'string') {
            throw Error('Movie.title is not a string')
        }

        if (node.properties.release != null && typeof node.properties.release !== 'bigint') {
            throw Error('Movie.release is not a string')
        }

        if (node.properties.tagline != null && typeof node.properties.tagline !== 'bigint') {
            throw Error('Movie.tagline is not a string')
        }

        return new Movie(node.properties.title, node.properties.release, node.properties.tagline)        
    }
    throw Error('Movie is not a valid node')
}

function fromNodeListToPersonList (list: any): Person[] {
    if (Array.isArray(list)) {
        return list.map(fromNodeToPerson)
    }
    throw Error('Person list not a valid list.') 
}
```

Using the new mapping functionality, the same result will be achieved using the following code:

```typescript
const personRules: Rules = {
    name: RulesFactories.asString(),
    born: RulesFactories.asNumber({ acceptBigInt: true, optional: true })
}

const movieRules: Rules = {
    title: RulesFactories.asString(),
    release: RulesFactories.asNumber({ acceptBigInt: true, optional: true }),
    tagline: RulesFactories.asString({ optional: true })
}

const actingJobsRules: Rules = {
    person: RulesFactories.asNode({
    }),
    movie: RulesFactories.asNode({
        convert: (node: Node) => node.as(Movie, movieRules)
    }),
    costars: RulesFactories.asList({
        apply: RulesFactories.asNode({
        })
    })
}
const { records: actingJobsList } = await driver.executeQuery(QUERY, undefined, {
    database: 'neo4j',
    resultTransformer: neo4j.resultTransformers.hydratedResultTransformer(ActingJobs, actingJobsRules)
})
```

The mapping will be done by combining rules definition and object properties created by the constructors. It's not necessary to use both (constructor and rules), but the usage of one of them is needed if you need a filled object since properties not present in instantiated object or rules are ignored by the method. 

For example, if values present in the `Movie` and `Person` objects are never null and no validation is needed. The code to process the result can be changed to:

```typescript

const actingJobsRules: Rules = {
    person: RulesFactories.asNode({
    }),
    movie: RulesFactories.asNode({
        convert: (node: Node) => node.as(Movie)
    }),
    costars: RulesFactories.asList({
        apply: RulesFactories.asNode({
        })
    })
}
const { records: actingJobsList } = await driver.executeQuery(QUERY, undefined, {
    database: 'neo4j',
    resultTransformer: neo4j.resultTransformers.hydratedResultTransformer(ActingJobs, actingJobsRules)

```

Another possible scenario is `Movie`, `Person` and `ActingJobs` be just typescript interfaces. The code to process the result can be changed to:

```typescript
const personRules: Rules = {
    name: RulesFactories.asString(),
    born: RulesFactories.asNumber({ acceptBigInt: true, optional: true })
}

const movieRules: Rules = {
    title: RulesFactories.asString(),
    release: RulesFactories.asNumber({ acceptBigInt: true, optional: true }),
    tagline: RulesFactories.asString({ optional: true })
}

const actingJobsRules: Rules = {
    person: RulesFactories.asNode({
        convert: (node: Node) => node.as(personRules)
    }),
    movie: RulesFactories.asNode({
        convert: (node: Node) => node.as(movieRules)
    }),
    costars: RulesFactories.asList({
        apply: RulesFactories.asNode({
            convert: (node: Node) => node.as(personRules)
        })
    })
}
const { records: actingJobsList } = await driver.executeQuery(QUERY, undefined, {
    database: 'neo4j',
    resultTransformer: neo4j.resultTransformers.hydratedResultTransformer<ActingJobs>(actingJobsRules)
})
````

_Note: In this scenario, the `ActingJobs` interface is set to the transformer for auto-complete and code validation, however this is not treated as an object constructor._

## Mapping from `Result` method

`Result.as` is introduced to the result for enable the mapping occur in the result.
Same rules of usage of constructor and rules are share with the `hydratedResultTransformer` transformer`.

```typescript 
const { records: actingJobsList } = await session.executeWrite(tx => tx.run().as(ActingJobs, actingJobsRules))
```

_Note: The current implementation transforms the Result in a promise. However, this should be update to transform the Result in a MappedResult with the hydrated objects being exposed instead of records. So, it will be possible to async iterate, subscribe and call other methods also._

## Direct mapping from an `Record`, `Node` and `Relationship` instance.

`Record.as`, `Node.as` and `Relationship.as` are introduced to enable the mapping of these types only.
Same rules of usage of constructor and rules are share with the `hydratedResultTransformer` transformer`.

```typescript
const actingJobs = record.as(ActingJobs, actingJobsRules) // example with construtor and rules
const person = node.as<Person>(personRules) // example using interfaces
const actedIn = relationship.as(ActedIn) // example without rules
```

_Note: In `Node` and `Relationship` only properties are mapped._
_Note: Properties not present in the `Record` will throw error since `Record.get` is used under the hood. However, properties which present and with value equals to `null` or `undefined` should only throw error when property is not defined as optional in the rules.
 
## Rules and Built-in Rules Factory

The driver provides RuleFactories for all data types which can be returned form a query.
There is no need for creating custom rules in most of cases. 
However, all the built-in rules are extensible.

```typescript
const personRules: Rules = {
    name: RulesFactories.asString({ 
       convert: name => `Name: $name` // change the convert method in the string rule
    }),
    born: RulesFactories.asNumber({ acceptBigInt: true, optional: true })
}
```

The Rule interface is defined as:

```typescript
interface Rule {
  optional?: boolean
  from?: string
  convert?: (recordValue: any, field: string) => any
  validate?: (recordValue: any, field: string) => void
}
```

where 

* `optional` (default: false): indicates with value accept undefined/null as value. 
In this case, the `convert` method will not be called.
* `from` (default: property name on the Rules object): indicates from which field on the Record, Node or Relationship the value will came from.
Used for cases where the property name in domain object is different to the one present in the database.
* `convert`: called to convert value to the domain value. 
Called after value be validated.
If optional is true and the value is null or undefined, the method is not called.
* `validate`: called to validate if values is valid.
Called before convert and after check if the value is optional. 
If optional is true and the value is null or undefined, the method is not called. 

⚠️ This a preview feature

### DIFF
--- a/packages/core/src/graph-types.ts
+++ b/packages/core/src/graph-types.ts
@@ -18,6 +18,7 @@
  */
 import Integer from './integer'
 import { stringify } from './json'
+import { Rules, GenericConstructor, as } from './mapping.highlevel'
 
 type StandardDate = Date
 /**
@@ -82,6 +83,15 @@ class Node<T extends NumberOrInteger = Integer, P extends Properties = Propertie
      * @type {string}
      */
     this.elementId = _valueOrGetDefault(elementId, () => identity.toString())
+  }
+
+  as <T extends {} = Object>(rules: Rules): T
+  as <T extends {} = Object>(genericConstructor: GenericConstructor<T>): T
+  as <T extends {} = Object>(genericConstructor: GenericConstructor<T>, rules?: Rules): T
+  as <T extends {} = Object>(constructorOrRules: GenericConstructor<T> | Rules, rules?: Rules): T {
+    return as({
+      get: (key) => this.properties[key]
+    }, constructorOrRules, rules)
   }
 
   /**
@@ -199,6 +209,15 @@ class Relationship<T extends NumberOrInteger = Integer, P extends Properties = P
      * @type {string}
      */
     this.endNodeElementId = _valueOrGetDefault(endNodeElementId, () => end.toString())
+  }
+
+  as <T extends {} = Object>(rules: Rules): T
+  as <T extends {} = Object>(genericConstructor: GenericConstructor<T>): T
+  as <T extends {} = Object>(genericConstructor: GenericConstructor<T>, rules?: Rules): T
+  as <T extends {} = Object>(constructorOrRules: GenericConstructor<T> | Rules, rules?: Rules): T {
+    return as({
+      get: (key) => this.properties[key]
+    }, constructorOrRules, rules)
   }
 
   /**
@@ -320,6 +339,15 @@ class UnboundRelationship<T extends NumberOrInteger = Integer, P extends Propert
       start.elementId,
       end.elementId
     )
+  }
+
+  as <T extends {} = Object>(rules: Rules): T
+  as <T extends {} = Object>(genericConstructor: GenericConstructor<T>): T
+  as <T extends {} = Object>(genericConstructor: GenericConstructor<T>, rules?: Rules): T
+  as <T extends {} = Object>(constructorOrRules: GenericConstructor<T> | Rules, rules?: Rules): T {
+    return as({
+      get: (key) => this.properties[key]
+    }, constructorOrRules, rules)
   }
 
   /**
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -94,6 +94,8 @@ import * as types from './types'
 import * as json from './json'
 import resultTransformers, { ResultTransformer } from './result-transformers'
 import * as internal from './internal' // todo: removed afterwards
+import { Rule, Rules } from './mapping.highlevel'
+import { RulesFactories } from './mapping.rulesfactories'
 
 /**
  * Object containing string constants representing predefined {@link Neo4jError} codes.
@@ -171,7 +173,8 @@ const forExport = {
   notificationCategory,
   notificationSeverityLevel,
   notificationFilterDisabledCategory,
-  notificationFilterMinimumSeverityLevel
+  notificationFilterMinimumSeverityLevel,
+  RulesFactories
 }
 
 export {
@@ -240,7 +243,8 @@ export {
   notificationCategory,
   notificationSeverityLevel,
   notificationFilterDisabledCategory,
-  notificationFilterMinimumSeverityLevel
+  notificationFilterMinimumSeverityLevel,
+  RulesFactories
 }
 
 export type {
@@ -265,7 +269,9 @@ export type {
   NotificationSeverityLevel,
   NotificationFilter,
   NotificationFilterDisabledCategory,
-  NotificationFilterMinimumSeverityLevel
+  NotificationFilterMinimumSeverityLevel,
+  Rule,
+  Rules
 }
 
 export default forExport
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -94,7 +94,7 @@ import * as types from './types'
 import * as json from './json'
 import resultTransformers, { ResultTransformer } from './result-transformers'
 import * as internal from './internal' // todo: removed afterwards
-import { Rule, Rules } from './mapping.highlevel'
+import { Rule, Rules, mapping } from './mapping.highlevel'
 import { RulesFactories } from './mapping.rulesfactories'
 
 /**
@@ -174,7 +174,8 @@ const forExport = {
   notificationSeverityLevel,
   notificationFilterDisabledCategory,
   notificationFilterMinimumSeverityLevel,
-  RulesFactories
+  RulesFactories,
+  mapping
 }
 
 export {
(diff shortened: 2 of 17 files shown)


AndyHeap-NeoTech:
Is there a default hydrator that doesn't require the user to specify the rules?

bigmontz:
> Is there a default hydrator that doesn't require the user to specify the rules?

Yes, but if the user is returning graph types is not that useful, unless they are okay with working with graph types in their domain. 

An example  of usage is the query:

```cypher
MATCH (p:Person)
RETURN p.name AS name, p.born AS born
```

The user can do:

```typescript
const { records: personList } = await driver.executeQuery(QUERY, undefined, {
    database: 'neo4j',
    resultTransformer: neo4j.resultTransformers.hydratedResultTransformer(Person)
})
```

However, in this scenario the user will not get validation about types, since the type information can not be extracted in runtime. This information is lost when the code is converted to javascript. 

A solution for not having to write the rules it to have a annotation framework and it can be https://github.com/adam-cowley/neode-ogm or some solution using graphql to define the structures and then some function to process and generate the rules for the driver use. 

This can be done in light-ogm on top of the driver, because it's maybe a bit too much for the driver get typescript specific annotations.

bigmontz:
Alternatively, the javascript driver might copy a strategy come from the [.net driver](https://github.com/neo4j/neo4j-dotnet-driver/pull/729). Create a register to associate rules with types, this way you can register once and reuse everywhere. Maybe also selecting differently by `Record` or `Node`, since the information can be returned slightly different in some cases.

bigmontz:
About a global registry for rules, I've got something like:

```typescript
// Defining the rules
const personRules: Rules = {
    name: RulesFactories.asString(),
    born: RulesFactories.asNumber({ acceptBigInt: true, optional: true })
}


const movieRules: Rules = {
    title: RulesFactories.asString(),
    release: RulesFactories.asNumber({ acceptBigInt: true, optional: true }),
    tagline: RulesFactories.asString({ optional: true })
}

const actingJobsRules: Rules = {
    person: RulesFactories.asNode({
    }),
    movie: RulesFactories.asNode({
        convert: (node: Node) => node.as(Movie)
    }),
    costars: RulesFactories.asList({
        apply: RulesFactories.asNode({
        })
    })
}

// registering the rules
neo4j.mapping.register(Person, personRules)
neo4j.mapping.register(Movie, movieRules)
neo4j.mapping.register(ActingJobs, actingJobsRules)
```

Running the query:

```typescript
const { records: actingJobsList } = await driver.executeQuery(QUERY, undefined, {
    database: 'neo4j',
    resultTransformer: neo4j.resultTransformers.hydratedResultTransformer(ActingJobs)
})
```

I didn't implemented any segmentation by driver type (Record, Node, Relationship). However, it's totally possible. The registered rule is used if none is set. 

The registry is global to the whole memory, it's not attached to the driver instance.

bigmontz:
[Domain Object Mapping [Javascript]](https://trello.com/c/Krl7PBEB/304-domain-object-mapping-javascript)

MaxAake:
Superseded by https://github.com/neo4j/neo4j-javascript-driver/pull/1248

PR #1160, opened 2023-11-22
Introduce `transactionConfig` to `Driver.executeQuery`

bigmontz:
Transaction config enables the user to configure timeouts and set metadata for the transaction.

This configuration is already present in APIs such as `Session.executeRead` and `Session.executeWrite`.

Usage example:

```typescript
const { records } = await driver.executeQuery(myQuery, myParams, {
    database: 'neo4j',
    transactionConfig: {
        timeout: 3000,
        metadata: { key: 'value' }
    }
})
```

### DIFF
--- a/packages/core/src/driver.ts
+++ b/packages/core/src/driver.ts
@@ -403,7 +403,7 @@ class QueryConfig<T = EagerResult> {
      * By default, it uses the driver's non mutable driver level bookmark manager. See, {@link Driver.executeQueryBookmarkManager}
      *
      * Can be set to null to disable causal chaining.
-     * @type {BookmarkManager|null}
+     * @type {BookmarkManager|undefined|null}
      */
     this.bookmarkManager = undefined
 
--- a/packages/core/src/driver.ts
+++ b/packages/core/src/driver.ts
@@ -30,7 +30,7 @@ import {
   DEFAULT_POOL_MAX_SIZE
 } from './internal/constants'
 import { Logger } from './internal/logger'
-import Session from './session'
+import Session, { TransactionConfig } from './session'
 import { ServerInfo } from './result-summary'
 import { ENCRYPTION_ON } from './internal/util'
 import {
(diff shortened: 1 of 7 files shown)


packages/core/src/driver.ts
```diff
@@ -405,6 +406,14 @@ class QueryConfig<T = EagerResult> {
...
      */
     this.bookmarkManager = undefined
+
+    /**
+     * Configuration for all transactions started to execute the query.
+     *
+     * @type {TransactionConfig|undefined}
+     *
+     */
+    this.transactionConfig = undefined
```
robsdedude: TS noob question incoming: are `XYZ|null` and `XYZ|undefined` interchangeable union types? It seems to be mixed here.

robsdedude (approved):
:ribbon:

packages/core/src/driver.ts
bigmontz: It can be null actually, but I'm setting undefined since it means the configuration is not set. 

In the case of the `bookmarkManager`, configuring it with `null` means disabling, which is different of configuring it with `undefined` or omitting the configuration.

robsdedude: So if it makes a difference, should `bookmarkManager` actually be typed `{BookmarkManager|null|undefined}`?

bigmontz: `?`enables the omit the configuration or configure with undefined. In the of `bookmarkManager`, we have defined as `?`and `BookmarkManager | null`. 

The null is not be accepted in typescript for `transactionConfig`, but in javascript, it will get the same behaviour as `undefined` since it's kind of expected.

robsdedude: I now understand how the function/constructor types are correct. But what about the comments? `@type {BookmarkManager|null}` vs `@type {TransactionConfig|undefined}`

bigmontz: yes, we should type it as `BookmarkManager|null|undefined`.

bigmontz: I will fix it.

packages/testkit-backend/src/request-handlers.js
```diff
@@ -696,6 +696,13 @@ export function ExecuteQuery ({ neo4j }, context, { driverId, cypher, params, co
         configuration.bookmarkManager = null
       }
     }
+
+    if ('txMeta' in config || 'timeout' in config) {
```
robsdedude: Currently, those properties will always be present, but might be `null`.

PR #1167, opened 2023-12-21
Fix connection getting timeout while idle

bigmontz:
When the connection goes idle, an observer is appended. And this causes the receive timeout to start to count. This behaviour causes connections timing out while being idle in the pool.

For fixing this, the connection received methods to handle its status changing to idle and back from idle. When the connection goes idle, it should stop the timeouts and do not start any timeout until it gets busy again. 

This change also impacts the `hasOngoingRequests` method, since idle connections don't have ongoing requests.

### DIFF
--- a/packages/bolt-connection/src/channel/browser/browser-channel.js
+++ b/packages/bolt-connection/src/channel/browser/browser-channel.js
@@ -54,6 +54,7 @@ export default class WebSocketChannel {
     this._receiveTimeout = null
     this._receiveTimeoutStarted = false
     this._receiveTimeoutId = null
+    this._closingPromise = null
 
     const { scheme, error } = determineWebSocketScheme(config, protocolSupplier)
     if (error) {
@@ -163,17 +164,23 @@ export default class WebSocketChannel {
    * @returns {Promise} A promise that will be resolved after channel is closed
    */
   close () {
-    return new Promise((resolve, reject) => {
-      this._clearConnectionTimeout()
-      if (this._ws && this._ws.readyState !== WS_CLOSED) {
-        this._open = false
-        this.stopReceiveTimeout()
-        this._ws.onclose = () => resolve()
-        this._ws.close()
-      } else {
-        resolve()
-      }
-    })
+    if (this._closingPromise === null) {
+      this._closingPromise = new Promise((resolve, reject) => {
+        this._clearConnectionTimeout()
+        if (this._ws && this._ws.readyState !== WS_CLOSED) {
+          this._open = false
+          this.stopReceiveTimeout()
+          this._ws.onclose = () => {
+            resolve()
+          }
+          this._ws.close()
+        } else {
+          resolve()
+        }
+      })
+    }
+
+    return this._closingPromise
   }
 
   /**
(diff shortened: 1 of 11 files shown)


packages/bolt-connection/src/connection/connection-channel.js
```diff
@@ -393,7 +394,27 @@ export default class ChannelConnection extends Connection {
...
   /**
-   * This method still here because it's used by the {@link PooledConnectionProvider}
+   * This method is used by the {@link PooledConnectionProvider}
+   *
+   * @param {any} observer
+   */
+  _setIdle (observer) {
+    this._idle = true
+    this._ch.stopReceiveTimeout()
+    return this._protocol.queueObserverIfProtocolIsNotBroken(observer)
```
robsdedude: This return value isn't being used, is it?

packages/bolt-connection/src/connection/connection-channel.js
```diff
@@ -393,7 +394,27 @@ export default class ChannelConnection extends Connection {
...
+  }
+
+  /**
+   * This method is used by the {@link PooledConnectionProvider}
+   *
+   */
```
robsdedude:
```suggestion
  /**
   * This method is used by the {@link PooledConnectionProvider}
   */
```

packages/bolt-connection/src/channel/browser/browser-channel.js
```diff
@@ -163,17 +164,23 @@ export default class WebSocketChannel {
...
-      if (this._ws && this._ws.readyState !== WS_CLOSED) {
-        this._open = false
-        this.stopReceiveTimeout()
-        this._ws.onclose = () => resolve()
-        this._ws.close()
-      } else {
-        resolve()
-      }
-    })
+    if (this._closingPromise === null) {
```
robsdedude: At first glance, this change seems unrelated. Is it? 

I'm pretty sure the change is, if anything, beneficial. Yet, I'd like to understand if it was necessary or precaution and whether/how it relates to the timeout fix.

robsdedude (approved):
:rabbit2:

packages/bolt-connection/src/connection/connection-channel.js
bigmontz: It isn't.

packages/bolt-connection/src/channel/browser/browser-channel.js
bigmontz: During the tests, I noticed problems with the connection get much time to closed in browser and then other parts of the code calling the close method also, this was causing very confusing error. This code protects for this cases. 

I can move it for other PR, if it fits better.

robsdedude: Sure would be cleaner to move it into a separate PR, but I don't mind sneaking in a small fix ;)

Nice catch either way :D

bigmontz: Moved to other PR https://github.com/neo4j/neo4j-javascript-driver/pull/1169

PR #1169, opened 2023-12-27
Fix BrowserChannel close when is already closing

bigmontz:
WebSockets can take a while for closing, and when `WebSocket.close` is called while the status is `WS_CLOSING`, an error is thrown.

This situation can happen when the driver is being closed while sessions are still releasing connections back to the pool or after a receive timeout while closing the session.

BrowserChannel should wait for the original close socket to finish whenever a new close request happens to avoid this kind of error and provide a graceful shutdown for driver and session.


robsdedude (approved):
:parasol_on_ground:

PR #1170, opened 2023-12-27
Skip test not supported on browser suite

bigmontz:
Closing sockets in browser takes too much time, so the test just times out.


robsdedude (approved):
:kangaroo:

PR #1172, opened 2024-01-09
Improve AuthProvider performance

bigmontz:
This method was doing expensive object comparison for objects when connection auth token is null.
This causes a performance penalty on the driver.


vongruenigen:
@bigmontz is there some kind of policy who and what can be reviewed by non-neo4j engineers? If it allows for I could very well imagine starting to tinker more with the driver and maybe assist you with some reviews for smaller MRs (such as this one), if my time permits.

PS: Sorry if this is kinda off-topic regarding this MR, wasn't really sure where to post this.

PR #1173, opened 2024-01-11
Fix duplicated `neo4j-javascript` string on  `boltAgent.product` object

bigmontz:
This duplication was happening because the neo4j methods were sending the agent string instead of the driver version the bolt agent factory.


robsdedude (approved):
:cactus:

PR #1177, opened 2024-02-01
Introduce `auth` config to `Driver.executeQuery`

bigmontz:
The AuthToken will be used for executing the query.

By default, the query executor will use connections authenticated with the AuthToken configured on driver creation. This new configuration allows switching user and/or authorization information for the underlying transaction's lifetime.

Example:

```javascript
const { records } = await driver.executeQuery('RETURN $abc', { abc: 'dfe' }, {
   database: 'neo4j',
   auth: neo4j.auth.basic('otheruser', 'sup3rDup3rS3cret')
})
```
> [!NOTE]
> This option is only available when the driver is connected to Neo4j Database servers which supports Bolt 5.1 or newer.

### DIFF
--- a/packages/core/src/driver.ts
+++ b/packages/core/src/driver.ts
@@ -418,12 +418,12 @@ class QueryConfig<T = EagerResult> {
     /**
      * The {@link AuthToken} which will be used for executing the query.
      *
-     * By default, the query executor will use connections authenticated with {@link AuthToken} configured in the
-     * driver creation. This configuration allows switch user and/or authorization information for the
-     * execution lifetime.
+     * By default, the query executor will use connections authenticated with the {@link AuthToken} configured on
+     * driver creation. This configuration allows switching user and/or authorization information for the
+     * underlying transaction's lifetime.
      *
-     * **Warning**: This option is only enable when the driver is connected with Neo4j Database servers
-     * which supports Bolt 5.1 and onwards.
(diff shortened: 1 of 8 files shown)


packages/core/src/driver.ts
```diff
@@ -413,6 +414,21 @@ class QueryConfig<T = EagerResult> {
...
+
+    /**
+     * The {@link AuthToken} which will be used for executing the query.
+     *
+     * By default, the query executor will use connections authenticated with {@link AuthToken} configured in the
+     * driver creation. This configuration allows switch user and/or authorization information for the
+     * execution lifetime.
+     *
+     * **Warning**: This option is only enable when the driver is connected with Neo4j Database servers
+     * which supports Bolt 5.1 and onwards.
```
robsdedude:
```suggestion
     * The {@link AuthToken} which will be used for executing the query.
     *
     * By default, the query executor will use connections authenticated with the {@link AuthToken} configured on
     * driver creation. This configuration allows switching user and/or authorization information for the
     * underlying transaction's lifetime.
     *
     * **Warning**: This option is only available when the driver is connected to Neo4j Database servers
     * which support Bolt 5.1 or newer.
```

I know this is mostly a copy pasta from the corresponding session config...

robsdedude (approved):
:tropical_drink:

packages/core/src/driver.ts
bigmontz: I will also address it on the session config

PR #1181, opened 2024-02-19
testkit: Enable browser tests log on terminal

bigmontz:
Introduce a Console message in the protocol between the LocalController on Browser and the RemoteController in Node for enabling the logs and other console messages to be printed in terminal and be available in artifacts.

This feature should help debugging the errors in the testing pipeline.

### DIFF
--- a/packages/testkit-backend/src/console.remote.js
+++ b/packages/testkit-backend/src/console.remote.js
@@ -1,0 +1,31 @@
+import { response } from './responses'
+
+const originalConsole = console
+
+export default {
+  install: (channel) => {
+    // eslint-disable-next-line no-global-assign
+    console = new Proxy({}, {
+      get: (_, method) => (...args) => {
+        originalConsole[method].apply(originalConsole, args)
+        channel.writeResponse(null, response('Console', {
+          method,
+          args
+        }), true)
+      }
+    })
+  },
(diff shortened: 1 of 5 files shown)


packages/testkit-backend/src/console.remote.js
```diff
@@ -0,0 +1,31 @@
...
+  handleConsole: (message) => {
+    if (message.response.name === 'Console') {
+      const { method, args } = message.response.data
+      args[0] = typeof args[0] === 'string' ? `[RemoteConsole] ${args[0]}` : args[0]
+      console[method].apply(console, args)
+      return true
+    }
+    return false
+  },
+  uninstall: () => {
```
StephenCathcart, line 27: More of a question - do we need/use this uninstall function or is this just for future-proofing?

StephenCathcart (approved):
LGTM! 🚀

packages/testkit-backend/src/console.remote.js
bigmontz: We don't need, this is only implemented for future-proofing and for keeping the symmetry.

StephenCathcart: Brilliant sounds good to me, just checking in case it was missing somewhere 👍

PR #1192, opened 2024-05-10
Update build tools in Monorepo

bigmontz:
The tooling on the mono repository are quite outdate and this is causing extra work for actions like update dependencies in the internal repositories. 

Changes:

- Move monorepo repo configuration to npm workspaces
- Packaging manager changed to Yarn


MaxAake:
The build tools were updated to pnpm in #1348

PR #1194, opened 2024-05-16
Introduce `GqlStatusObject` support as notifications to `ResultSummary`

bigmontz:

Introduces `ResultSummary.gqlStatusObjects` as preview feature. 

`GqlStatusObject` is a GQL compliant `Notification` object and status of query execution, this new object includes `gqlStatus` and `statusDescription`.

`ResultSummary.gqlStatusObjects` always contains at least 1 status representing the `Success`, `No Data` or `Omitted Result`. When discarding records in a RxSession while connected to a non-GQL-aware server, the driver might not be able to tell apart `Success` and `No Data`.

The GqlStatusObjects will be presented in the following order:

* A “no data” (`02xxx`) has precedence over a warning;
* A warning (`01xxx`) has precedence over a success.
* A success (`00xxx`) has precedence over anything informational (`03xxx`)

### Migrating from Notification

Most of the properties present in the `Notification` are present in the new object, except `code` and `title`. 
The `GqlStatusObject.gqlStatus` is equivalent of `code` in the `Notification`, but with values compliant with GQL.
The properties `GqlStatusObject.classification` and `GqlStatusObject.rawClassification` are equivalent to `Notification.category` and `Notification.rawCategory`. The name change is needed because `category` has a different meaning in GQL.

### Configuration
Filtering `gqlStatusObjects` can be done using in the same way as filtering `notifications`. However, non-notification status can not be filtered (such as `Success` or `No Data`).
The property `disabledClassifications` was added to `NotificationFilter` for being named consistently with GQL. This property is equivalent to `disabledCategories` and it be can used interchangeably. However, they can not be used at the same time. An exception will be raised in this scenario.   


packages/core/src/driver.ts
```diff
@@ -901,6 +901,11 @@ function validateConfig (config: any, log: Logger): any {
       'where a new connection is created while it is acquired'
     )
   }
+
+  if (config.notificationFilter?.disabledCategories != null && config.notificationFilter?.disabledClassifications != null) {
+    throw new Error('The notificationFilter can have both "disabledCategories" and  "disabledClassifications" configured at same time.')
```
StephenCathcart:
```suggestion
    throw new Error('The notificationFilter can't have both "disabledCategories" and  "disabledClassifications" configured at the same time.')
```

packages/bolt-connection/test/bolt/behaviour/notification-filter.js
```diff
@@ -242,8 +375,12 @@ export function notificationFilterFixture () {
 function notificationFilterSetFixture () {
   const minimumSeverityLevelSet = Object.values(notificationFilterMinimumSeverityLevel)
   const disabledCategories = Object.values(notificationFilterDisabledCategory)
+  // TODO: Fix the type
```
StephenCathcart: Is this something that will be fixed when the deprecated categories are removed?

packages/core/src/notification.ts
```diff
@@ -0,0 +1,581 @@
...
+  '02': 0,
+  '01': 1,
+  '00': 2
+})
+/**
+ * GqlStatus weight
+ *
+ * @private
+ */
+function calculateWeight (gqlStatusObject: GqlStatusObject): number {
```
StephenCathcart, line 527: Is the weight stuff just to ensure the status object is first in the array? I didn't see any weight mentioned in the ADR.

StephenCathcart (approved):
Couple of general questions but other than that it seems to meet the ADR spec well 😄 👍

packages/bolt-connection/test/bolt/behaviour/notification-filter.js
bigmontz: When I wrote the test, the classification type wasn't written yet. So, I started with the categories since the value set is the same. A let this hint for the future me fix, good that you found it. Now, this is fixed. with `notificationFilterDisabledClassification`.

packages/core/src/notification.ts
bigmontz: We have it defined here: 

https://github.com/neo-technology/drivers-adr/pull/89/files#diff-2c370520cd34650b73ee220b6414ba2ed657d6ee495849ac3f20eee2112439f7R167

robsdedude:
PR description is not up-to-date.

> `ResultSummary.gqlStatusObjects` always contains at least 1 status representing the `Success`, `No Data` or `Omitted Result`. This status will be always the first one.

I thought this isn't the case.

> When discarding records while connected to a non-GQL aware server, the driver might not be able to tell apart `Success` and `No Data`.

This only applies to the reactive driver, doesn't it? Might be worth being explicit here.

> All following status are notifications like warnings about problematic queries or other valuable information that can be presented in a client.

See first point.

packages/bolt-connection/src/bolt/bolt-protocol-v5x5.js
```diff
@@ -0,0 +1,189 @@
...
+  /**
+   *
+   * @param {object} metadata
+   * @returns {object}
+   */
+  static _enrichMetadata (metadata) {
+    if (Array.isArray(metadata.statuses)) {
+      metadata.statuses = metadata.statuses.map(status => ({
+        ...status,
+        diagnostic_record: status.diagnostic_record !== null ? { ...DEFAULT_DIAGNOSTIC_RECORD, ...status.diagnostic_record } : null
```
robsdedude, line 183: Now this implementation is not what I went with just reading the ADR. I decided to also fill in the `DEFAULT_DIAGNOSTIC_RECORD` if the server sent `"diagnostic_record": null` or any other unexpected type.

I reckon that this difference is important to leave us the option to null out the whole diag. record in future server versions. Seems like we should have a TestKit test for this.

packages/bolt-connection/src/bolt/request-message.js
```diff
@@ -325,6 +367,33 @@ export default class RequestMessage {
...
 
+  /**
+   * Create a new RUN message with additional metadata.
+   * @param {string} query the cypher query.
+   * @param {Object} parameters the query parameters.
+   * @param {Bookmarks} bookmarks the bookmarks.
+   * @param {TxConfig} txConfig the configuration.
+   * @param {string} database the database name.
+   * @param {string} mode the access mode.
+   * @param {string} impersonatedUser the impersonated user.
```
robsdedude: Docs comment is missing `notificationFilter`. Also applies to `runWithMetadata`.

But let me be frank here: this comment isn't really useful at all. For one, it's an internal function. But more importantly, the param docs are all of the sorts "XYZ: it's XYZ". No :open_mouth:, you don't say! :stuck_out_tongue_winking_eye:

packages/bolt-connection/src/bolt/request-message.js
```diff
@@ -439,7 +508,7 @@ export default class RequestMessage {
  * @param {notificationFilter} notificationFilter the notification filter
  * @return {Object} a metadata object.
  */
-function buildTxMetadata (bookmarks, txConfig, database, mode, impersonatedUser, notificationFilter) {
+function buildTxMetadata (bookmarks, txConfig, database, mode, impersonatedUser, notificationFilter, functions = {}) {
```
robsdedude, line 517: The new parameter is not documented. Neither is it's naming self-explanatory.

packages/bolt-connection/src/bolt/stream-observers.js
```diff
@@ -303,7 +317,9 @@ class ResultStreamObserver extends StreamObserver {
         for (let i = 0; i < meta.fields.length; i++) {
           this._fieldLookup[meta.fields[i]] = i
         }
+      }
 
+      if (meta.fields != null) {
```
robsdedude, line 322:
```suggestion
      if (meta.fields !== undefined) {
```
Not 100% sure on this one, but wouldn't you want to strip `null` for fields as well? Sure, it's a protocol violation, so maybe it goes :boom: before this code already.

packages/bolt-connection/src/bolt/stream-observers.js
```diff
@@ -501,7 +518,7 @@ class ResetObserver extends StreamObserver {
     this.onError(
       newError(
         'Received RECORD when resetting: received record is: ' +
-          json.stringify(record),
+        json.stringify(record),
```
robsdedude, line 521: I don't know much about JS style guides out there so this might be the norm. But personally, I find it hard to read where which argument starts and ends.

packages/bolt-connection/test/bolt/behaviour/notification-filter.js
```diff
@@ -178,6 +223,46 @@ export function shouldSupportNotificationFilterOnBeginTransaction (createProtoco
   })
 }
 
+/**
+ * @param {function(recorder:MessageRecorder):BoltProtocolV1} createProtocol The protocol factory
+ */
+export function shouldSupportGqlNotificationFilterOnBeginTransaction (createProtocol) {
+  it.each(
+    notificationFilterFixture()
+  )('should send notificationsFilter=%o on begin a transaction', (notificationFilter) => {
```
robsdedude:
```suggestion
  )('should send notificationsFilter=%o on beginning a transaction', (notificationFilter) => {
```

packages/bolt-connection/test/bolt/bolt-protocol-v5x5.test.js
```diff
@@ -0,0 +1,1499 @@
...
+      expect(packable).not.toThrow()
+
+      buffer.reset()
+
+      const unpacked = protocol.unpack(buffer)
+      expect(unpacked).toEqual(struct)
+    })
+  })
+
+  describe('result metadata enrich', () => {
```
robsdedude:
```suggestion
  describe('result metadata enrichment', () => {
```

packages/bolt-connection/test/bolt/bolt-protocol-v5x5.test.js
```diff
@@ -0,0 +1,1499 @@
...
+          }),
+          statusesWithDiagnosticRecord({
+            OPERATION: undefined,
+            OPERATION_CODE: undefined,
+            CURRENT_SCHEMA: undefined,
+            _status_parameters: undefined,
+            _severity: undefined,
+            _classification: undefined,
+            _position: undefined
+          })
```
robsdedude, line 1468: For my understanding: is this code path even possible? Can the server send something that will make the driver call `protocol._enrichMetadata` with `undefined` values in the diag. record?

packages/bolt-connection/test/bolt/stream-observer.test.js
```diff
@@ -680,6 +701,191 @@ describe('#unit ResultStreamObserver', () => {
...
+      await new Promise((resolve, reject) => {
+        setImmediate(() => {
+          try {
+            streamObserver.onCompleted({ key: 42, has_more: false })
+            resolve()
+          } catch (e) {
+            reject(e)
+          }
+        })
+      })
```
robsdedude, line 767: What is this async dance good for if the promise is awaited right away?

(Applies to more sections below, won't comment again)

packages/core/src/index.ts
```diff
@@ -57,20 +57,26 @@ import { isPoint, Point } from './spatial-types'
...
   QueryStatistics,
-  Stats,
+  Stats
+} from './result-summary'
+import Notification, {
+  NotificationPosition,
   NotificationSeverityLevel,
+  NotificationClassification,
   NotificationCategory,
+  GqlStatusObject,
```
robsdedude, line 70: How much of a breaking change will it be to move things within the "internal" package structure? Because I anticipate that `GqlStatusObject` might move. It's not specific to notifications but will also be used for errors later down the line.

packages/core/src/notification-filter.ts
```diff
@@ -83,10 +98,27 @@ class NotificationFilter {
 
     /**
      * Categories the user would like to opt-out of receiving.
+     *
+     *
+     * This property is equivalent to {@link NotificationFilter#disabledClassifications}
+     * and it should not be enabled at same time.
```
robsdedude:
```suggestion
     * and it must not be enabled at same time.
```

packages/core/src/notification-filter.ts
```diff
@@ -83,10 +98,27 @@ class NotificationFilter {
...
      * @type {?NotificationFilterDisabledCategory[]}
      */
     this.disabledCategories = undefined
 
+    /**
+     * Classifications the user would like to opt-out of receiving.
+     *
+     * This property is equivalent to {@link NotificationFilter#disabledCategories}
+     * and it should not be enabled at same time.
+     *
```
robsdedude:
```suggestion
     * and it must not be enabled at same time.
```

packages/core/src/notification.ts
```diff
@@ -0,0 +1,592 @@
...
+  },
+  INFORMATION: {
+    gql_status: '03N42',
+    status_description: 'info: informational - unknown notification'
+  },
+  ERROR: {
+    gql_status: '50N42',
+    status_description: 'error: general processing exception - unknown error'
+  }
+}
```
robsdedude:
```suggestion
  WARNING: {
    gql_status: '01N42',
    status_description: 'warn: unknown warning'
  },
  NO_DATA: {
    gql_status: '02N42',
    status_description: 'note: no data - unknown subcondition'
  },
  INFORMATION: {
    gql_status: '03N42',
    status_description: 'info: unknown notification'
  },
  ERROR: {
    gql_status: '50N42',
    status_description: 'error: general processing exception - unknown error'
  }
}
```
Please double-check. This is how I read section 4.9.2 of the standard anyway.

packages/core/src/notification.ts
```diff
@@ -0,0 +1,592 @@
...
+     *             break
+     *         case neo4j.notificationSeverityLevel.UNKNOWN: // or simply 'UNKNOWN'
+     *         default:
+     *             // the raw info came from the server could be found at notification.rawSeverityLevel
+     *             console.log(`${notification.title} - ${notification.description}`)
+     *             break
+     *     }
+     * }
+     */
+    this.severityLevel = _asEnumerableSeverity(this.severity)
```
robsdedude:
```suggestion
    this.severityLevel = _asEnumerableSeverity(notification.severity)
```
nit: potential for spooky action at a distance.

packages/core/src/notification.ts
```diff
@@ -0,0 +1,592 @@
...
+}
+
+interface NotificationDiagnosticRecord {
+  OPERATION: string
+  OPERATION_CODE: string
+  CURRENT_SCHEMA: string
+  _severity?: string
+  _classification?: string
+  _position?: object
+  _status_parameters?: object
```
robsdedude: Any reason why you decided to not go down further down with the typing?

packages/core/src/notification.ts
```diff
@@ -0,0 +1,592 @@
...
+    /**
+     * The diagnostic record as it is.
+     *
+     * @type {object}
+     * @public
+     */
+    this.diagnosticRecord = rawGqlStatusObject.diagnostic_record ?? {}
+
+    /**
+     * The position which the notification had occur.
```
robsdedude:
```suggestion
     * The position at which the notification had occurred.
```

packages/core/src/notification.ts
```diff
@@ -0,0 +1,592 @@
...
+     *     switch(gqlStatusObject.classification) {
+     *         case neo4j.notificationClassification.QUERY: // or simply 'QUERY'
+     *             console.info(gqlStatusObject.statusDescription)
+     *             break
+     *         case neo4j.notificationClassification.PERFORMANCE: // or simply 'PERFORMANCE'
+     *             console.warn(gqlStatusObject.statusDescription)
+     *             break
+     *         case neo4j.notificationClassification.UNKNOWN: // or simply 'UNKNOWN'
+     *         default:
+     *             // the raw info came from the server could be found at notification.rawCategory
```
robsdedude:
```suggestion
     *             // the raw info came from the server can be found at notification.rawCategory
```

packages/core/src/notification.ts
```diff
@@ -0,0 +1,592 @@
...
+
+  /**
+   * The json string representation of the diagnostic record.
+   * The goal of this method is provide a serialized object for human inspection.
+   *
+   * @type {string}
+   * @public
+   */
+  public get diagnosticRecordAsJsonString (): string {
+    return json.stringify(this.diagnosticRecord)
```
robsdedude: I wondered what would happen if the diag. record contained some non-trivial type like a node or datetime. While it works, the representation isn't ideal, I'd argue. Here's an example:
```
{"OPERATION":"","OPERATION_CODE":"0","CURRENT_SCHEMA":"/","_severity":"","_classification":"","_position":{"offset":0,"line":0,"column":0},"_status_parameters":{"key1":{"identity":1,"labels":["Label"],"properties":{"foo":"bar"},"elementId":"1"}}}
```
Somewhat hard to see (unless you know the neo4j ecosystem well) and impossible to know for sure that the value under `key1` is actually a Node. I could just as well be a map.

packages/core/src/result-summary.ts
```diff
@@ -105,7 +108,24 @@ class ResultSummary<T extends NumberOrInteger = Integer> {
...
+    /**
+     * A list of GqlStatusObjects that arise when executing the query.
+     *
+     * The list always contains at least 1 status representing the Success, No Data or Omitted Result.
+     * This status will be always the first one.
+     * When discarding records while connected to a non-gql aware server, the driver might not be able to
+     * tell apart Success and No Data.
+     *
+     * All following status are notifications like warnings about problematic queries or other valuable
+     * information that can be presented in a client.
```
robsdedude: Outdated info. See my comment here https://github.com/neo4j/neo4j-javascript-driver/pull/1194#issuecomment-2160565256

packages/core/test/notification.test.ts
```diff
@@ -0,0 +1,1516 @@
...
+          _position: {
+            offset: 0,
+            line: 0,
+            column: 0
+          },
+          _status_parameters: {}
+        }
+      }
+
+      const gqlStatusObject = polyfillNotification(status)
```
robsdedude: nit: isn't really a GQL status object

packages/core/test/notification.test.ts
```diff
@@ -0,0 +1,1516 @@
...
+        position: {
+          offset: 1,
+          line: 2,
+          column: 3
+        },
+        category: 'UNSUPPORTED'
+      }))
+    })
+
+    it('should polyfill when diagnostic records is not present', () => {
```
robsdedude:
```suggestion
    it('should polyfill when diagnostic record is not present', () => {
```

packages/core/test/notification.test.ts
```diff
@@ -0,0 +1,1516 @@
...
+      const rawGqlStatusObject = {
+        diagnostic_record: diagnosticRecord
+      }
+
+      const gqlStatusObject = new GqlStatusObject(rawGqlStatusObject)
+
+      expect(gqlStatusObject.diagnosticRecord).toBe(diagnosticRecord)
+    })
+
+    it('should fill position with values came from raw.diagnostic_record', () => {
```
robsdedude:
```suggestion
    it('should fill position with values from raw.diagnostic_record', () => {
```

robsdedude: More "came"s below. Won't comment again.

packages/core/test/notification.test.ts
```diff
@@ -0,0 +1,1516 @@
...
+      }))
+    })
+
+    it.each([
+      'UNKNOWN',
+      null,
+      undefined,
+      'I_AM_NOT_OKAY',
+      'information'
+    ])('should polyfill UNKNOWN', (severity) => {
```
robsdedude:
```suggestion
    ])('should polyfill severity UNKNOWN', (severity) => {
```
+ next test description

packages/core/test/notification.test.ts
```diff
@@ -0,0 +1,1516 @@
...
+          position: {
+            offset: 0,
+            line: 0,
+            column: 0
+          },
+          category: 'TOPOLOGY'
+        }]
+      }
+    ]
+  ])('should build from notifications when not available', (filledObject: GqlStatusObject, position: number, metadata: any) => {
```
robsdedude:
```suggestion
  ])('should build from notifications when statuses not available', (filledObject: GqlStatusObject, position: number, metadata: any) => {
```

packages/core/test/notification.test.ts
```diff
@@ -0,0 +1,1516 @@
...
+            offset: 0,
+            line: 0,
+            column: 0
+          },
+          category: 'TOPOLOGY'
+        }]
+      }
+    ],
+    [
+      getSuccessStatusObject(), 0, {
```
robsdedude: Too many arguments.

packages/bolt-connection/src/bolt/request-message.js
bigmontz: This docs are more for get some auto-complete in the JS code. I might bring this library to TS in the future, :D.

bigmontz: And now, I realize the params defined in the object destructors ({ param1 }) are not well documented in the file. I will fix for the `runWithMedata` functions on this PR and I will create a PR fixing the rest in the future.

packages/bolt-connection/src/bolt/stream-observers.js
bigmontz: at this point, the fields were already processed by the driver. This is only removing from the completion metadata, the one used to create the summary.

packages/bolt-connection/src/bolt/stream-observers.js
bigmontz: This was some auto-formatting in action

packages/core/src/notification.ts
bigmontz: true, better get the information from the notification instead of this, this way if we change the meaning of this.severity (or delete), severityLevel will not break

packages/core/src/notification.ts
bigmontz: What do you mean? we don't know for sure the shape of _status_parameter, we just know this is a key/value

packages/bolt-connection/test/bolt/stream-observer.test.js
bigmontz: this is just for the event-loop get a spin.

robsdedude: I see. Why is that needed for this test?

packages/bolt-connection/test/bolt/bolt-protocol-v5x5.test.js
bigmontz: No, the null values from the server are set as null, not as undefined.

packages/core/src/index.ts
bigmontz: I can extract a common base without breaking and the type is marked as @experimental, so we have space for moving case something went wrong. However, in my error spike, the Error doesn't need this object to exist in a gql compliant way.

packages/core/src/notification.ts
bigmontz: The description is define das `${short severity}: ${condition} - ${subcondtion}`

for 03N42, condition is `informational` and subcondition is `unknown notification`, for example. See the codes spreadsheet.

packages/core/src/result-summary.ts
bigmontz: updated

packages/core/src/notification.ts
bigmontz: Maybe I can move it use the `toString()` of the custom type, what do you think?

packages/bolt-connection/test/bolt/stream-observer.test.js
bigmontz: The is some async dance happening on the `onCompleted` call. This is not need for this test in specific.

packages/core/src/notification.ts
bigmontz: Something like: 

```
"{\"OPERATION\":\"\",\"OPERATION_CODE\":\"0\",\"CURRENT_SCHEMA\":\"/\",\"_severity\":\"\",\"_classification\":\"\",\"_position\":{\"offset\":0,\"line\":0,\"column\":0},\"_status_parameters\":{\"some_node\":\"(myElementId:ABC {bla:\\\"string\\\"})\"}}"
```

packages/bolt-connection/src/bolt/request-message.js
```diff
@@ -437,9 +510,11 @@ export default class RequestMessage {
  * @param {string} mode the access mode.
  * @param {string} impersonatedUser the impersonated user mode.
  * @param {notificationFilter} notificationFilter the notification filter
+ * @param {Object} functions Transformation functions applied to metadata
+ * @param {function(metadata,notificationFilter):void} functions.appendNotificationFilter Changes metadata by append the Notification Filter to it.
```
robsdedude:
```suggestion
 * @param {function(metadata,notificationFilter):void} functions.appendNotificationFilter Changes metadata by appending the Notification Filter to it.
```

packages/core/src/notification.ts
robsdedude: I mean that I was silly and didn't understand TypeScript properly :upside_down_face: 

All good here.

packages/core/src/result-summary.ts
```diff
@@ -105,7 +108,30 @@ class ResultSummary<T extends NumberOrInteger = Integer> {
...
+
+    /**
+     * A list of GqlStatusObjects that arise when executing the query.
+     *
+     * The list always contains at least 1 status representing the Success, No Data or Omitted Result.
+     *
+     * When discarding records while connected to a non-gql aware server and using a RxSession,
+     * the driver might not be able to tell apart Success and No Data.
+     *
+     * All following status are notifications like warnings about problematic queries or other valuable
```
robsdedude:
```suggestion
     * All other status are notifications like warnings about problematic queries or other valuable
```

robsdedude (approved):
:bell:

PR #1199, opened 2024-06-17
Introduce `AbortSignal` to `Driver.executeQuery`

bigmontz:
This configuration property enables the cancellation of ongoing queries which was previous not possible when using `Driver.executeQuery`.

The cancellation is done by closing the session used internally by the driver when execute the query. This means the cancellation is not safe and transaction might be commit apart of the execution being aborted. The cancellation might also triggers errors on the `Driver.executeQuery` execution.

Example:

```javascript
const abortController = new AbortController()

// Some event which cancels the query
process.on('SIGINT', () => abortController.abort() )

const result = await driver.executeQuery(query, params, {
    database: 'neo4j',
    signal: abortController.signal
})
```

### DIFF
--- a/packages/core/src/driver.ts
+++ b/packages/core/src/driver.ts
@@ -358,6 +358,7 @@ class QueryConfig<T = EagerResult> {
   resultTransformer?: ResultTransformer<T>
   transactionConfig?: TransactionConfig
   auth?: AuthToken
+  signal?: AbortSignal
 
   /**
    * @constructor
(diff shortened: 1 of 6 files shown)


packages/core/src/driver.ts
```diff
@@ -429,6 +430,23 @@ class QueryConfig<T = EagerResult> {
      * @see {@link driver}
      */
     this.auth = undefined
+
+    /**
+     * The {@link AbortSignal} for aborting query execution.
+     *
+     * When aborted, the signal triggers the result consumption cancelation and
+     * transactions are reset. However, due to race conditions,
+     * there is no warranty the transaction will be rolled back.
```
gjmwoods:
```suggestion
     * there is no guarantee the transaction will be rolled back.
```

gjmwoods (approved):

PR #1200, opened 2024-06-19
Introduce `resultTransformer.first`

bigmontz:
This function enables fetching only the first record in the Result. If any other record is present, it will be discarded and network optimization might be applied.

Examples:

```javascript
// using in the execute query
const maybeFirstRecord = await driver.executeQuery('MATCH (p:Person{ age: $age }) RETURN p.name as name', { age: 25 }, {
    database: 'neo4j,
    resultTransformer: neo4j.resultTransformers.first()
})
```

```javascript
// using in other a result

const maybeFirstRecord = await session.executeRead(tx => {
    // do not `await` or `resolve` the result before send it to the transformer
    const result = tx.run('MATCH (p:Person{ age: $age }) RETURN p.name as name', { age: 25 })
    return  neo4j.resultTransformers.first()(result)
})
```

### DIFF
--- a/packages/core/src/result-transformers.ts
+++ b/packages/core/src/result-transformers.ts
@@ -21,12 +21,6 @@ import EagerResult from './result-eager'
 import ResultSummary from './result-summary'
 import { newError } from './error'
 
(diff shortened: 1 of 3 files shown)


RichardIrons-neo4j (approved):
All looks good.

PR #1210, opened 2024-08-09
Fix `npm test` when running test containers

bigmontz:
The tests were failing because Deno integration tests can't access a database.
The problem happens because tests are using `testcontainers` by default and Deno doesn't have support for it yet.

So, the problem is circumvented by skipping the Deno integration tests for the case we are using `testcontainers`.
This is not a big issue since,

1. Pipelines and testkit doesn't make use of testcontainers, so the tests will not be skipped in important places.
2. The amount of testing scenarios are quite small and not likely to change in behaviour.


MaxAake (approved):
🥇

PR #1212, opened 2024-08-27
Move Pool module to `core`

bigmontz:
The `pool` holds logic on resource pooling and it is used to avoid better resource management, enabling the driver to re-use connections and avoiding overflow server and cluster with too much connections. The pool also regulates acquisition timeouts, this avoiding clients to wait an undetermined amount time for acquiring resource when driver is busy. 

This module is being moved to the core module to enabling drivers developers to re-use this module in context which is not Bolt.


MaxAake (approved):
💯 LGTM

PR #1213, opened 2024-08-29
Removal of duplicate error listener on socket

MaxAake:
The socket adds a listener for 'error' events twice, this results in the same error messages being handled twice, causing confusing logs and potentially unwanted behaviors.

This PR removes one of the listeners, resulting in just one listener on the socket.


MaxAake:
[[JS] Investigate ECONNRESET in JS](https://trello.com/c/rIdMNq9h/439-js-investigate-econnreset-in-js)

bigmontz (approved):
💯

PR #1214, opened 2024-08-29
bump deno version to 5.25.0

MaxAake:
bump deno version to 5.25.0 for next release


bigmontz (approved):
25 follows 24, for sure.

PR #1215, opened 2024-08-30
Close channels on destroying errors

MaxAake:
Set channels to close when connections are destroyed by errors. This fixes an issue where sessions would not retry if they were closed by a connection error rather than a connection termination.

includes unit tests for both destroying and non-destroying errors, ensuring that channels are closed and kept open respectively.


MaxAake:
[[JS] Investigate ECONNRESET in JS](https://trello.com/c/rIdMNq9h/439-js-investigate-econnreset-in-js)

bigmontz (approved):
🥇

PR #1216, opened 2024-09-02
Fix `maxConnectionPoolSize` verification

bigmontz:
Under high load of sessions, the connection pool is allowing new connections to be created.
This is happening because, while the development of `AuthTokenManager` and `connection liveness check`,
the methods do validate connection on acquired and return to the pool need to be async.
This changes creates a situation of racing condition which doesn't exist the original code and this introduces this bug.

Increasing the resouce acquired count before validating the connection solves the issue, since the race condition is removed. However, the pool needs also to release the resource when the validation return `false` or fails for some reason.
This is important to avoid broken connection still be count for the max pool size.

### DIFF
--- a/packages/core/src/internal/pool/pool.ts
+++ b/packages/core/src/internal/pool/pool.ts
@@ -251,19 +251,32 @@ class Pool<R extends unknown = unknown> {
           continue
         }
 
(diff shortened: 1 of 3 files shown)


MaxAake (approved):
LGTM! 💯

PR #1220, opened 2024-09-06
Throw error from validation failure

MaxAake:
Changes the connection pool behavior on errors during validation, these would be unintended behavior and as such the driver should throw an error if it occurs, rather than simply logging as was the case in the 5.24.1 patch release fix.

Also updates the test to control for this behavior.


bigmontz (approved):
🍎

PR #1225, opened 2024-09-20
GQL errors and update to bolt protocol 

MaxAake:
Implements GQL compliant properties on the Neo4jError object, which handles both error messages from the database and driver-side errors. Also polyfills these properties if an error is created without them. Old error messages and codes will still be available even for new GQL compliant errors to accommodate existing error handling.


MaxAake:
[[JavaScript] Implement GQL Errors as per ADR](https://trello.com/c/cht3fUuj/469-javascript-implement-gql-errors-as-per-adr)

packages/bolt-connection/src/bolt/bolt-protocol-v5x6.js
```diff
@@ -31,7 +31,7 @@ const DEFAULT_DIAGNOSTIC_RECORD = Object.freeze({
   CURRENT_SCHEMA: '/'
 })
 
-export default class BoltProtocol extends BoltProtocolV5x5 {
+export default class BoltProtocolV5x6 extends BoltProtocolV5x5 {
```
bigmontz:
```suggestion
export default class BoltProtocol extends BoltProtocolV5x5 {
```

packages/bolt-connection/src/bolt/response-handler.js
```diff
@@ -196,6 +193,15 @@ export default class ResponseHandler {
...
   }
+
+  _handleErrorPayload (payload) {
+    const standardizedCode = _standardizeCode(payload.code)
+    const cause = payload.cause != null ? this._handleErrorPayload(payload.cause) : undefined
+    const error = newError(payload.message, standardizedCode, cause, payload.gql_status, payload.description, payload.diagnostic_record)
+    return this._observer.onErrorApplyTransformation(
+      error
+    )
+  }
```
bigmontz: The code will be not present in the cause. I suggest to make a small in the Neo4jError, extract a `GqlError` object which will be used for the causes and `Neo4jError` extends it adding the code. What do you think? This will make clear the cause don't have code.

packages/bolt-connection/test/bolt/response-handler.test.js
```diff
@@ -69,8 +70,84 @@ describe('response-handler', () => {
       expect(receivedError.message).toBe(expectedError.message)
       expect(receivedError.code).toBe(expectedError.code)
     })
+
+    it('should correctly handle errors with gql data', () => {
```
bigmontz, line 75: Might need to check if the enrich method is being called when needed.

packages/core/src/error.ts
```diff
@@ -60,22 +77,80 @@ class Neo4jError extends Error {
...
+     * @public
+     */
     this.code = code
+    /**
+     * Optional, nested error which caused the error
+     *
+     * @type {Error?}
+     * @public
+     */
+    this.cause = cause != null ? cause : undefined
```
bigmontz, line 109: Could you check how the stacktrace is presented?

packages/core/src/error.ts
```diff
@@ -96,17 +171,32 @@ class Neo4jError extends Error {
...
+ * @param {String} [gqlStatus]
+ * @param {String} [gqlStatusDescription]
+ * @param {ErrorDiagnosticRecord} diagnosticRecord - the error message
  * @return {Neo4jError} an {@link Neo4jError}
  * @private
  */
-function newError (message: string, code?: Neo4jErrorCode, cause?: Error): Neo4jError {
-  return new Neo4jError(message, code ?? NOT_AVAILABLE, cause)
+function newError (message: string, code?: Neo4jErrorCode, cause?: Neo4jError, gqlStatus?: string, gqlStatusDescription?: string, diagnosticRecord?: ErrorDiagnosticRecord): Neo4jError {
+  return new Neo4jError(message, code ?? NOT_AVAILABLE, gqlStatus ?? '50N42', gqlStatusDescription ?? 'error: general processing exception - unexpected error. ' + message, diagnosticRecord, cause)
```
bigmontz: we have some constants defined at `notification.ts`. It might worth to extract some gql common to a single file.

packages/core/src/error.ts
```diff
@@ -96,17 +171,32 @@ class Neo4jError extends Error {
...
+ * @param {Neo4jError} [cause]
+ * @param {String} [gqlStatus]
+ * @param {String} [gqlStatusDescription]
+ * @param {ErrorDiagnosticRecord} diagnosticRecord - the error message
  * @return {Neo4jError} an {@link Neo4jError}
  * @private
  */
-function newError (message: string, code?: Neo4jErrorCode, cause?: Error): Neo4jError {
-  return new Neo4jError(message, code ?? NOT_AVAILABLE, cause)
+function newError (message: string, code?: Neo4jErrorCode, cause?: Neo4jError, gqlStatus?: string, gqlStatusDescription?: string, diagnosticRecord?: ErrorDiagnosticRecord): Neo4jError {
```
bigmontz: The cause is a Error, it used in other places already.

packages/neo4j-driver/test/internal/connection-channel.test.js
```diff
@@ -300,7 +300,7 @@ describe('#integration ChannelConnection', () => {
             .then(() => done.fail('Should fail'))
             .catch(error => {
               expect(error.message).toEqual(
-                'Received FAILURE as a response for RESET: Neo4jError: Hello'
+                'Received FAILURE as a response for RESET: Neo4jError'
```
bigmontz: Why did it change?

testkit/testkit.json
```diff
@@ -1,6 +1,6 @@
 {
   "testkit": {
     "uri": "https://github.com/neo4j-drivers/testkit.git",
-    "ref": "5.0"
+    "ref": "bolt-5x7-and-gql-errors"
```
bigmontz: Remember to remove it before merge

bigmontz (changes requested):
The changes look good in general, I have some questions and suggestions.

packages/bolt-connection/test/bolt/response-handler.test.js
```diff
@@ -142,7 +144,7 @@ describe('response-handler', () => {
   })
 
   function testDiagnosticRecord (diagnostic_record, expected_diagnostic_record) {
-    expect(diagnostic_record.OPERATION).toBe(expected_diagnostic_record.OPERATION)
+    expect(diagnostic_record.OPERATION).toBe(expected_diagnostic_record.OPERATION ?? '')
```
bigmontz: The expected diagnostic record should be informed with the actual value, we should not to this defaulting here because it can be confusing.

packages/core/src/error.ts
```diff
@@ -188,10 +188,6 @@ class Neo4jError extends GQLError {
     this.retriable = _isRetriableCode(code)
   }
 
-  toString (): string {
-    return this.name + ': ' + this.message
-  }
```
bigmontz: why?

MaxAake: This was an addition made when it seemed that the toString was broken, when that was solved this custom toString was not necessary.

bigmontz (approved):
q.e.d.

PR #1228, opened 2024-09-26
TestKit: add NumberIsNumber feature flag

robsdedude:
This implementation detail flag signals that the driver is not (fully) capable of differentiating integer and floating point types.

See also:
 * https://github.com/neo4j-drivers/testkit/pull/632


MaxAake (approved):
Truisms are Truisms!

PR #1234, opened 2024-10-16
make isUnboundRelationship private

MaxAake:
UnboundRelationship is a class that is used internally in the driver, and we do not need to document functions related to it to the users.


AndyHeap-NeoTech (approved):
Yup, those two lines check out :)

PR #1235, opened 2024-10-16
Home Database Cache : Stripped Back Solution

MaxAake:
Spikes implementation of a home database cache in the driver object, used to reduce round trips associated with identifying the home database of a user when a database is not set manually.

This behavior is turned on by a flag introduced in bolt 5.8, so the PR also includes support for Bolt 5.8

This work was migrated from PR [#1230](https://github.com/neo4j/neo4j-javascript-driver/pull/1230)


MaxAake:
[[JavaScript] HomeDatabase stripped back solution spike](https://trello.com/c/LveZmsf4/477-javascript-homedatabase-stripped-back-solution-spike)

packages/bolt-connection/src/connection/connection-channel.js
```diff
@@ -331,6 +335,16 @@ export default class ChannelConnection extends Connection {
               if (telemetryEnabledHint === true) {
                 this._telemetryDisabledConnection = false
               }
+
+              const SSREnabledHint = metadata.hints['ssr.enabled']
+              if (SSREnabledHint) {
```
robsdedude:
```suggestion
              if (SSREnabledHint === true) {
```
Realistically, we should never see any other truthy value from the server, but I'm in favor of practicing defensive input validation :)

packages/bolt-connection/src/connection/connection-channel.js
```diff
@@ -331,6 +335,16 @@ export default class ChannelConnection extends Connection {
...
+                this.serversideRouting = true
+              } else {
+                this.serversideRouting = false
+              }
+              if (this._ssrCallback !== undefined) {
+                this._ssrCallback(this.serversideRouting, true)
```
robsdedude: nit: hard to read code. I can't see what this boolean flag does without looking at some/all implementations of the callback. I suggest to either use an [enum-ish](https://masteringjs.io/tutorials/fundamentals/enum) construct, different callbacks for opening and closing, or a single callback that gets passed the connection as only argument and uses `con.isOpen()` and `con.serversideRouting` to determine what to do.

packages/bolt-connection/src/connection/connection-channel.js
```diff
@@ -43,6 +43,7 @@ export function createChannelConnection (
   log,
   clientCertificate,
   serversideRouting = null,
+  ssrCallback = (_) => {},
```
robsdedude: nit (because internal function): new parameter not in doc comment

packages/bolt-connection/src/connection/connection-channel.js
```diff
@@ -331,6 +335,16 @@ export default class ChannelConnection extends Connection {
...
                 this._telemetryDisabledConnection = false
               }
+
+              const SSREnabledHint = metadata.hints['ssr.enabled']
+              if (SSREnabledHint) {
+                this.serversideRouting = true
+              } else {
+                this.serversideRouting = false
+              }
+              if (this._ssrCallback !== undefined) {
```
robsdedude: Is this check necessary? Can one pass in undefined with the argument having a default value? (same check also in `close()`)

packages/bolt-connection/test/bolt/bolt-protocol-v5x8.test.js
```diff
@@ -0,0 +1,1579 @@
...
+    '',
+    undefined,
+    null
+  ])('should always use the user agent set by the user', (userAgent) => {
+    const recorder = new utils.MessageRecordingConnection()
+    const protocol = new BoltProtocolV5x8(recorder, null, false)
+    utils.spyProtocolWrite(protocol)
+
+    const boltAgent = {
+      product: 'neo4j-javascript/5.8',
```
robsdedude: For the sake of the test, this string can be practically whatever. But it being bumped with the protocol version makes me think maybe that's not clear :upside_down_face: in a real conversation with a server, this represents the driver's version, not the protocol's.

(other places where the bolt or user agent string has been bumped to be in sync with the protocol version under test)

packages/bolt-connection/src/bolt/bolt-protocol-v5x8.js
```diff
@@ -0,0 +1,39 @@
...
+const {
+  constants: { BOLT_PROTOCOL_V5_8 }
+} = internal
+
+export default class BoltProtocol extends BoltProtocolV5x7 {
+  get version () {
+    return BOLT_PROTOCOL_V5_8
+  }
+
+  get transformer () {
```
robsdedude, line 35: JS noob question: Why is this override that's the exact same as its parent class necessary?

packages/core/src/session.ts
```diff
@@ -508,7 +519,12 @@ class Session {
    * @returns {void}
    */
   _onDatabaseNameResolved (database?: string): void {
+    this._databaseGuess = database
     if (!this._databaseNameResolved) {
+      if (this._homeDatabaseCallback != null) {
+        // eslint-disable-next-line
+        this._homeDatabaseCallback((this._impersonatedUser ? 'basic:' + this._impersonatedUser : undefined)  ?? this._auth?.cacheKey ?? 'DEFAULT', database)
```
robsdedude: For better maintainability, this logic should live somewhere central (close to all the other logic that's concerned with home db caching and cache keying).

(There's a copy of this line further down - won't comment again)

robsdedude (commented):
Leaving the comments of today so I don't loose them in draft mode. I'll continue the review tomorrow.

packages/core/src/driver.ts
```diff
@@ -108,6 +117,7 @@ interface DriverConfig {
   logging?: LoggingConfig
   notificationFilter?: NotificationFilter
   connectionLivenessCheckTimeout?: number
+  user?: string | undefined
```
robsdedude: Is this the public driver config type? What is this new config option used for?

packages/core/src/connection-provider.ts
```diff
@@ -57,15 +57,18 @@ class ConnectionProvider {
    * @property {Bookmarks} param.bookmarks - the bookmarks to send to routing discovery
    * @property {string} param.impersonatedUser - the impersonated user
    * @property {function (databaseName:string?)} param.onDatabaseNameResolved - Callback called when the database name get resolved
+   * @property {function (databaseName:string?)} param.removeFailureFromCache - Callback for deleting lost db from cache
```
robsdedude: Documented arg type `string?` does not match the declared type below `string`. Which on is the intended way?

packages/core/src/connection-provider.ts
```diff
@@ -57,15 +57,18 @@ class ConnectionProvider {
...
   acquireConnection (param?: {
     accessMode?: string
     database?: string
     bookmarks: bookmarks.Bookmarks
     impersonatedUser?: string
-    onDatabaseNameResolved?: (databaseName?: string) => void
+    onDatabaseNameResolved?: (database: string) => void
+    removeFailureFromCache?: (database: string) => void
     auth?: AuthToken
+    homeDb?: any
```
robsdedude: Can a narrower type be used? Maybe `string?` or whatever is appropriate?

packages/core/src/driver.ts
```diff
@@ -55,6 +56,12 @@ const DEFAULT_MAX_CONNECTION_LIFETIME: number = 60 * 60 * 1000 // 1 hour
  */
 const DEFAULT_FETCH_SIZE: number = 1000
 
+/**
+ * The maximum number of entries allowed in the home database cache before pruning
+ * @type {number}
```
robsdedude: I see this is a copy pasta :page_facing_up::spaghetti:. Why declare the type of the const _and_ put the type in the comment? Is there any benefit or is this just introducing the potential for out-of-sync information duplication?

packages/core/src/session.ts
```diff
@@ -101,7 +104,9 @@ class Session {
     bookmarkManager,
     notificationFilter,
     auth,
-    log
+    log,
+    homeDatabaseCallback,
```
robsdedude: Doc comment missing new (and some old :melting_face:) parameters.

packages/bolt-connection/test/bolt/bolt-protocol-v5x8.test.js
MaxAake: You're entirely correct. These lines were left un-bumped since their introduction in 5.6 driver testing bolt 5.3, then when creating the bolt 5.7 test I ended up bumping them. I'll correct this for other bolt tests in a separate PR to not clutter this one.

packages/bolt-connection/src/connection-provider/connection-provider-routing.js
```diff
@@ -139,32 +142,50 @@ export default class RoutingConnectionProvider extends PooledConnectionProvider
...
+      },
+      (error, address) => {
+        if (removeFailureFromCache !== undefined) {
+          removeFailureFromCache(homeDb ?? context.database)
+        }
+        return this._handleWriteFailure(error, address, homeDb ?? context.database)
+      },
       (error, address, conn) =>
         this._handleSecurityError(error, address, conn, context.database)
     )
```
robsdedude: I didn't see any mention of this in the ADR. Do we want to remove a cache entry on connection failure? Pros/cons? Either way, uniform behavior is desirable.

packages/core/src/auth.ts
robsdedude, line 1: I don't like this approach for 2 reasons:
 * it affects the public API for auth tokens (in multiple ways)
   * if users where using a `cacheKey` property, this will clash. I don't see a good use-case for this, so I think it's unlikely, but still possible
   * users that build their own custom auth objects (not using one of these factory methods) would have to change their code *and* understand what a correct cache value looks like for their custom auth tokens to work properly with the cache
 * the chosen cache keys are not collision free. Please find an injective mapping `TOKENS → CACHE_KEYS`. Where any custom token ∈ `TOKENS`.

robsdedude, line 1: The bolt server passes pretty much anything in the auth token untouched further down the DBMS stack for authentication. Custom fields are being ingested. I don't know if they can be used by custom auth plugins, I think not, but I'm not 100% sure on this. I'm gravitating towards not introducing specially handled auth token entries (like `cacheKey` [here](https://github.com/neo4j/neo4j-javascript-driver/pull/1235/files#diff-7860614c4e10f84a509cc8be4dd6dd7a1ad213dedc105e2363956aee3e2c597eR87-R88)). But this is more of an overly cautious future-proofing gut feeling. So maybe you want to ignore this comment :upside_down_face:

robsdedude (commented):
Next batch of comments. Sorry the PR is quite extensive :sweat_smile:. So I'll continue my review after the holidays :christmas_tree:. Have a lovely time :snowboarder:!

packages/bolt-connection/src/connection/connection-channel.js
MaxAake: It is not. The check came before the default value and weren't removed when it was added.

packages/core/src/driver.ts
MaxAake: Very good catch! This is a remnant of older spike work that should absolutely not still be there

packages/neo4j-driver-deno/lib/core/internal/connection-holder.ts
```diff
@@ -161,9 +166,9 @@ class ConnectionHolder implements ConnectionHolderInterface {
     return this._referenceCount
   }
 
-  initializeConnection (): boolean {
+  initializeConnection (homeDatabaseTable?: any): boolean {
```
robsdedude:
```suggestion
  initializeConnection (homeDatabaseTable?: string): boolean {
```
At least that's what `_createConnectionPromise` expects and I don't see any narrowing type checks or such here.

packages/bolt-connection/src/bolt/bolt-protocol-v5x8.js
MaxAake: It really isn't. Every 5.x release has been overriding with it's own transformer, despite no changes having been made, so effectively they're all using the same 5.0 transformer. Then at the jump to 5.6 the import wasn't bumped. 

Looking at it I'm leaning towards not creating new transformer versions with every release and cleaning up empty ones, but in this PR I will just fix the last 3 bolt versions to create transformers of their own version, and leave the larger task to another PR.

packages/core/src/internal/connection-holder.ts
```diff
@@ -172,14 +177,16 @@ class ConnectionHolder implements ConnectionHolderInterface {
     return true
   }
 
-  private async _createConnectionPromise (connectionProvider: ConnectionProvider): Promise<Connection & Releasable | null> {
+  private async _createConnectionPromise (connectionProvider: ConnectionProvider, homeDatabase?: string): Promise<Connection & Releasable | null> {
     return await connectionProvider.acquireConnection({
       accessMode: this._mode,
-      database: this._database,
+      database: this._database ?? '',
```
robsdedude: To make sure I've not missed something: is this change necessary?

I see that `ConnectionHolder.setDatabase` would accept and set `_database` to `null` (which , but I don't think was was happening before or is happening now. I think `_database` should maybe be typed `string`, not `?: string`.

packages/bolt-connection/src/connection-provider/connection-provider-routing.js
robsdedude: Note to myself: should we decide to go with this approach, double check the argument passed to `removeFailureFromCache` and convince yourself that this is the right cache entry to invalidate.

packages/core/src/internal/homedb-cache.ts
```diff
@@ -0,0 +1,98 @@
...
+  maxSize: number
+  map: Map<string, HomeDatabaseEntry>
+
+  constructor (maxSize: number) {
+    this.maxSize = maxSize
+    this.map = new Map()
+  }
+
+  /**
+   * Updates or add an entry to the cache, and prunes the cache if above the maximum allowed size
```
robsdedude:
```suggestion
   * Updates or adds an entry to the cache, and prunes the cache if above the maximum allowed size
```

packages/core/src/internal/homedb-cache.ts
```diff
@@ -0,0 +1,98 @@
...
+    })
+  }
+
+  /**
+   * Removes a number of the oldest entries in the cache if the number of entries has exceeded the maximum size.
+   */
+  private _pruneCache (): void {
+    if (this.map.size > this.maxSize) {
+      const sortedArray = Array.from(this.map.entries()).sort((a, b) => a[1].lastUsed - b[1].lastUsed)
+      for (let i = 0; i < 70; i++) { // c * max_size * logn(max_size), c is set to 0.01
```
robsdedude: I think it's quite likely that this will be forgotten should the `maxSize` ever be adjusted.

I suggest to either compute this dynamically in the constructor or to make sure the definition of the truncation size lives right next to the definition of the `maxSize` along with this comment.

packages/core/src/internal/homedb-cache.ts
robsdedude, line 1: Unit tests? :mag:

packages/core/src/session.ts
```diff
@@ -517,6 +533,24 @@ class Session {
...
+  _beginDbCallback (database: string): void {
+    // eslint-disable-next-line
+    if (this._connectionHolderWithMode(this._mode).connectionProvider()?.SSREnabled()) { 
+      this._databaseGuess = database
+      if (!this._databaseNameResolved) {
+        if (this._homeDatabaseCallback != null) {
+          // eslint-disable-next-line
+          this._homeDatabaseCallback((this._impersonatedUser ? 'basic:' + this._impersonatedUser : undefined) ?? this._auth?.cacheKey ?? 'DEFAULT', database)
+        }
+        const normalizedDatabase = database ?? ''
```
robsdedude, line 529: Something doesn't line up here with the typing. This line implies that `database` can be `null` or `undefined` but the function signature states that it's `string`.

packages/core/src/session.ts
```diff
@@ -517,6 +533,24 @@ class Session {
     }
   }
 
+  _beginDbCallback (database: string): void {
+    // eslint-disable-next-line
+    if (this._connectionHolderWithMode(this._mode).connectionProvider()?.SSREnabled()) {
```
robsdedude: This is something we should probably discuss with the team to make sure the drivers are aligned: if SSR is not enabled (or we don't know because of a too old protocol version), do we still want to maintain the home db cache in the hopes that a DBMS upgrade occurs later and we're having a cache already warm and ready?

packages/core/src/session.ts
```diff
@@ -517,6 +533,24 @@ class Session {
     }
   }
 
+  _beginDbCallback (database: string): void {
```
robsdedude: I think I'm missing something here: why is there a difference between `_onDatabaseNameResolved` and `_beginDbCallback`? Can't we always safely pin the resolved home db to the session regardless of SSR or not (cache updates discussion above pending)?

packages/core/src/session.ts
```diff
@@ -632,6 +669,7 @@ class Session {
  */
 function _createTransactionExecutor (config?: {
   maxTransactionRetryTime: number | null
+  commitCallback: any
```
robsdedude: Unused parameter?

packages/core/src/session.ts
robsdedude, line 1: Potential for optimization (dismiss if too much hassle to implement): `session.run` uses `afterComplete` to process the resolved home db. This means the cache only gets updated after all records have been streamed even though the information is already available earlier. This makes a difference if either the result stream dies/errors out or if multiple transactions are being run concurrently.

robsdedude: Looking at the code again, I'm unsure: does `afterComplete` only get to process the metadata of the last SUCCESS (i.e., the one without `has_more`)? In that case, this seems to be a left-over from an earlier design. The server echoes back the resolved home db on `RUN` `SUCCESS`. I assume I'm misreading the code... TestKit should've caught such thing :thinking:

packages/core/src/session.ts
MaxAake: The differing behavior was required when the COMMIT SUCCESS message was used, and the improvement was not made when we switched to the BEGIN SUCCESS. Good catch

packages/bolt-connection/src/connection-provider/connection-provider-routing.js
```diff
@@ -663,6 +682,28 @@ export default class RoutingConnectionProvider extends PooledConnectionProvider
...
+        this._withSSR = this._withSSR + 1
+      } else if (isEnabled === false) {
+        this._withoutSSR = this._withoutSSR + 1
+      }
+    } else if (action === 'CLOSE') {
+      if (isEnabled === true) {
+        this._withSSR = this._withSSR - 1
+      } else if (isEnabled === false) {
+        this._withoutSSR = this._withoutSSR - 1
+      }
```
robsdedude:
```suggestion
      if (isEnabled === true) {
        this._withSSR = this._withSSR + 1
      } else {
        this._withoutSSR = this._withoutSSR + 1
      }
    } else if (action === 'CLOSE') {
      if (isEnabled === true) {
        this._withSSR = this._withSSR - 1
      } else {
        this._withoutSSR = this._withoutSSR - 1
      }
```
Nit: I suggest to fall back to assuming no SSR on the connection if the server is sending garbage.
This is not a blocker though as I don't expect this to happen (requires a buggy server).

packages/core/src/session.ts
```diff
@@ -508,12 +523,18 @@ class Session {
...
+        if (this._homeDatabaseCallback != null) {
+          this._homeDatabaseCallback(cacheKey(this._auth, this._impersonatedUser), database)
+        }
+        const normalizedDatabase = database ?? ''
+        this._database = normalizedDatabase
+        this._readConnectionHolder.setDatabase(normalizedDatabase)
+        this._writeConnectionHolder.setDatabase(normalizedDatabase)
+        this._databaseNameResolved = true
+      }
     }
```
robsdedude:
```suggestion
    if (!this._isRoutingSession) {
      return
    }
    this._databaseGuess = database
    if (this._databaseNameResolved) {
      return
    }
    if (this._homeDatabaseCallback != null) {
      this._homeDatabaseCallback(cacheKey(this._auth, this._impersonatedUser), database)
    }
    const normalizedDatabase = database ?? ''
    this._database = normalizedDatabase
    this._readConnectionHolder.setDatabase(normalizedDatabase)
    this._writeConnectionHolder.setDatabase(normalizedDatabase)
    this._databaseNameResolved = true
```

Less indentation and mental nesting. However, more lines. A matter of taste for sure. Just wanted to raise the option.

packages/core/src/driver.ts
robsdedude: Aha! I just realized that this is valuable for the exported values as they're also available via plain JS where all the TS information has been stripped. But since this is not exposed, I believe removing it was the better call.

packages/core/src/session.ts
```diff
@@ -89,8 +91,11 @@ class Session {
    * @param {boolean} args.reactive - Whether this session should create reactive streams
    * @param {number} args.fetchSize - Defines how many records is pulled in each pulling batch
    * @param {string} args.impersonatedUser - The username which the user wants to impersonate for the duration of the session.
-   * @param {AuthToken} args.auth - the target auth for the to-be-acquired connection
+   * @param {BookmarkManager} args.bookmarkManager - The bookmark manager used for this session.
    * @param {NotificationFilter} args.notificationFilter - The notification filter used for this session.
+   * @param {AuthToken} args.auth - the target auth for the to-be-acquired connection
+   * @param {Logger} args.log - the logger used for logs in this session.
+   * @param {(impersonatedUser:string, user:string database:string) => void} args.homeDatabaseCallback - callback used to update the home database cache
```
robsdedude:
```suggestion
   * @param {(user: string, database: string) => void} args.homeDatabaseCallback - callback used to update the home database cache
```
Judging by the type below.

packages/core/src/internal/auth-util.ts
```diff
@@ -0,0 +1,48 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { AuthToken } from '../types'
+
+export function cacheKey (auth?: AuthToken, impersonatedUser?: string): string {
+  if (impersonatedUser !== undefined && impersonatedUser !== null) {
```
robsdedude:
```suggestion
  if (impersonatedUser != null) {
```

packages/core/src/internal/auth-util.ts
```diff
@@ -0,0 +1,48 @@
...
+      Object.keys(auth.parameters).sort().forEach((key: string) => {
+        // @ts-expect-error: undefined check is already made
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+        ordered += `${key}:${auth.parameters[key]}`
+      })
+    }
+    const credentialString = (auth.credentials !== undefined && auth.credentials !== '') ? 'credentials:' + auth.credentials : ''
+    const realmString = (auth.realm !== undefined && auth.realm !== '') ? 'realm:' + auth.realm : ''
+    return 'scheme:' + auth.scheme + 'principal:' + (auth.principal ?? '') + credentialString + realmString + 'parameters:' + ordered
+  }
```
robsdedude: Still not collision free. What if a key or value in `parameters` contains a `:`? What if `principal` is `"foo:credentials:bar"` while credentials is is empty?

packages/core/src/session.ts
MaxAake: Made a smaller change to this section to improve readability.

robsdedude (approved):
:muscle:

PR #1238, opened 2024-11-22
fixing minute offset issue in certain time zones

MaxAake:
Timezones with minute offsets not divisible by 60 causes issues for converting from the driver implementation of Date and Datetime to JS Date and Datetime


StephenCathcart (approved):
Looks good well documented 👍

PR #1243, opened 2024-12-19
ADR 30 : New handshake for the bolt protocol

MaxAake:
Implements the handshake v2 protocol, allowing for more granular selection of bolt protocol versions.

### DIFF
--- a/packages/bolt-connection/src/bolt/handshake.js
+++ b/packages/bolt-connection/src/bolt/handshake.js
@@ -19,7 +19,7 @@ import { alloc } from '../channel'
 import { newError } from 'neo4j-driver-core'
 
 const BOLT_MAGIC_PREAMBLE = 0x6060b017
-const AVAILABLE_BOLT_PROTOCOLS = [5.7, 5.6, 5.4, 5.3, 5.2, 5.1, 5.0, 4.4, 4.3, 4.2, 4.1, 4.0, 3.0] // bolt protocols the client supports, ordered by preference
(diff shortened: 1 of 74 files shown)


packages/bolt-connection/src/bolt/handshake.js
```diff
@@ -70,15 +72,63 @@ function parseNegotiatedResponse (buffer, log) {
...
+      buffer.readUInt8()
+    ]
+    versions = versions.concat(getVersions(versionRange))
+  }
+  const capabilityBitMask = buffer.readVarInt()
+  const capabilites = selectCapabilites(capabilityBitMask)
+
+  let major
+  let minor
+  versions.sort((a, b) => Number(b.major + '.' + b.minor) - Number(a.major + '.' + a.minor))
```
StephenCathcart: It may not be an issue, but have you checked this covers trailing zeros in the conversation?

So `Number(major + '.' + minor)` converts versions into floats (5.7 becomes 5.7 but 4.10 might not behave as expected?)

packages/bolt-connection/src/bolt/handshake.js
```diff
@@ -70,15 +72,63 @@ function parseNegotiatedResponse (buffer, log) {
...
+  for (let i = 0; i < versions.length; i++) {
+    const version = versions[i]
+    if (AVAILABLE_BOLT_PROTOCOLS.includes(Number(version.major + '.' + version.minor))) {
```
StephenCathcart, line 101: If no version matches, major and minor will be undefined, does this lead to any unexpected errors?

packages/bolt-connection/src/bolt/handshake.js
MaxAake: Good catch that I really should have realized myself! Thank you!

StephenCathcart (approved):
Nice job 👍

PR #1245, opened 2025-01-20
Minor correction in bolt protocol tests

MaxAake:
Changes the driver version in bolt protocol tests to be the version where the protocol was added.

Not really necessary, but it doesn't hurt to be correct!


StephenCathcart (approved):
Looks good!

PR #1247, opened 2025-01-24
Better error handling for circular references in custom auth tokens

MaxAake:
It is quite easy for a user to provide an object with a circular reference in the parameters of a custom auth token, despite this being impossible to send over Bolt. The current handling for this error is not very clear, this PR ensures an understandable error is thrown


StephenCathcart (approved):
Nice fix 👍

PR #1249, opened 2025-02-05
Update example code in Readme

MaxAake:
Replaces deprecated function use and moves session.run usage examples down to not be the top example.

Additionally removes linter for the readme, as it's not useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,106 @@ var rxSession = driver.rxSession({
 })
 ```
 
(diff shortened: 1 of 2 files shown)


AndyHeap-NeoTech (approved):
Looks good. I have one question for going forward though. Should we move it away from session.Run to the tx functions? I'm generally wary of using session.run in examples, even with it being so simple. Open to thoughts.

PR #1250, opened 2025-02-05
Disable missbehaving esdocs plugin

MaxAake:
esdoc-type-inference-plugin is a proof of concept plugin causing an error when building docs of the new version. disabling it solves the issue


robsdedude (approved):
🥾💥

PR #1253, opened 2025-02-11
Fix handshake manifest parsing of future bolt versions

MaxAake:
When checking the whitelist of available bolt protocols, the versions are parsed as floats, which leads to the future version 5.10 being marked as "available" since it is seen as the same as 5.1.

The fix is simple and makes the code simpler, treating them as strings in this step instead.


robsdedude (approved):
`"✍️"`

PR #1254, opened 2025-02-12
Correct list of possible timeout error codes.

MaxAake:
The error code ```Neo.ClientError.Transaction.TransactionTimedOutClientConfiguration``` was neglected when listing the possible transaction timeout error messages in one test, this causes some flakiness in integration tests as it is sometimes but not always the resulting error. It was previously correctly added to one test, but not another.


AndyHeap-NeoTech (approved):
Looks good.

PR #1256, opened 2025-02-13
Extent timeout on lite driver tests

MaxAake:
A test on the lite driver is intermittently timing out with the default 5 second timeout, this PR doubles the timeout.


robsdedude (approved):
:hourglass_flowing_sand:

PR #1260, opened 2025-02-19
configure timeout example test has too short timeout

MaxAake:
Configure Timeout example sets the transaction timeout to 10ms, this causes flakiness because it's too low of a value.


AndyHeap-NeoTech (approved):
Math checks out, 100 is indeed bigger than 10 :)

PR #1262, opened 2025-02-25
Skip new summary tests on RX session

MaxAake:
The RX session handles sending DISCARDS slightly differently than other drivers, which breaks some new tests being added to testkit. This skips those new tests on the rx driver until a more elegant solution is found

### DIFF
--- a/packages/testkit-backend/src/skipped-tests/rx.js
+++ b/packages/testkit-backend/src/skipped-tests/rx.js
@@ -1,9 +1,13 @@
-import { skip, ifEquals } from './skip.js'
+import { skip, ifEquals, ifStartsWith } from './skip.js'
 
 const skippedTests = [
   skip(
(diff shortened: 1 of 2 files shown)


robsdedude:
Leaving this here for traceability https://github.com/neo4j-drivers/testkit/pull/653

packages/testkit-backend/src/skipped-tests/rx.js
robsdedude, line 1: Smells like a regex, glob matcher, or similar predicate could simplify the code a little. No blocker though.

robsdedude (approved):

packages/testkit-backend/src/skipped-tests/rx.js
MaxAake: I like the idea, but it'd take a rewrite of the skipping part that would make this significantly bigger PR. Worth putting on my backlog, though!

PR #1264, opened 2025-03-06
Fix typo in API docs.

stefano-ottolenghi:
Fixes this nasty, _nasty_ visual bug.

Screenshot from 2025-03-06 07-28-19

I've raised this to `5.0` as it is still the default branch. I assume it needs to be cherry-picked forward too.


MaxAake (approved):
💸 Fly away now, dirty dollar!

Good catch! I'll make sure this goes into 6.x as well

PR #1265, opened 2025-03-21
Fix inconsistent typing of duration properties

MaxAake:
When creating a Duration the seconds and nanoseconds will be converted to Integers, no matter if provided with Integers, Numbers, or BigInts. This is awkward and requires special handling by users

### DIFF
--- a/packages/core/src/integer.ts
+++ b/packages/core/src/integer.ts
@@ -1101,6 +1101,13 @@ const inSafeRange = Integer.inSafeRange
 const toNumber = Integer.toNumber
 
 /**
+ * Converts a variable to a BigInt
+ * @access public
+ * @param {Mixed} value - The variable to convert
+ * @return {BigInt} - the variable as a BigInt
+ */
+const toBigInt = Integer.valueOf
+/**
  * Converts the integer to a string representation
  * @access public
  * @param {Mixed} value - The variable to convert
@@ -1109,6 +1116,6 @@ const toNumber = Integer.toNumber
  */
 const toString = Integer.toString
 
-export { int, isInt, inSafeRange, toNumber, toString }
+export { int, isInt, inSafeRange, toNumber, toBigInt, toString }
 
 export default Integer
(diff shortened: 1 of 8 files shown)


packages/core/src/integer.ts
```diff
@@ -1100,6 +1100,13 @@ const inSafeRange = Integer.inSafeRange
  */
 const toNumber = Integer.toNumber
 
+/**
+ * Converts a variable to a BigInt
+ * @access public
+ * @param {Mixed} value - The variable to convert
```
robsdedude:
```suggestion
 * @param {!Integer} value - The variable to convert
```

At least if I understood the code right. What's the purpose of exposing this method in the first place? It only takes Integer and is also bound to Integer. So if the user has an Integer, they can just call `int.valueOf` on it. Or am I missing the point?

packages/core/src/temporal-types.ts
```diff
@@ -79,6 +79,14 @@ export class Duration<T extends NumberOrInteger = Integer> {
      * @type {NumberOrInteger}
      */
     this.nanoseconds = util.normalizeNanosecondsForDuration(nanoseconds) as T
+    if (typeof this.months === 'number' && isInt(this.nanoseconds) && isInt(this.seconds)) {
```
robsdedude: What if Duration is constructed with `Duration(someNumber, someInteger, someBigint)`. I mean it's really ugly and the types state that this is not supported. But 
 1. In JS (not TS), nothing is strongly typed
 2. This if is already covering the use-case of mixing different number types.

packages/core/src/integer.ts
MaxAake: You're entirely correct.

packages/core/src/temporal-types.ts
MaxAake: It's not the biggest issue, since the packer would handle that just fine, but there's always a change that a user does something odd and creates a duration object for their own use somewhere. I'm making a change so that the normalized seconds and nanos will keep the same type as the unnormalized values.

robsdedude (approved):
Optional nit: the fact that your last commits changed behavior while all tests stayed the same (and passed before and still do) suggests that test coverage could be improved :)

PR #1269, opened 2025-04-14
4.4 Document that concurrently closing drivers is a bad idea.

MaxAake:


robsdedude (approved):
:cherries::older_man:

PR #1273, opened 2025-04-25
Update testkit dockerfile

MaxAake:
Updates dockerfile ubuntu and sets node 18 as the default

### DIFF
--- a/testkit/Dockerfile
+++ b/testkit/Dockerfile
@@ -1,10 +1,10 @@
-FROM ubuntu:20.04
+FROM ubuntu:24.04
 
-ARG NODE_VERSION=10
-ARG DENO_VERSION=1.37.2
+ARG NODE_VERSION=18
+ARG DENO_VERSION=2.2.11
 
-ENV DEBIAN_FRONTEND noninteractive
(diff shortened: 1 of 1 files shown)


testkit/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN apt-get update && \
     apt-get upgrade -y && \
     apt-get install -y curl
 
-RUN curl -sL https://deb.nodesource.com/setup_${NODE_VERSION:=10}.x | bash -
+RUN curl -sL https://deb.nodesource.com/setup_${NODE_VERSION:=18}.x | bash -
```
robsdedude: I see that this PR didn't change this, but I wonder: what does `:=` in `${VAR=:foo}` do? I couldn't find that syntax in the docs https://docs.docker.com/reference/dockerfile/#environment-replacement

robsdedude (approved):

testkit/Dockerfile
MaxAake: It appears to be an undocumented alias of `:-`. I tested it locally and it functioned the same, for clarity I will change the to the documented version.

It is arguably a bit redundant to have both an arg with a default and a fallback value, but I see no reason to remove either.

PR #1276, opened 2025-05-05
Fix testkit import for node 20+

MaxAake:
When running the testkit backend on node 20 or later the lax imports are not accepted, this PR is functionally identical, but accepted by the newer node


StephenCathcart (approved):
🚢

PR #1277, opened 2025-05-07
Update firefox install in testkit docker

MaxAake:
Newer version of ubuntu install a snap-dependent version of firefox via apt-get. This gets a Docker-compatible version from firefox.


StephenCathcart (approved):
🚢

PR #1278, opened 2025-05-08
Fix logging of driver errors containing bigints in testkit backend

MaxAake:
The JavaScript driver testkit backend handles all ints as bigints, when these end up in some errors (as in syntax errors on 2025.x servers) they cause the backend to crash. This PR solves the issue by using a replacer function.


robsdedude (approved):
:adhesive_bandage:

PR #1281, opened 2025-05-12
Update skipped temporal type test names

MaxAake:
A few temporal type stub tests were moved on the 6.x branch, 4 of these were skipped on JS and this skip check needs to be updated.


robsdedude (approved):
:ski:

PR #1282, opened 2025-05-12
Fix checks for deno v2

MaxAake:
Bumping the build dockerfile to use docker v2 enabled a few checks by default, this ignores two lines where violation is unavoidable, and disables another check that would involve a larger rework that is not a priority at this time.


AndyHeap-NeoTech (approved):
LGTM

PR #1284, opened 2025-05-19
Unify Duration String Representation With Cypher

MaxAake:
Simplifies the stringified representation of Durations to display 12 months as a year, and seconds into minutes and hours to align with Cypher behavior as below

`RETURN toString(duration({months: 13}))` -> `"P1Y1M"`
`RETURN toString(duration({minutes: 61}))` -> `"PT1H1M"`
`RETURN toString(duration({seconds: 61}))` -> `"PT1M1S"`


StephenCathcart (approved):
👍

PR #1287, opened 2025-05-20
Deprecate and Stabilize Result Transformer Aliases

MaxAake:


packages/core/src/result-transformers.ts
```diff
@@ -52,7 +52,8 @@ class ResultTransformers {
    * const { keys, records, summary } = await driver.executeQuery('CREATE (p:Person{ name: $name }) RETURN p', { name: 'Person1'})
    *
    * @returns {ResultTransformer<EagerResult<Entries>>} The result transformer
-   * @alias {@link ResultTransformers#eager}
+   * @deprecated This is deprecated, use {@link ResultTransformers#eager} instead
+   * @since 6.0.0
```
robsdedude: Reading https://jsdoc.app/tags-since and https://jsdoc.app/tags-deprecated it seems that `@since` is not for marking since when something is deprecated. Consider including that information after the `@deprecated` tag instead.

Same for `#mappedResultTransformer`, won't comment again.

packages/core/src/result-transformers.ts
```diff
@@ -52,7 +52,8 @@ class ResultTransformers {
    * const { keys, records, summary } = await driver.executeQuery('CREATE (p:Person{ name: $name }) RETURN p', { name: 'Person1'})
    *
    * @returns {ResultTransformer<EagerResult<Entries>>} The result transformer
-   * @alias {@link ResultTransformers#eager}
+   * @deprecated This is deprecated, use {@link ResultTransformers#eager} instead
```
robsdedude: Might be worth mentioning that `#eager` is a drop-in replacement/alias/exactly the same.

Similarly for `#mapped`, won't comment again.

packages/core/src/result-transformers.ts
```diff
@@ -74,9 +74,6 @@ class ResultTransformers {
    * const { keys, records, summary } = await driver.executeQuery('CREATE (p:Person{ name: $name }) RETURN p', { name: 'Person1'})
    *
    * @returns {ResultTransformer<EagerResult<Entries>>} The result transformer
-   * @experimental this is a preview
-   * @since 5.22.0
```
robsdedude, line 78: Is this info no longer valuable? (more `@since` removals in patch & `#summary` didn't have a `@since` to begin with, won't comment again)

packages/core/src/result-transformers.ts
```diff
@@ -91,7 +88,7 @@ class ResultTransformers {
    * @example
    * // Mapping the records
    * const { keys, records, summary } = await driver.executeQuery('MATCH (p:Person{ age: $age }) RETURN p.name as name', { age: 25 }, {
-   *   resultTransformer: neo4j.resultTransformers.mappedResultTransformer({
+   *   resultTransformer: neo4j.resultTransformers.mapped({
```
robsdedude: Not sure the documentation for `mappedResultTransformer` should show how to use `mapped` :upside_down_face: (same change 2 more times, won't comment again).

MaxAake: Yes. Obviously I was overeager when it came to purging the old names from documentation 🙃

packages/core/src/result-transformers.ts
MaxAake: I aligned with `#summary` which didn't have a tag, but looking at it again I think its better to keep the tags and add the missing tag to summary. Thank you!

packages/core/src/result-transformers.ts
```diff
@@ -91,7 +88,7 @@ class ResultTransformers {
    * @example
    * // Mapping the records
    * const { keys, records, summary } = await driver.executeQuery('MATCH (p:Person{ age: $age }) RETURN p.name as name', { age: 25 }, {
-   *   resultTransformer: neo4j.resultTransformers.mappedResultTransformer({
+   *   resultTransformer: neo4j.resultTransformers.mappedResultTransformer ({
```
robsdedude:
```suggestion
   *   resultTransformer: neo4j.resultTransformers.mappedResultTransformer({
```
Mind the gap :wink:

packages/core/src/result-transformers.ts
```diff
@@ -103,7 +100,7 @@ class ResultTransformers {
    * @example
    * // Mapping records and collect result
    * const names = await driver.executeQuery('MATCH (p:Person{ age: $age }) RETURN p.name as name', { age: 25 }, {
-   *   resultTransformer: neo4j.resultTransformers.mappedResultTransformer({
+   *   resultTransformer: neo4j.resultTransformers.mappedResultTransformer ({
```
robsdedude:
```suggestion
   *   resultTransformer: neo4j.resultTransformers.mappedResultTransformer({
```

packages/core/src/result-transformers.ts
```diff
@@ -117,7 +114,7 @@ class ResultTransformers {
    *
    * @example
    * // The transformer can be defined one and used everywhere
-   * const getRecordsAsObjects = neo4j.resultTransformers.mappedResultTransformer({
+   * const getRecordsAsObjects = neo4j.resultTransformers.mappedResultTransformer ({
```
robsdedude:
```suggestion
   * const getRecordsAsObjects = neo4j.resultTransformers.mappedResultTransformer({
```

robsdedude (approved):
:speech_balloon: :arrows_counterclockwise:

PR #1288, opened 2025-05-23
Bump Typescript and Replace neo4j-driver Browser Integration Test Framework (#1288)

MaxAake:
Bumps the used version of typescript in the driver devDependencies. This breaks the usage of karma to run browser intregration tests in the neo4j-driver package, so the browser tests are migrated to jest, which forced a number of workaround changes as tests are now being run with two different test environments. 

Further improvement is desired, but this is a first step at moving away from the large number of deprecated dev dependencies currently in use.


StephenCathcart (approved):
👍

PR #1290, opened 2025-05-27
Replace unclear fromStandardDate()

MaxAake:
The `Neo4j.Date` functions `fromStandardDate()` and `toStandardDate()` are not inverse functions of each other, as `fromStandardDate()` takes the date that the StandardDate would correspond to in local time, and `toStandardDate()` creates a Date set at Midnight UTC.

This PR deprecates the old `fromStandardDate()` and replaces it with a drop-in replacement `fromStandardDateLocal()` which functions the same and `fromStandardDateUTC()` which takes the date as interpreted in UTC.

### DIFF
--- a/packages/core/src/temporal-types.ts
+++ b/packages/core/src/temporal-types.ts
@@ -352,7 +352,7 @@ export class Date<T extends NumberOrInteger = Integer> {
 
   /**
    * Create a {@link Date} object from the given standard JavaScript `Date`.
-   * Hour, minute, second, millisecond and time zone offset components of the given date are ignored.
+   * Hour, minute, second and millisecond components of the given date are ignored.
    *
    * NOTE: the function {@link toStandardDate} and {@link fromStandardDate} are not inverses of one another. {@link fromStandardDate} takes the Day, Month and Year in local time from the supplies JavaScript Date object, while {@link toStandardDate} creates a new JavaScript Date object at midnight UTC. This incongruity will be rectified in 6.0
    * If your timezone has a negative offset from UTC, creating a JavaScript Date at midnight UTC and converting it with {@link fromStandardDate} will result in a Date for the day before.
@@ -367,7 +367,7 @@ export class Date<T extends NumberOrInteger = Integer> {
 
   /**
    * Create a {@link Date} object from the given standard JavaScript `Date` using the Year, Month and Date in Local Time.
-   * Hour, minute, second, millisecond and time zone offset components of the given date are ignored.
+   * Hour, minute, second and millisecond components of the given date are ignored.
    *
    * NOTE: this function and {@link toStandardDate} are not inverses of one another.
    * This takes the Day, Month and Year in local time from the supplies JavaScript Date object, while {@link toStandardDate} creates a new JavaScript Date object at midnight UTC.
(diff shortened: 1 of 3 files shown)


packages/core/src/temporal-types.ts
```diff
@@ -359,8 +359,24 @@ export class Date<T extends NumberOrInteger = Integer> {
...
   static fromStandardDate (standardDate: StandardDate): Date<number> {
+    return this.fromStandardDateLocal(standardDate)
+  }
+
+  /**
+   * Create a {@link Date} object from the given standard JavaScript `Date` using the Year, Month and Date in Local Time.
+   * Hour, minute, second, millisecond and time zone offset components of the given date are ignored.
+   *
+   * NOTE: this function and {@link toStandardDate} are not inverses of one another.
+   * This takes the Day, Month and Year in local time from the supplies JavaScript Date object, while {@link toStandardDate} creates a new JavaScript Date object at midnight UTC.
```
robsdedude:
```suggestion
   * This takes the Day, Month and Year in local time from the supplied JavaScript Date object, while {@link toStandardDate} creates a new JavaScript Date object at midnight UTC.
```
I see this is a copy-pasta, so consider fixing the existing doc comment as well.

packages/core/src/temporal-types.ts
robsdedude, line 1: My brain was close to melting while trying to understand this PR :melting_face: 

I think example code would go a long way. Something like this (maybe split the examples and spread them across the functions' docs)

```js
// will do what it says on the tin
Neo4jDate.fromStandardDateLocal(new Date(2020, 01, 01))
Neo4jDate.fromStandardDateUTC(new Date(Date.UTC(2020, 01, 01)))

// Depending on the host's locale, might end up with a different calendar date
// than 2020-01-01
Neo4jDate.fromStandardDateLocal(new Date(Date.UTC(2020, 01, 01)))
Neo4jDate.fromStandardDateUTC(new Date(2020, 01, 01))
```

packages/core/test/temporal-types.test.ts
robsdedude, line 1: Optional: I think it might be very much worthwhile to run these tests for different time zones/locales. In particular the ones that are far off of UTC (and maybe also UTC itself).

https://stackoverflow.com/questions/8083410/how-can-i-set-the-default-timezone-in-node-js

robsdedude (approved):
:fire: :watch: :no_good:

:pray: :peace_symbol:

packages/core/src/temporal-types.ts
```diff
@@ -370,14 +390,33 @@ export class Date<T extends NumberOrInteger = Integer> {
     )
   }
 
+  /**
+   * Create a {@link Date} object from the given standard JavaScript `Date` using the Year, Month and Date in UTC time.
+   * Hour, minute, second and millisecond components of the given date are ignored.
+   *
+   * @example
+   * fromStandardDateLocal(new Date("2010-10-10T00:00:00")) // This may cause issues as JS Dates are created at local time by default
+   * fromStandardDateLocal(new Date("2010-10-10T00:00:00Z")) // Will create a date at 2010-10-10 as this date is created at UTC with the trailing "Z"
```
robsdedude:
```suggestion
   * fromStandardDateUTC(new Date("2010-10-10T00:00:00")) // This may cause issues as JS Dates are created at local time by default
   * fromStandardDateUTC(new Date("2010-10-10T00:00:00Z")) // Will create a date at 2010-10-10 as this date is created at UTC with the trailing "Z"
```

PR #1291, opened 2025-06-02
Require Node 18 for drivers 6.x

MaxAake:
This will permit the merging of the queryAPI wrapper into the main driver repo without a sudden breaking change in the middle of 6.x


StephenCathcart (approved):
👍

PR #1292, opened 2025-06-04
Unify Connection Acquisition Timeout with other Drivers

MaxAake:
we’ve decided upon unifying the drivers to consider the timeout to count towards **everything** needed for acquiring a connection. This can include multiple connection acquisitions, ROUTE request, custom resolver/provider calls, etc; anything that’s needed to get a connection for the session.

tested against testkit branch: https://github.com/neo4j-drivers/testkit/pull/674


StephenCathcart (approved):
👍

PR #1293, opened 2025-06-10
Bolt 6.0 and Vector Type

MaxAake:
Introduces the Neo4j Vector type, a wrapper for JavaScript TypedArrays, allowing Vectors to be sent to and from the database.

To support these, the bolt protocol version 6.0 is introduced.


README.md
```diff
@@ -503,3 +503,29 @@ var driver = neo4j.driver(
   { disableLosslessIntegers: true }
 )
 ```
+
+#### Writing and reading Vectors
+
+Neo4j supports storing vector embddings in a dedicated vector type. Sending large lists with the driver will result in significant overhead as each value will be transmitted with type information, so the 6.0.0 release of the driver introduced the neo4j Vector type.
```
StephenCathcart:
```suggestion
Neo4j supports storing vector embeddings in a dedicated vector type. Sending large lists with the driver will result in significant overhead as each value will be transmitted with type information, so the 6.0.0 release of the driver introduced the Neo4j Vector type.
```

packages/bolt-connection/src/bolt/bolt-protocol-v6x0.transformer.js
```diff
@@ -0,0 +1,133 @@
...
+  FLOAT32: FLOAT_32,
+  FLOAT64: FLOAT_64
+}
+
+function createVectorTransformer () {
+  return new TypeTransformer({
+    signature: VECTOR,
+    isTypeInstance: object => object instanceof Vector,
+    toStructure: vector => {
+      const typeMarker = typeToTypeMarker[vector.type]
```
StephenCathcart: Question - is it possible for `vector.type` to not match any markers and return undefined instead, and if so does this need a guard/error if that's the case?

packages/bolt-connection/test/bolt/bolt-protocol-v1.test.js
```diff
@@ -444,6 +446,19 @@ describe('#unit BoltProtocolV1', () => {
       expect(unpacked).toBeInstanceOf(Object)
       expect(unpacked).toMatchSnapshot()
     })
+
+    it('Should error out when trying to pack a Vector', () => {
```
StephenCathcart:
```suggestion
    it('should error out when trying to pack a Vector', () => {
```

packages/bolt-connection/src/bolt/bolt-protocol-v6x0.transformer.js
MaxAake: I'm fairly certain this would then error out at the default case in the switch statement of `fixBufferEndianness`, but it would not be the most elegant error message, since it would say the type of the vector is "undefined". It would require the user to have manually changed the type of their vector, but it's something that could happen, I'll take some time to find a good way to throw here!

packages/bolt-connection/test/bolt/bolt-protocol-v6x0.test.js
```diff
@@ -0,0 +1,1606 @@
...
+        mode: WRITE,
+        impersonatedUser
+      })
+    )
+    expect(protocol.messages[1]).toBeMessage(RequestMessage.pull())
+    expect(protocol.observers).toEqual([observer, observer])
+    expect(protocol.flushes).toEqual([false, true])
+  })
+
+  it('should begin a transaction', () => {
```
StephenCathcart, line 196: Might be nice to rename this slightly to avoid confusion with the test on line 396 that is also called "should begin a transaction"

packages/core/src/vector.ts
```diff
@@ -0,0 +1,85 @@
...
+  FLOAT32: 'FLOAT32',
+  FLOAT64: 'FLOAT64'
+}
+Object.freeze(vectorTypes)
+
+/**
+ * A wrapper class for JavaScript TypedArrays that makes the driver send them as a Vector type to the database.
+ * @access public
+ * @exports Vector
+ * @class A Integer class for representing a 64 bit two's-complement integer value.
```
StephenCathcart: Does this class description need updating for Vector?

StephenCathcart (approved):
👍  great just minor suggestions

packages/core/src/vector.ts
MaxAake: It sure does! missed that bit of copy-pasta 🍝

packages/bolt-connection/test/bolt/bolt-protocol-v6x0.test.js
MaxAake: I agree, that'll be done in a separate PR though, as the bolt protocols all share that test and I'll want to rename it for the former versions as well, and I'd rather not pollute this PR.

PR #1295, opened 2025-06-27
Fix acquisition timeout unification

MaxAake:
The original solution would still allow minor pieces of work to slip through before a timeout fired.


StephenCathcart (approved):
👍

PR #1300, opened 2025-07-14
Fix undocumented members on GQLErrors and Neo4jErrors

MaxAake:
In addition, this also fixes the exporting of isRetriableError, which was incorrectly exported as isRetryableError (breaking it)


StephenCathcart (approved):
👌

PR #1301, opened 2025-07-14
Vector Integration test error

MaxAake:
This integration test had not been run since the server added support for 6.0 when the branch was merged, which led to a an error in the code slipping through

### DIFF
--- a/packages/core/src/vector.ts
+++ b/packages/core/src/vector.ts
@@ -59,6 +59,13 @@ export default class Vector<K extends Float32Array | Float64Array | Int8Array | 
       this._type = vectorTypes.FLOAT32
     } else if (typedArray instanceof Float64Array) {
       this._type = vectorTypes.FLOAT64
+    } else if (
+      typedArray instanceof Uint8Array ||
+      typedArray instanceof Uint16Array ||
+      typedArray instanceof Uint32Array ||
+      typedArray instanceof BigUint64Array
+    ) {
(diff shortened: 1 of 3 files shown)


packages/core/src/vector.ts
```diff
@@ -59,6 +59,13 @@ export default class Vector<K extends Float32Array | Float64Array | Int8Array |
       this._type = vectorTypes.FLOAT32
     } else if (typedArray instanceof Float64Array) {
       this._type = vectorTypes.FLOAT64
+    } else if (
+      typedArray instanceof Uint8Array ||
+      typedArray instanceof Uint16Array ||
+      typedArray instanceof Uint32Array ||
+      typedArray instanceof BigUint64Array
```
robsdedude, line 66: There's also `Float16Array` and `Uint8ClampedArray`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray

MaxAake: Float16Array does not have node support, so I cannot do a instanceof check for it. It will still result in an error, but I cannot make a specific one for it. The clampedArray is a good catch, though

robsdedude: You could possibly check whether `globalThis["Float16Array"]` is defined or similar. :shrug: no strong opinion here though.

```js
... || (globalThis["Float16Array"] != null && typedArray instanceof globalThis["Float16Array"])
```

robsdedude (approved):
:arrow_left:

PR #1303, opened 2025-07-15
Set sideEffects false in all npm package jsons

MaxAake:
This will allow for tree shaking to remove dead code from the neo4j packages in user code.


StephenCathcart (approved):
👍

PR #1305, opened 2025-07-15
Deprecate isRetriableError and ensure isRetryableError functions

MaxAake:
deprecates the old isRetriableError and introduces isRetryableError in it's place. also changes internal names with "retriable" to "retryable"


packages/core/src/error.ts
```diff
@@ -211,23 +219,44 @@ class Neo4jError extends GQLError {
...
+
+    /**
+     * If the error is considered retryable.
+     * This does not apply when running auto-commit transactions using {@link Session#run}
+     *
      * @type {boolean}
      * @public
      */
-    this.retriable = _isRetriableCode(code)
+    this.retryable = _isRetryableCode(code)
```
robsdedude: minor optimization potential
```typescript
    let isRetryableCode = _isRetryableCode(code)

    /* ... [docs] ... */
    this.retriable = isRetryableCode

    /* ... [docs] ... */
    this.retryable = isRetryableCode
```

packages/bolt-connection/test/connection-provider/connection-provider-direct.test.js
```diff
@@ -205,7 +205,7 @@ it('should call authenticationAuthProvider.handleError when TokenExpired happens
   expect(handleError).toBeCalledWith({ connection: conn, code: 'Neo.ClientError.Security.TokenExpired' })
 })
 
-it('should change error to retriable when error when TokenExpired happens and staticAuthTokenManager is not being used', async () => {
+it('should change error to retriyable when error when TokenExpired happens and staticAuthTokenManager is not being used', async () => {
```
robsdedude: Best of both worlds, eh? :smile: 

```suggestion
it('should change error to retryable when error when TokenExpired happens and staticAuthTokenManager is not being used', async () => {
```

packages/neo4j-driver/src/index.js
```diff
@@ -20,6 +20,7 @@ import VERSION from './version'
 import {
   authTokenManagers,
   Neo4jError,
+  isRetriableError,
```
robsdedude, line 23: Just to double-check:

This change exports a deprecated API that wasn't exported before. Generally undesirable. We do it anyhow to achieve API uniformity across the different driver bundles/packages.

Did I get that right?

packages/bolt-connection/test/connection-provider/connection-provider-direct.test.js
```diff
@@ -222,10 +222,10 @@ it('should change error to retriable when error when TokenExpired happens and st
 
   const error = conn.handleAndTransformError(expectedError, address)
 
-  expect(error.retriable).toBe(true)
+  expect(error.retryable).toBe(true)
```
robsdedude, line 226: Up to you: might be worth testing both (deprecated and renamed) attributes. (more places this applies to, won't comment again)

packages/core/src/error.ts
```diff
@@ -211,23 +219,44 @@ class Neo4jError extends GQLError {
...
+    return this.isRetryable(error)
+  }
+
+  /**
+   * Verifies if the given error is retryable.
+   *
+   * @param {object|undefined|null} error the error object
+   * @returns {boolean} true if the error is retryable
+   */
+  static isRetryable (error?: any | null): boolean {
```
robsdedude, line 256: I can imagine user code that writes to `error.retriable`. With this PR the driver starts to basically ignore that field all together (users now have to write to `error.retryable` to achieve the same behavior). I'm not sure there's a legitimate use-case for such a thing or whether we'd want to support this at all, but it's a subtle breakage worth considering.

packages/neo4j-driver/src/index.js
robsdedude: Yeah ok. Just looked at https://github.com/neo4j/neo4j-javascript-driver/pull/1309#issue-3238371228 and the description answers my question :dotted_line_face:

MaxAake: Yes, that's correct. It's also done so that anyone currently importing from the core package (only way isRetriableError currently works) can migrate so all their imports are from neo4j-driver without any errors.

packages/core/src/error.ts
MaxAake: This is true and to some extent a breaking change. However I can't see any real use-case for writing to retriable, and I'm okay with such usage needing to do a small bit of migration since it's a major version release. The alternative would be to still use the value of retriable internally, but then the user would need to use the deprecated member for their code to work, which is worse to me.

robsdedude: Convinced :heavy_check_mark: 

Just make sure that the changelog, migration guide, etc. clearly highlights this subtle breakage.

robsdedude (approved):
:mortar_board: 
:teacher:

PR #1309, opened 2025-07-17
[5.x] Make neo4j-driver export isRetriableError

MaxAake:
Making isRetryableError work would cause the name to differ between neo4j-driver and neo4j-driver-lite, which would be a major headache. instead isRetriable is made to work properly in 5.x, and will be deprecated in 6.0.0

cherry picks: #1305


robsdedude (approved):
Probably not worth the hassle now, but as a future improvement: feels like a test should've caught a broken export.

PR #1313, opened 2025-07-29
rename begin transaction test to avoid duplicate names

MaxAake:


robsdedude (commented):
What about 

https://github.com/neo4j/neo4j-javascript-driver/blob/c2646f3eab772624ed350e6557feb0e17039fb7e/packages/bolt-connection/test/bolt/bolt-protocol-v4x0.test.js#L81

robsdedude (approved):
:broom:

PR #1321, opened 2025-08-21
Switch Bolt protocol version to dedicated type

MaxAake:
The selected bolt protocol version is represented as a floating point number in multiple places in the code. This is currently fine, but would cause the whole driver to explode if/when we introduce bolt 5.10 or 6.10. 

This PR addresses this tech debt before it causes a larger issue


packages/neo4j-driver-deno/lib/bolt-connection/bolt/create.js
```diff
@@ -112,8 +112,11 @@ function createProtocol (
   onProtocolError,
   log
 ) {
-  switch (version) {
-    case 1:
+  if (!(version instanceof ProtocolVersion) || version === undefined || version === null) {
+    throw newError('Unknown Bolt protocol version: ' + version)
+  }
+  switch (version.toString()) {
```
AndyHeap-NeoTech: I'm not hugely experienced in JS, and I know that you can't override the equality operator like you can in some other languages. But is there a way to do the switch without string comparison? Something like:

```
switch (true) {
     case version.equals(BOLT_PROTOCOL_V1):
         break;
     case version.equals(BOLT_PROTOCOL_V5):
          break;
     default:
    // default case
}
```

No idea if that would be preferable or idiomatic in JS, but to my C# brain it's preferable.

AndyHeap-NeoTech (approved):
looks good

PR #1322, opened 2025-08-21
TestKit backend: truncate logging to avoid clogging ram

robsdedude:
Writing to stdout in nodejs is done asynchronously. So when adding data to the queue faster than it can be flushed, the process will eventually crash because the system is running out of memory. By limiting the length of log messages, this problem can be avoided.


MaxAake (approved):
✅ 🙇🏻

PR #1327, opened 2025-09-15
Vector type stringer unifiction and isVector function

MaxAake:
Resolves DRIVERS-51

### DIFF
--- a/packages/core/src/vector.ts
+++ b/packages/core/src/vector.ts
@@ -132,11 +132,10 @@ export function vector<K extends Float32Array | Float64Array | Int8Array | Int16
 }
 
 /**
- * Test if given object is an instance of {@link Point} class.
+ * Test if given object is an instance of the {@link Vector} class.
  * @param {Object} obj the object to test.
- * @return {boolean} `true` if given object is a {@link Point}, `false` otherwise.
+ * @return {boolean} `true` if given object is a {@link Vector}, `false` otherwise.
  */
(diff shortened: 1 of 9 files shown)


packages/core/src/vector.ts
```diff
@@ -98,3 +130,13 @@ export default class Vector<K extends Float32Array | Float64Array | Int8Array |
 export function vector<K extends Float32Array | Float64Array | Int8Array | Int16Array | Int32Array | BigInt64Array> (typedArray: K): Vector<K> {
   return new Vector(typedArray)
 }
+
+/**
+ * Test if given object is an instance of {@link Point} class.
+ * @param {Object} obj the object to test.
+ * @return {boolean} `true` if given object is a {@link Point}, `false` otherwise.
```
robsdedude:
```suggestion
 * Test if given object is an instance of the {@link Vector} class.
 * @param {Object} obj the object to test.
 * @return {boolean} `true` if given object is a {@link Vector}, `false` otherwise.
```

packages/core/test/vector-type.test.ts
```diff
@@ -26,10 +26,27 @@ describe('Vector', () => {
       ['BigInt64Array', BigInt64Array.from([BigInt(0)]), 'INT64'],
       ['Float32Array', Float32Array.from([0]), 'FLOAT32'],
       ['Float64Array', Float64Array.from([0]), 'FLOAT64']
+
```
robsdedude:
```suggestion
```

robsdedude (commented):
:books:

robsdedude (approved):
:chains:

PR #1329, opened 2025-09-15
FailFastSecurityError check bug

MaxAake:
The _isFailFastSecurityError() check checks if the error code starts with Neo.ClientError.Security. which causes a crash if a non neo4j error surfaces. This PR fixes that issue.

Cherry picks: #1328


StephenCathcart (approved):
👍

PR #1343, opened 2025-09-30
Improve Vector constructor error message

MaxAake:


packages/core/src/vector.ts
```diff
@@ -67,10 +67,10 @@ export default class Vector<K extends Float32Array | Float64Array | Int8Array |
       typedArray instanceof Uint32Array ||
       typedArray instanceof BigUint64Array
     ) {
-      throw newError('The neo4j Vector class does not support Unsigned Integer Arrays, please use a signed IntArray')
+      throw newError('Invalid argument type passed to Vector constructor: should be signed integer or float TypedArray, got unsigned integer TypedArray')
```
robsdedude: Would it make sense/improve things to also include the exact type/value in this error message in a similar way the other changed error message does?

MaxAake: An updated version has been pushed, with just 1 error each in the constructor and function

robsdedude (approved):
:detective: :zap:

PR #1348, opened 2025-10-03
Replace lerna bootstrap with pnpm

MaxAake:
Closes DRIVERS-65


packages/neo4j-driver-lite/test/integration/driver.test.ts
```diff
@@ -22,11 +22,11 @@ describe('neo4j-driver-lite', () => {
 
   beforeAll(async () => {
     await Config.startNeo4j()
-  }, 20000)
+  }, 60000)
```
AndyHeap-NeoTech, line 25: I'm assuming this is due to flakiness arising from slower start up times or something?

testkit/integration.py
```diff
@@ -12,15 +12,14 @@
     os.environ["TEST_CONTAINERS_DISABLED"] = "True"
 
     if is_lite():
-        ignore = "--ignore=neo4j-driver"
+        filter = "--filter=!neo4j-driver"
```
AndyHeap-NeoTech, line 15: Just want to confirm that this should be reversing the logical condition with !

AndyHeap-NeoTech (approved):
Looks good, just a couple of small questions.

packages/neo4j-driver-lite/test/integration/driver.test.ts
MaxAake: That is correct, it's had some flakiness before but it seemed to get a bit worse here.

testkit/integration.py
MaxAake: Yes, `ignore `selects packages to not run on. `filter ` decides packages TO run on.

Issue #1351, opened 2025-10-13
Add early return for `Transformer.toStructure` for primitive values


zemberdotnet:
# Feature Request
Implement early return in `Transformer.toStructure` for primitives.

https://github.com/neo4j/neo4j-javascript-driver/blob/8716a7d0d4b8aa8658bdfc59ab59874a6ae35f76/packages/bolt-connection/src/bolt/transformer.js#L63-L70

# Pitch

The `toStructure` method is expensive when the `type` parameter is a primitive value.

My understanding is this method is called once for every value that goes into the packstream including objects, arrays, keys of objects, and object values recursively. When the value is primitive, the `find` on line 64 loops through all `TypeTransformers`, calling `isTypeInstance` on each one. However, no `transformer` is found for primitives, and we return the value as is on line 68.

## Proposed Change

I added the following check at the top of `toStructure`:
```js
const valueType = typeof type;
// captures 'string' | 'boolean' | 'bigint' | 'number' | 'undefined' | null
if ((valueType !== 'object' && valueType !== 'function' && valueType !== 'symbol') || type === null) {
  return type
}
```
The time to execute `DelegateConnection.run` went from ~1300ms to ~670ms. The total duration for `Transformer.toStructure` went from ~520ms to ~50ms.

My profile was based on running a query with a parameter that is a list of 100,000 objects, each with 20 keys:
- 11 string values
- 6 undefined
- 1 null value
- 1 array of 1 string
- 1 number

## Alternative Solution

The above solution might assume too much about what the `_transformers` will provide. If that's a concern, an alternative would be to implement a `TypeTransformer` whose `isTypeInstance` checks for primitive values and whose `toStructure` is an identity function. This transformer would then need to be placed at the start of the `_transformers` array to ensure it's checked first.

MaxAake:
Hello @zemberdotnet !

Thank you for flagging this, this absolutely looks like a place where we can score some cheap performance improvements. I'll take some time to investigate alternative designs to make sure the best solution is chosen, but this improvement should be ready to be released in the next minor release of the diver

zemberdotnet:
Awesome! Thank you for implementing this. It looks like the PR was merged, so I'll close the issue now.

PR #1352, opened 2025-10-15
add quick return in toStructure() for primitive values

MaxAake:
The toStructure function is run on every element being sent over bolt, which loops over all available typeTransformers and checks if they're valid for transforming what is being sent to a structure. This is rather wasteful for things like strings and numbers, which we do not have any transformers for.

Based on a suggestion in  #1351, the improvement will only work for as long as we do not have any transformers for primitive types, but there are currently no plans which would require that.


StephenCathcart (approved):
👍

PR #1353, opened 2025-10-27
Add todo to point stringer for 7.0 update

MaxAake:
closes DRIVERS-231
The toString of the Point object is almost correctly formatted for it to be copy-pasteable into a cypher query. Given that the Vector object stringer was designed to allow it to be copy-pasted, it would make sense to make the same true for Points.


AndyHeap-NeoTech (approved):
LGTM

PR #1356, opened 2025-10-31
Correct exporting of vector features in ts files

MaxAake:
Primarily adds `vector` and `Vector` to the index.d.ts exports. Also ensures that `VectorType` is consistently exported only as a type (for type annotations)


StephenCathcart (approved):
👍

PR #1358, opened 2025-11-06
Remove jasmine usage in neo4j-driver tests

MaxAake:
Closes Drivers-191

The driver has been using a mix of Jest and Jasmine for tests (Karma was also used until recently). This PR removes all use of Jasmine in favor of a jest only approach


packages/neo4j-driver/test/rx/transaction.test.js
```diff
@@ -535,9 +458,8 @@ describe('#integration-rx transaction', () => {
     await verifyNoFailureIfNotExecuted(false)
   })
 
-  it('should not propagate run failure from summary', async () => {
-    pending('behaviour difference across drivers')
-
+  // Skipped due to behavioral difference in the JS driver
```
StephenCathcart, line 461: Do you think this behavioural difference needs to be triaged against the other drivers?

StephenCathcart (approved):
💯

packages/neo4j-driver/test/rx/transaction.test.js
MaxAake: I'm not actually sure. I need to investigate this further. This is specifically a matter on reactive sessions, and this test has been like this (`pending` is just a jasmine way to skip) since the original PR by Ali that introduced reactive sessions to the JS driver. So this has gone untouched by both me and Antonio. I'll card up an investigation on Linear, good call.

PR #1363, opened 2025-11-20
Improve error message for starting work on a busy session

MaxAake:
closes DRIVERS-220

The message is unclear for users running into this due to having ongoing `session.run()` work


AndyHeap-NeoTech (approved):
Looks good to me.

PR #1373, opened 2025-12-18
Revert "bump deno to 5.28.4 (#1371)"

MaxAake:
deno version was erroneously bumped twice after last patch release. upcoming version is 5.28.3


MaxAake:
closed as the release can be run on the previous commit without issue. Easier than reverting and then redoing

PR #1379, opened 2026-01-20
debump deno version to 5.28.3

MaxAake:
reported deno version had erroneously been bumped twice, this must be corrected before a patch release can be made.


robsdedude (approved):
:elevator: :arrow_down:

PR #1389, opened 2026-02-09
Improve Session.run docs

MaxAake:
Closes DRIVERS-261

### DIFF
--- a/packages/core/src/session.ts
+++ b/packages/core/src/session.ts
@@ -177,6 +177,17 @@ class Session {
    * Could be called with a query object i.e.: `{text: "MATCH ...", parameters: {param: 1}}`
    * or with the query and parameters as separate arguments.
    *
+   * This is an auto-commit query, meaning the server will commit the results without further input.
+   *
+   * The transaction is guaranteed to be committed anywhere between the start of
+   * it (i.e., calling :meth:`neo4j.Session.run`) and fully consuming the
+   * returned {@link Result}. The Result can be consumed as a promise or lazily using {@link Result#subscribe}.
+   *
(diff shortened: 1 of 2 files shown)


AndyHeap-NeoTech (approved):
All good

packages/core/src/session.ts
```diff
@@ -177,6 +177,17 @@ class Session {
    * Could be called with a query object i.e.: `{text: "MATCH ...", parameters: {param: 1}}`
    * or with the query and parameters as separate arguments.
    *
+   * This is an auto-commit query, meaning the server will commit the results without further input.
+   *
+   * The transaction is guaranteed to be committed anywhere between the start of
+   * it (i.e., calling :meth:`neo4j.Session.run`) and fully consuming the
```
robsdedude, line 183: > :meth:`neo4j.Session.run`

Looks like a copy-pasta left-over.

packages/core/src/session.ts
```diff
@@ -177,6 +177,17 @@ class Session {
...
+   * returned {@link Result}. The Result can be consumed as a promise or lazily using {@link Result#subscribe}.
+   *
+   * If the code moves on (i.e by calling {@link Session#close} or by starting another session)
```
robsdedude, line 186:
```suggestion
   * If the code moves on (i.e., by calling {@link Session#close} or by starting another session)
```

packages/core/src/session.ts
```diff
@@ -177,6 +177,17 @@ class Session {
...
+   * before the Result is fully consumed, it will be consumed in th background. However,
```
robsdedude, line 187:
```suggestion
   * before the Result is fully consumed, it will be consumed in the background. However,
```

PR #1396, opened 2026-03-10
ProfiledPlan: expose existence for all metrics

robsdedude:
Deprecate `ResultSummary.profile` of type `ProfiledPlan` in favor of `ResultSummary.queryProfile` of type `Profile`.

The main difference between the profile types is that the new one exposes whether stats (such as `dbHits`, `time`, etc.) have been recorded by being nullable.

According to a chat with the Cypher runtime team, the absence of the value is not equivalent to it being zero. Therefore it makes sense to express this information in the API. The information is encoded in the bolt protocol through absence or existence of the respective fields.

Closes: DRIVERS-283

Depends on:
 * https://github.com/neo4j-drivers/testkit/pull/705


robsdedude:
Please don't merge yet. I'd like to wait for this work to be ready for merging across all drivers first.

packages/core/src/result-summary.ts
```diff
@@ -200,6 +211,7 @@ class Plan {
 /**
  * Class for execution plan received by prepending Cypher with PROFILE.
  * @access public
+ * @deprecated has been superseded by {@link Plan}
```
MaxAake:
```suggestion
 * @deprecated has been superseded by {@link Profile}
```
`Plan` is the already existing class that you get on an `EXPLAIN`, `Profile` is the new class you will get on a `PROFILE` query

packages/core/src/result-summary.ts
```diff
@@ -91,16 +92,26 @@ class ResultSummary<T extends NumberOrInteger = Integer> {
...
      * @public
+     * @deprecated has been superseded by {@link ResultSummary.queryProfile}
      */
     this.profile = metadata.profile != null ? new ProfiledPlan(metadata.profile) : false
 
+    /**
+     * This describes how the database did execute your query. This will contain detailed information about what
+     * each step of the plan did. Profiled query plan for the executed query if available, otherwise undefined.
+     * Will only be populated for queries that start with "PROFILE".
+     * @type {Plan}
```
MaxAake:
```suggestion
     * @type {Profile}
```
Same as other comment

MaxAake (changes requested):

packages/core/src/result-summary.ts
MaxAake, line 185: I believe there should be a version of this function for the new `queryProfile` as well, and potentially this should be marked deprecated.

MaxAake (changes requested):
Apologies for missing this bit the first time round

MaxAake (approved):
✅

PR #1400, opened 2026-03-26
Alter connection release reset logic to avoid breaking state machine

MaxAake:
Closes DRIVERS-303

If referenceCount isn't 0 then something is using the ConnectionHolder, meaning the connection is in use and needs to be reset before being returned to the pool, even if there is not an active explicit transaction.

### DIFF
--- a/packages/core/src/connection.ts
+++ b/packages/core/src/connection.ts
@@ -122,6 +122,14 @@ class Connection {
 
   /**
    *
+   * @returns {BoltProtocol}
+   */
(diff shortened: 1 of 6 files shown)


AndyHeap-NeoTech (approved):
Looks good. I'm assuming the testkit tests will cover this as they were failing before this fix.

MaxAake:
While this is an improvement, it doesn't fully solve the issue, which is related to #1402

PR #1403, opened 2026-04-01
Result.close() send discard directly if waiting for more

MaxAake:
closes: DRIVERS-303


packages/bolt-connection/src/bolt/stream-observers.js
```diff
@@ -173,13 +175,20 @@ class ResultStreamObserver extends StreamObserver {
...
+    this._waitingForMore = false
   }
 
   /**
    * Cancel pending record stream
    */
   cancel () {
     this._discard = true
+    if (this._waitingForMore) {
+      this._discardFunction(this._queryId, this)
```
robsdedude: `_waitingForMore` could end up `true` while a `PULL` is in flight. I won't for the 1st badge, but every consecutive batch it will.

robsdedude (changes requested):

packages/bolt-connection/src/bolt/stream-observers.js
robsdedude:
```suggestion
    this._setupAutoPull()
```
🧹

packages/bolt-connection/src/bolt/stream-observers.js
```diff
@@ -173,13 +175,21 @@ class ResultStreamObserver extends StreamObserver {
...
    * Cancel pending record stream
    */
   cancel () {
     this._discard = true
+    // if the stream observer is waiting for user input to pull next, such as for an async iterator, we need to send the discard right away.
+    if (this._waitingToPullMore && (!this._autoPull || this._paused)) {
+      this._discardFunction(this._queryId, this)
+      this._waitingToPullMore = false
+      this._eagerlyDiscarded = true
+      this._setState(_states.STREAMING)
```
robsdedude: Isn't this equivalent to 

```suggestion
      this._waitingToPullMore = false
      this._more()
```

while getting rid of `_eagerlyDiscarded`?

packages/bolt-connection/src/bolt/stream-observers.js
robsdedude: I think this should also set `streamObserver._waitingToPullMore = false`. Maybe `ResultStreamObserver._more` should generally set `streamObserver._waitingToPullMore = false`. After all it's the function that dispatches the `RUN`/`DISCARD` messages. Please double check, but I've not found a place where that would be the wrong course of action.

robsdedude, line 1: I wonder if it would make for easier to follow code, if the state were used to track how a cancel needs to be handled. Something like adding a `discard` method to all states and

 * `READY_STREAMING`, `FAILED`, `SUCCEEDED` are no-op `() => {}`
 * `READY` eagerly discards (`streamObserver._discard = true; streamObserver._more()`)
 * `STREAMING` only schedules a discard `streamObserver._discard = true`

And `ResultStreamObserver.cancel` than just has to call `this._state.discard()`.

robsdedude: Maybe add a test provoking this misbehavior. I think you'd have to get the driver to pause right after a batch so it's in the `READY` and `_waitingToPullMore = true` state. Then `resume` and `cancel` right after.

MaxAake: I agree that the suggested change should be done, however I am unable to add a test that fails under the conditions described. This is because the `resume` function sets `_autoPull` to true, which blocks eager discards

MaxAake: I made a similar test, to catch issues with future changes.

packages/bolt-connection/src/bolt/stream-observers.js
```diff
@@ -717,7 +722,14 @@ const _states = {
     name: () => {
       return 'READY'
     },
-    pull: streamObserver => streamObserver._more()
+    pull: streamObserver => streamObserver._more(),
+    discard: (obs) => {
+      obs._discard = true
+      // if the stream observer is waiting for user input to pull next, such as for an async iterator, we need to send the discard right away.
+      if (obs._autoPull !== true || obs._paused) {
```
robsdedude:
```suggestion
      if (obs._autoPull === false || obs._paused) {
```

Or not equivalent but even more readable and practically equivalent
```suggestion
      if (!obs._autoPull || obs._paused) {
```

packages/bolt-connection/test/bolt/stream-observer.test.js
```diff
@@ -886,6 +886,86 @@ describe('#unit ResultStreamObserver', () => {
...
+    it.each([
+      ['reactive/subscribed', true, true],
+      ['reactive/unsubscribed', true, false],
+      ['not reactive/subscribed', false, true],
+      ['not reactive/unsubscribed', false, false]
+    ])('should handle cancel before run success well: \'%s\'', (_, reactive, shouldSubscribe) => {
+      const discardFunction = jest.fn()
+      const moreFunction = jest.fn()
+      const obs = new ResultStreamObserver({ reactive, moreFunction, discardFunction })
+      if (shouldSubscribe === true) {
```
robsdedude, line 914:
```suggestion
      if (shouldSubscribe) {
```
Nit: wouldn't have bothered, but blow it's `if (!reactive) {...}`.

packages/bolt-connection/test/bolt/stream-observer.test.js
```diff
@@ -886,6 +886,86 @@ describe('#unit ResultStreamObserver', () => {
...
+      }
+      obs.cancel()
+      expect(discardFunction).toHaveBeenCalledTimes(0)
+      expect(moreFunction).toHaveBeenCalledTimes(0)
+      expect(obs._state.name()).toBe(reactive ? 'READY' : 'READY_STREAMING')
+      expect(obs._discard).toBe(true)
+      obs.onCompleted({ fields: ['A'] })
+      if (!reactive) {
+        // unreactive runs always pipeline their pulls, and will not discard until the pull success returns.
+        obs.onCompleted({ key: 42, has_more: true })
```
robsdedude, line 925: Optional: add a test without `has_more` / `hast_more: false` and assert the result doesn't try to send a DISCARD on a finished record stream.

packages/neo4j-driver/test/session.test.js
```diff
@@ -983,6 +983,31 @@ describe('#integration session', () => {
...
+      }
+    }
+
+    const conn = await session._writeConnectionHolder.getConnection()
+    await session.close()
+    expect(conn._delegate._protocol._lastMessageSignature).toBe(0x0f)
+
+    session = driver.session({ fetchSize: 2 })
+
+    await expect(session.run('RETURN 1 as n')).resolves.not.toThrow()
```
robsdedude:
```suggestion
    await expect(session.run('RETURN 1 AS n')).resolves.not.toThrow()
```
🫥

robsdedude (approved):

packages/bolt-connection/test/bolt/stream-observer.test.js
MaxAake: I prefer doing to explicit check, partially because the linter for TypeScript code enforces it. I'll instead update the other check.

PR #1404, opened 2026-04-15
Add retries for idempotent errors on Session.run()

MaxAake:
Closes: DRIVERS-309

### DIFF
--- a/packages/core/src/result.ts
+++ b/packages/core/src/result.ts
@@ -21,7 +21,7 @@ import ResultSummary from './result-summary'
 import Record, { RecordShape } from './record'
 import { Query, PeekableAsyncIterator } from './types'
 import { observer, util, connectionHolder } from './internal'
-import { newError, PROTOCOL_ERROR } from './error'
+import { Neo4jError, newError, PROTOCOL_ERROR } from './error'
 import { NumberOrInteger } from './graph-types'
 import Integer from './integer'
 import { GenericConstructor, Rules } from './mapping.highlevel'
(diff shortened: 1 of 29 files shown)


packages/core/src/result.ts
```diff
@@ -529,14 +544,28 @@ class GenericResult<R, T extends GenericQueryResult<R>> implements Promise<T> {
...
-      // notify connection holder that the used connection is not needed any more because error happened
-      // and result can't bee consumed any further; call the original onError callback after that
-      this._connectionHolder.releaseConnection().then(() => {
-        replaceStacktrace(error, this._stack)
-        this._error = error
-        onErrorOriginal.call(observer, error)
-      }).catch(onErrorOriginal)
+    const onErrorWrapper = (error: Error, runError?: boolean): void => {
+      if (
+        runError === true && !this._hasRetried && error instanceof Neo4jError &&
```
StephenCathcart: might be misreading the flow here - `_hasRetried` is checked inside each wrapper, and `_handleError` walks the observer list calling them in turn. if someone attaches two consumers to the same result (`result.keys()` then `await result` or similar), does that mean the first wrapper sets `_hasRetried = true` and queues the retry, while the second wrapper sees the flag, falls into the `else`, releases the connection holder and rejects with the idempotent error?

do you think this is guarded elsewhere, or maybe just "bad usage" and ignore?

StephenCathcart (approved):
Other than open question looks good to me 👍

packages/core/src/result.ts
MaxAake: No this is indeed an issue, thank you! I think in my mental model, the fact that we unsubscribe from the stream observer would protect against this, but since the `.forEach` loop has started it will run the error callback from every observer that was subscribed when it started, even if the first one removes the subsequent entries! 

I altered the code to instead of tracking if a result **has** retried, it tracks the exact error that it retried due to. As all the observers on the result will receive the exact same error object (not just an identical object but the same one) this can be used to safeguard against treating the same error as a 2nd error

StephenCathcart (approved):
👍

PR #1406, opened 2026-05-06
Fix error piping in gulpfile

MaxAake:
Closes: DRIVERS-355

### DIFF
--- a/packages/neo4j-driver/gulpfile.babel.js
+++ b/packages/neo4j-driver/gulpfile.babel.js
@@ -158,11 +158,11 @@ gulp.task('set', function () {
 })
 
 gulp.task('start-neo4j', function (done) {
-  sharedNeo4j.start().then(done).catch(error => done.fail(error))
+  sharedNeo4j.start().then(done).catch(done)
 })
 
 gulp.task('stop-neo4j', function (done) {
-  sharedNeo4j.stop().then(done).catch(error => done.fail(error))
+  sharedNeo4j.stop().then(done).catch(done)
 })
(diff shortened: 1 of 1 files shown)


packages/neo4j-driver/gulpfile.babel.js
```diff
@@ -158,11 +158,11 @@ gulp.task('set', function () {
 })
 
 gulp.task('start-neo4j', function (done) {
-  sharedNeo4j.start().then(done).catch(error => done.fail(error))
+  sharedNeo4j.start().then(done).catch(done)
```
robsdedude, line 162: Unfortunately this way of defining gulp tasks is deprecated. These day exporting the task functions is the recommended way to go. I'm not saying we should update this in this PR. I just mentioned it because the docs don't explain how to signal errors with the deprecated `gulp.task` API. However, looking at https://gulpjs.com/docs/en/getting-started/async-completion#using-an-error-first-callback this approach might be applicable to the legacy API, too.

It seems odd to just silence errors when the task fails to start up the DBMS.

robsdedude: If however, you decide to stick with the silencing, I suggest


```suggestion
  sharedNeo4j.start().finally(done)
```

robsdedude: Never mind me 🙈 I didn't realize that `.catch(cb)` will actually forward the error. Kthxbye.

robsdedude (approved):

PR #1415, opened 2026-06-03
Bolt 6.1 and UUID

MaxAake:
closes: DRIVERS-364


packages/core/src/uuid.ts
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Represents a Neo4j UUID value, stored as a Uint8Array of length 16, which can be gotten as a uuid string.
+ * @access public
+ * @exports UUID
+ * @class A class for writing/reading the Neo4j UUID type as a Uint8Array.
+ * @param { Uint8Array } typedArray A Uint8Array of length 16 representing the UUID
+ */
+export default class UUID {
+  public readonly _typedArray: Uint8Array
+  constructor (typedArray: Uint8Array) {
```
StephenCathcart, line 34: just a thought, the go driver implementation was immutable but here the uuid looks mutable. happy to go with whatever, might be worth raising with the team?

packages/core/src/uuid.ts
```diff
@@ -0,0 +1,115 @@
...
+  enumerable: false,
+  configurable: false,
+  writable: false
+})
+
+/**
+ * Created a {@link UUID} from a uuid string or length 16 Uint8Array, allowing it to be sent as a parameter of a Neo4j query.
+ * @access public
+ * @param {Uint8Array | string} rawUUID - a uuid string or length 16 Uint8Array with the values of the UUID.
+ * @return {Vector} - The Neo4j UUID ready to be used as a query parameter
```
StephenCathcart:
```suggestion
 * @return {UUID} - The Neo4j UUID ready to be used as a query parameter
```

StephenCathcart (approved):
👍

packages/core/src/uuid.ts
MaxAake: I think I agree with you that they should be made immutable. The other drivers all have language-native uuids though so I'm not sure if it's relevant to anyone else? (unless I'm mistaken)

packages/core/src/uuid.ts
MaxAake: 🤦🏻

packages/core/src/uuid.ts
StephenCathcart: Yeah it might just be this and the go driver with the custom types, and I don't think it's much concern in hindsight 👍

PR #1442, opened 2026-06-29
add UUID to object mapping

MaxAake:
Closes: DRIVERS-479

### DIFF
--- a/packages/core/src/mapping.rulesfactories.ts
+++ b/packages/core/src/mapping.rulesfactories.ts
@@ -490,7 +490,7 @@ export const rule = Object.freeze({
     }
   },
   /**
-   * Create a {@link Rule} that validates the value is a Vector.
+   * Create a {@link Rule} that validates the value is a {@link Vector}.
    *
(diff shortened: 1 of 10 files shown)


packages/core/test/mapping.rulesfactories.test.ts
robsdedude, line 1: Also consider extending other tests `mapping.highlevel.test.ts`'s `should be able to read all property types`.

packages/core/src/mapping.rulesfactories.ts
```diff
@@ -518,6 +521,32 @@ export const rule = Object.freeze({
       ...rule
     }
   },
+  /**
+   * Create a {@link Rule} that validates the value is a UUID.
```
robsdedude:
```suggestion
   * Create a {@link Rule} that validates the value is a {@link UUID}.
```

packages/core/src/mapping.rulesfactories.ts
robsdedude:
```suggestion
   * Create a {@link Rule} that validates the value is a {@link Vector}.
```

robsdedude (commented):
Only minor suggestions. No blockers.

packages/core/test/mapping.rulesfactories.test.ts
MaxAake: good suggestion!

robsdedude (approved):
📂

PR #1446, opened 2026-08-17
Derive the AES key from the DEK with HKDF-SHA256

RichardIrons-neo4j:
Investigated why the JS and dotnet drivers weren't compatible and discovered that the JS driver was using the decapsulated DEK directly as the AES-GCM key, missing the key derivation step, but the dotnet driver does do that and when it tried to decrypt the JS-generated fixture it failed with an invalid key.  I don't know if this change is idiomatic as Claude helped me find it, but with this change in place and the latest pull to the testkit encryption tests branch, both drivers can decrypt each other's encrypts.

I couldn't figure out how to link to the right part of the ADR so I'm just pasting the relevant bit:

> Rather than using the primary encryption key directly, drivers derive purpose-specific AES keys using HKDF (HMAC-based Extract-and-Expand Key Derivation Function) with HMAC-SHA256. The derivation context (info) is set to neo4j/property-encryption/v1, providing domain separation for property encryption keys and allowing independent keys to be derived from the same master key for different purposes.


MaxAake:
Due to github site issues this change was made in a separate commit on the branch rather than via this PR.

PR #1449, opened 2026-09-03
Fix record prototype walkup

MaxAake:
Part of: DRIVERS-544

`Record.has()` can find properties that are in the base `Object` prototype such as `constructor` and `toString`, and will report them as being present.


StephenCathcart (approved):
👍

PR #1450, opened 2026-09-03
Fix buffer reads and remove dead code

MaxAake:
Part of: DRIVERS-543

Adds errors to reads beyond the end of the buffer, fixes the fact that `getUInt32` actually gets a signed int, Improves multi-byte VarInts (4 or more bytes still broken) and removes dead, broken code related to reading and writing int64s

### DIFF
--- a/packages/bolt-connection/src/buf/base-buf.js
+++ b/packages/bolt-connection/src/buf/base-buf.js
@@ -14,7 +14,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { newError } from 'neo4j-driver-core'
 
 /**
  * Common base with default implementation for most buffer methods.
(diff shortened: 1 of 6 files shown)


packages/bolt-connection/src/buf/base-buf.js
```diff
@@ -311,14 +280,6 @@ export default class BaseBuffer {
     this.putUInt32(this._updatePos(4), val)
   }
 
-  /**
-   * Write to state position.
-   * @param val
-   */
-  writeInt64 (val) {
```
StephenCathcart, line 318: checked and `readInt64` is still in the file, and i don't think anything calls it. can that be removed too?

StephenCathcart (approved):
one question otherwise looks good 👍

packages/bolt-connection/src/buf/base-buf.js
MaxAake: 🤦🏻 yep. That should go too. Good catch!

PR #1451, opened 2026-09-04
Do not silently coerce schemeless tokens into none

MaxAake:
Part of: DRIVERS-547 

The driver will currently interpret any token without a set scheme as none auth silently.

### DIFF
--- a/packages/neo4j-driver-lite/src/index.ts
+++ b/packages/neo4j-driver-lite/src/index.ts
@@ -57,6 +57,7 @@ import {
   LocalTime,
   ManagedTransaction,
   Neo4jError,
+  newError,
   Node,
   Notification,
   GqlStatusObject,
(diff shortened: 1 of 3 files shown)


packages/neo4j-driver-lite/src/index.ts
```diff
@@ -165,10 +166,17 @@ function createAuthManager (authTokenOrProvider: AuthToken | AuthTokenManager):
     return authTokenOrProvider
   }
 
-  let authToken: AuthToken = authTokenOrProvider
+  const authToken: AuthToken = { ...authTokenOrProvider }
   // Sanitize authority token. Nicer error from server when a scheme is set.
-  authToken = authToken ?? {}
-  authToken.scheme = authToken.scheme ?? 'none'
+  if (authToken.scheme == null) {
+    if (authToken.principal != null || authToken.credentials != null || authToken.realm != null || authToken.parameters != null) {
```
StephenCathcart: tested `{ principal, credentials }`, got the new error. tested `{ username, password }`, got:

Neo.ClientError.Security.Unauthorized: Unsupported authentication token, scheme 'none' is only allowed when auth is disabled.

would checking if the object has _**any**_ keys cover both, or happy with the list for now? the username/password might not be valid - was just messed around, looks good either way

StephenCathcart (approved):
small comment otherwise looks good

PR #1454, opened 2026-09-07
Use isolated string decoder

MaxAake:
Part of: DRIVERS-546

The driver currently uses the same string decoder for every time multiple buffers need to be concatenated. This currently has no known consequences, but if an error were to occur during this handling, there could be remaining data in the decoder that could cause an error the next time the decoder is used.


StephenCathcart (approved):
👍

PR #1455, opened 2026-09-07
Fix assignment to __proto__ when unpacking maps

MaxAake:
Part of: DRIVERS-545

The way keys are assigned to created objects when unpacking maps means that the prototype of the map can be overwritten. This is unlikely to cause a major issue, and would require a user to return a map with a field __proto__. However it should still be fixed.


StephenCathcart (approved):
👍

PR #1456, opened 2026-09-09
Fix deno generate on windows

MaxAake:
Closes: DRIVERS-554

Due to windows using `\` rather than `/` in paths, the deno generate file has not functioned when executed on plain windows. I decided to fix that to avoid the performance hit of using WSL.


StephenCathcart (approved):
👍

PR #1457, opened 2026-09-10
Fix small issue in Integer test

MaxAake:
Closes: DRIVERS-555

A logic error if the rng rolled an exact 0 for the test value, very rare flakiness but a quick fix


StephenCathcart (approved):
👍

